PR #2, opened 2022-03-29
Updates the example project's code to the latest client release (0.2.1)

AdamGS:


examples/java-basic-mvn/src/main/java/pineconeexamples/MinimalUpsertAndQueryExample.java
```diff
@@ -1,62 +1,85 @@
...
 
+    public static void main(String[] cliArgs) {
         System.out.println("Starting application...");
 
+        Args args = new Args();
+
         PineconeClientConfig configuration = new PineconeClientConfig()
-                .withApiKey(apiKey);
+                .withApiKey(args.apiKey)
+                .withEnvironment(args.environment)
```
benjaminran, line 27: I wish we could put a comment here pointing to some docs about how to get env/project name. Best candidate would probably be https://www.pinecone.io/docs/quickstart/ > "About your index endpoint", but maybe that's too likely to be moved in future.

examples/java-basic-mvn/src/main/java/pineconeexamples/UpsertsAndQueriesConcurrentExample.java
```diff
@@ -1,46 +1,36 @@
...
-import io.pinecone.PineconeClientConfig;
-import io.pinecone.PineconeConnection;
-import io.pinecone.PineconeConnectionConfig;
-import io.pinecone.PineconeException;
-import io.pinecone.PineconeResponse;
-import io.pinecone.QueryResponse;
-import io.pinecone.UpsertResponse;
+import com.google.common.primitives.Floats;
+import io.pinecone.*;
+import io.pinecone.proto.*;
```
benjaminran, line 5: Could you expand these so that readers see exactly where symbols like UpsertResponse are coming from below?

benjaminran (approved):

PR #12, opened 2023-08-16
Add example for querying by single vector

rajat08:
## Problem
Current examples in the test file only talk about the deprecated query pattern; change that to add other versions. 

## Solution
Add more examples in the PineconeLiveIntegTest file.

## Type of Change

## Test Plan
No test is required for this change.

N/A


rohanshah18 (approved):
Thanks for adding more examples!

PR #13, opened 2023-08-17
Add delete index support

rohanshah18:
## Problem

The java client is missing delete index support, so this PR aims at adding the function to make http call to the delete index endpoint.

## Solution

I've added the delete index client along with a unit test. As a follow up to this PR, I'll be adding an example to demonstrate how to consume the delete index client to the examples subfolder.

## Type of Change

## Test Plan

A unit test is added to test the functionality.

### DIFF
--- a/src/main/java/io/pinecone/PineconeIndexOperationClient.java
+++ b/src/main/java/io/pinecone/PineconeIndexOperationClient.java
@@ -1,25 +1,34 @@
 package io.pinecone;
 
 import org.asynchttpclient.AsyncHttpClient;
-import org.asynchttpclient.DefaultAsyncHttpClient;
 
 import java.io.IOException;
 
 public class PineconeIndexOperationClient {
-
     private AsyncHttpClient client;
+    private PineconeClientConfig clientConfig;
+    private PineconeConnectionConfig connectionConfig;
+    private String url;
 
-    PineconeIndexOperationClient(AsyncHttpClient client) {
+    PineconeIndexOperationClient(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig, AsyncHttpClient client) {
+        this.clientConfig = clientConfig;
+        this.connectionConfig = connectionConfig;
         this.client = client;
+        this.url = "https://controller." + clientConfig.getEnvironment() + ".pinecone.io/databases/";
     }
 
-    public void deleteIndex(String environment, String apiKey, String indexName) throws IOException {
-        String url = "https://controller." + environment + ".pinecone.io/databases/" + indexName;
+    public PineconeIndexOperationClient(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig) {
+        this.clientConfig = clientConfig;
+        this.connectionConfig = connectionConfig;
+        this.url = "https://controller." + clientConfig.getEnvironment() + ".pinecone.io/databases/";
+    }
+
+    public void deleteIndex() throws IOException {
         System.out.println("Sending delete index request:");
         // ToDo: Capture Response response and customized error messages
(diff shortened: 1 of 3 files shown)


src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class PineconeIndexOperationClient {
+
+    private AsyncHttpClient client;
+
+    PineconeIndexOperationClient(AsyncHttpClient client) {
+        this.client = client;
+    }
+
+    public void deleteIndex(String environment, String apiKey, String indexName) throws IOException {
```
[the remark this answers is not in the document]
rajat08: Cool, I leave that decision to you and Jen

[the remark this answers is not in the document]
rohanshah18: Thanks!

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -0,0 +1,31 @@
...
+        String url = "https://controller." + environment + ".pinecone.io/databases/" + indexName;
+        System.out.println("Sending delete index request:");
+        // ToDo: Capture Response response and customized error messages
+        client.prepare("DELETE", url)
```
rajat08, line 20: This is a blocking call; if we want to mimic Python client behavior, we should add a timeout. The default is 300s and the user can pass a custom timeout, with -1 being returned immediately. Not needed immediately though I suppose

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
+
+import static org.mockito.Mockito.*;
+
+public class PineconeIndexOperationClientTest {
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testSuccessfulIndexDeletion() throws IOException {
```
rajat08, line 15: A negative test case would be nice! does not have to be as comprehensive but here is the test suite for python in case you want ideas : 
https://github.com/pinecone-io/pinecone-db/blob/main/tests/integration/test_control_plane.py

rajat08 (approved):

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: Good point!

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: Currently the unit test is not making a call to the database and as a part of integration test, I'll be adding various possible negative tests. Thanks for providing me with the python reference.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
jhamon, line 16: Could we move the `environment` and `apiKey` params to the client constructor so this method only takes the `indexName`?

jhamon (approved):
This looks like a good start. If you want to change the method signature (based on my feedback), I'm fine with handling that as a follow-up PR.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: Yes, my plan is to use two existing config objects and remove any parameters from deleteIndex method. Let me actually make that change now.

PR #15, opened 2023-08-21
Add support for optional fields to create index

rohanshah18:
## Problem
The sdk currently lacks support to allow configuring optional fields while creating a new index. 

## Solution
Add support to add optional fields while creating a new index. Newly supported optional fields are pods, replicas, pod_type, source_collection, and metadata_config.

## Type of Change

## Test Plan
Added unit tests to test for the required and the optional fields.

Along with unit tests, an integration test to create the index was successfully ran locally with required fields and verified that the index was created using console.

### DIFF
--- a/src/main/java/io/pinecone/model/CreateIndexRequest.java
+++ b/src/main/java/io/pinecone/model/CreateIndexRequest.java
@@ -70,7 +70,6 @@ public class CreateIndexRequest {
      * The distance metric to be used for similarity search. You can use 'euclidean', 'cosine', or 'dotproduct'.
      * @return metric
      **/
-    @javax.annotation.Nullable
     public String getMetric() {
         return metric;
     }
@@ -88,7 +87,6 @@ public class CreateIndexRequest {
      * The number of pods for the index to use,including replicas.
      * @return pods
      **/
-    @javax.annotation.Nullable
     public Integer getPods() {
         return pods;
     }
@@ -106,7 +104,6 @@ public class CreateIndexRequest {
      * The number of replicas. Replicas duplicate your index. They provide higher availability and throughput.
      * @return replicas
      **/
-    @javax.annotation.Nullable
     public Integer getReplicas() {
         return replicas;
     }
@@ -124,7 +121,6 @@ public class CreateIndexRequest {
      * The type of pod to use. One of s1, p1, or p2 appended with . and one of x1, x2, x4, or x8.
      * @return podType
      **/
-    @javax.annotation.Nullable
     public String getPodType() {
         return podType;
     }
@@ -160,6 +156,7 @@ public class CreateIndexRequest {
      * The name of the collection to create an index from
      * @return sourceCollection
      **/
+    @javax.annotation.Nullable
     public String getSourceCollection() {
         return sourceCollection;
     }
--- a/src/main/java/io/pinecone/model/CreateIndexRequest.java
+++ b/src/main/java/io/pinecone/model/CreateIndexRequest.java
@@ -25,7 +25,7 @@ public class CreateIndexRequest {
     public CreateIndexRequest() {
     }
 
-    public CreateIndexRequest indexName(String name) {
+    public CreateIndexRequest withIndexName(String name) {
         this.indexName = name;
         return this;
     }
@@ -34,7 +34,7 @@ public class CreateIndexRequest {
      * The name of the index to be created. The maximum length is 45 characters.
      * @return name
      **/
-    @javax.annotation.Nullable
+    @javax.annotation.Nonnull
     public String getIndexName() {
         return indexName;
     }
@@ -43,7 +43,7 @@ public class CreateIndexRequest {
         this.indexName = indexName;
     }
 
-    public CreateIndexRequest dimension(Integer dimension) {
+    public CreateIndexRequest withDimension(Integer dimension) {
         this.dimension = dimension;
         return this;
     }
@@ -61,13 +61,13 @@ public class CreateIndexRequest {
         this.dimension = dimension;
     }
 
-    public CreateIndexRequest metric(String metric) {
+    public CreateIndexRequest withMetric(String metric) {
         this.metric = metric;
         return this;
     }
 
     /**
-     * The distance metric to be used for similarity search. You can use &#39;euclidean&#39;, &#39;cosine&#39;, or &#39;dotproduct&#39;.
+     * The distance metric to be used for similarity search. You can use 'euclidean', 'cosine', or 'dotproduct'.
      * @return metric
      **/
     @javax.annotation.Nullable
@@ -79,7 +79,7 @@ public class CreateIndexRequest {
         this.metric = metric;
     }
 
-    public CreateIndexRequest pods(Integer pods) {
+    public CreateIndexRequest withPods(Integer pods) {
         this.pods = pods;
         return this;
     }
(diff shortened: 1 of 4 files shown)


src/main/java/io/pinecone/model/CreateIndexRequest.java
```diff
@@ -0,0 +1,197 @@
...
+        this.dimension = dimension;
+    }
+
+    public CreateIndexRequest metric(String metric) {
+        this.metric = metric;
+        return this;
+    }
+
+    /**
+     * The distance metric to be used for similarity search. You can use &#39;euclidean&#39;, &#39;cosine&#39;, or &#39;dotproduct&#39;.
```
jhamon, line 70: Seems like these quotation marks got messed up during a copy/paste. Let's change them to normal `'` marks.

src/main/java/io/pinecone/model/CreateIndexRequest.java
```diff
@@ -0,0 +1,197 @@
...
+    public CreateIndexRequest indexName(String name) {
+        this.indexName = name;
+        return this;
+    }
+
+    /**
+     * The name of the index to be created. The maximum length is 45 characters.
+     * @return name
+     **/
+    @javax.annotation.Nullable
```
jhamon, line 37: What's the purpose of this annotation? We don't ever want to create an index wtihout a name, so it should not be Nullable. Maybe you meant `Nonnull`?

src/main/java/io/pinecone/model/CreateIndexRequest.java
```diff
@@ -0,0 +1,197 @@
...
+    public CreateIndexRequest metric(String metric) {
+        this.metric = metric;
+        return this;
+    }
+
+    /**
+     * The distance metric to be used for similarity search. You can use &#39;euclidean&#39;, &#39;cosine&#39;, or &#39;dotproduct&#39;.
+     * @return metric
+     **/
+    @javax.annotation.Nullable
```
jhamon, line 73: Same question on this as above for index name. Should this be Nullable?

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -53,12 +58,113 @@ public void testIndexCreation() throws IOException {
         when(mockResponseFuture.toCompletableFuture()).thenReturn(mockCompletableFuture);
 
         PineconeIndexOperationClient indexDeletionService = new PineconeIndexOperationClient(clientConfig, mockClient);
-        indexDeletionService.createIndex("testEnvironment", 128, "cosine");
+        CreateIndexRequest createIndexRequest = new CreateIndexRequest().indexName("test_name").dimension(3);
```
jhamon, line 61: What do you think about changing these builder methods to have the `with` prefix? E.g. `new CreateIndexRequest().withIndexName("my_index").withDimension(5);`

That would make them more consistent with the approach on the `PineconeClientConfig` class.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -53,12 +58,113 @@ public void testIndexCreation() throws IOException {
...
+        CompletableFuture<Response> mockCompletableFuture = mock(CompletableFuture.class);
+
+        when(mockResponse.getStatusCode()).thenReturn(201);
+        when(mockClient.prepare(anyString(), anyString())).thenReturn(mockBoundRequestBuilder);
+        when(mockBoundRequestBuilder.setHeader(anyString(), anyString())).thenReturn(mockBoundRequestBuilder);
+        when(mockBoundRequestBuilder.setBody(anyString())).thenReturn(mockBoundRequestBuilder);
+        when(mockBoundRequestBuilder.execute()).thenReturn(mockResponseFuture);
+        when(mockResponseFuture.toCompletableFuture()).thenReturn(mockCompletableFuture);
+
+        PineconeIndexOperationClient indexDeletionService = new PineconeIndexOperationClient(clientConfig, mockClient);
```
jhamon, line 87: I think this variable name should not have "deletion" in it since the class does more things now. I'd say it's just the `indexService` now.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -53,12 +58,113 @@ public void testIndexCreation() throws IOException {
...
+                .podType("p1.x2")
+                .replicas(2)
+                .metadataConfig(metadataConfig)
+                .sourceCollection("step");
+
+        try{
+            indexDeletionService.createIndex(createIndexRequest);
+        }
+        catch (IOException e) {
+
```
jhamon, line 167: Should this be a `fail()` scenario? When would this occur?

jhamon (approved):
Left a lot of small comments, but nothing super critical except maybe clarifying the way annotations are being used. 

For the most part it looks good. Tighten up a few of these details and then feel free to merge.

src/main/java/io/pinecone/model/CreateIndexRequest.java
rohanshah18: Since metric is already set to cosine by default, the nullable is expected. In the api docs, only index name and dimension are set to be required fields.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: Yes, I thought about the same before naming it the way it is but since I've started to make use of openAPI, I leaned towards their (openAPI's) naming convention. Although I agree with your point of being consistent with what is in the PineconeClientConfig already, so I'll fix the naming.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: Thank you for pointing it out! Will fix it.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: Did you mean the IOException? Or testing create index if all fields are passed?
If you meant:
1. IOException, then the createIndex method is supposed to throw it, so I've ignored the catching of it. 
2. If creating a new index with all fields test case is required or not, then you're correct it'll not fail right now. In future, I'll like to have more logic included in the validateJsonObject() part of CreateIndexRequest class. I'll remove it for now though.

src/main/java/io/pinecone/model/CreateIndexRequest.java
rohanshah18: Thanks for pointing it out, will fix it soon!

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
jhamon: Yeah, it was the IOException I was wondering about since you catch it and then do nothing. If that's intended, some kind of comment would be nice to explain why it's expected.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
jhamon: I don't feel super strongly about it, I leave it up to you. I can see pros and cons to both ways of doing it.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: Will update this, thanks.

PR #17, opened 2023-08-22
Update changelog for 0.3.0 release

rohanshah18:
## Problem

The changelog.md file needs to be updated for next release.

## Solution

Updated the changelog.md file with latest changes.

## Type of Change

## Test Plan

No testing is required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 [comment]: <> (When bumping [pc:VERSION_LATEST_RELEASE] create a new entry below)
 
-### v0.2.4
-- Added support for index operations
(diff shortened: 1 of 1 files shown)


jhamon (approved):
Looks good.

Going forward I think we can add running notes to the `CHANGELOG.md` in a section called **Unreleased changes** or something instead of trying to guess what the next version will be (since we don't know what the next version number will be until we are ready to cut a release). And then in the pre-release commit where we bump the version number we can adjust the title to reflect the new version number.

The way this was before you change made it look like there was a 0.2.4 release but no such release exists. That's the danger of guessing the version number in advance of the release.

PR #18, opened 2023-08-24
Parse connection URL string to autofill indexName, environment, and projectName

rohanshah18:
## Problem
Connection url contains indexName, environment, and projectName but currently these values are manually set by the users.

## Solution
To reduce user errors, parse the connection url string to get indexName, environment, and projectName and add them to the configurations. 

There are two config classes, PineconeClientConfig (will refer as clientConfig) and PineconeConnectionConfig (will refer as connectionConfig). The clientConfig contains environment and projectName, and connectionConfig contains indexName, so there were two approaches possible to set these values:
1. Delete connectionConfig class completely and add the members (indexName and customChannelBuilder for custom GRPC calls) in the clientConfig. This approach would have made things cleaner since the end user will only have to define one config object and use it to instantiate the PineconeClient class but the downside was in order for it to be backwards compatible, we'll have to maintain the connectionConfig for a while and secondly, the clientConfig would be bigger and harder to maintain over time.
2. Create PineconeConfigs class which has clientConfig and connectionConfig. Parse the connection url string as a part of the PineconeConfigs constructor that accepts apiKey and url, and set the values of clientConfig (apiKey, environment, and projectName) as well as of the connectionConfig (indexName). If the user is now interested in setting serverSideTimeouts to clientConfig or use customChannelBuilder of connectionConfig, they can do so by extracting the config objects and use there respective .with(). 

Below is the sample of how this config will be consumed:
PineconeConfigs configs = new PineconeConfigs("apiKey", "url");
        PineconeClient pineconeClient = new PineconeClient(configs.getClientConfig());
        PineconeConnection connection = pineconeClient.connect(configs.getConnectionConfig());
        connection.doVectorOperations()

Also, note that this can be improved even further as shown below by adding an overloaded constructor in Pineconeclient class as shown below but in order for the user to set serverSideTimeout or use the custom GRPC channel, they'll have to extract the configs from PineconeConfigs. So the code below can be only helpful if the user is fine with the default serverSideTimeout and GRPC channel:
PineconeClient client = new PineconeClient();
PineconeConnection connection = client.connect("apiKey", "url");
        
        
## Type of Change

## Test Plan
A series of unit tests have been added. The follow up subtask to this ticket is integration test.

### DIFF
--- a/src/main/java/io/pinecone/PineconeConfigs.java
+++ b/src/main/java/io/pinecone/PineconeConfigs.java
@@ -18,7 +18,7 @@ public class PineconeConfigs {
     }
 
     public PineconeConfigs(String apiKey, String connectionURL) {
-        String pattern = "https://([a-zA-Z0-9-]{1,45})-([a-fA-F0-9]+)\\.svc\\.([a-zA-Z0-9]+)\\.pinecone\\.io";
+        String pattern = "https://([a-zA-Z0-9-]{1,45})-([a-fA-F0-9]+)\\.svc\\.([a-zA-Z0-9-]+)\\.pinecone\\.io";
         Pattern regexPattern = Pattern.compile(pattern);
         Matcher matcher = regexPattern.matcher(connectionURL);
 
--- a/src/main/java/io/pinecone/PineconeConfigs.java
+++ b/src/main/java/io/pinecone/PineconeConfigs.java
@@ -7,9 +7,6 @@ public class PineconeConfigs {
     private PineconeClientConfig clientConfig;
     private PineconeConnectionConfig connectionConfig;
 
-    public PineconeConfigs() {
-    }
-
     protected  PineconeConfigs(PineconeConfigs other) {
         clientConfig = other.clientConfig;
         connectionConfig = other.connectionConfig;
@@ -21,7 +18,6 @@ public class PineconeConfigs {
     }
 
     public PineconeConfigs(String apiKey, String connectionURL) {
-        this();
         String pattern = "https://([a-zA-Z0-9-]{1,45})-([a-fA-F0-9]+)\\.svc\\.([a-zA-Z0-9]+)\\.pinecone\\.io";
         Pattern regexPattern = Pattern.compile(pattern);
         Matcher matcher = regexPattern.matcher(connectionURL);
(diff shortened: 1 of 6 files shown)


src/main/java/io/pinecone/PineconeConfigs.java
```diff
@@ -0,0 +1,53 @@
...
+
+        if (matcher.matches()) {
+            String indexName = matcher.group(1);
+            String projectName = matcher.group(2);
+            String environment = matcher.group(3);
+            clientConfig = clientConfig.withApiKey(apiKey).withProjectName(projectName).withEnvironment(environment);
+            connectionConfig = connectionConfig.withIndexName(indexName);
+        }
+        else {
+            throw new PineconeValidationException("Unable to parse connection url");
```
rohanshah18, line 34: Throwing exception if the url wasn't parsable but please feel free to comment on if this user flow doesn't look good. Another option is to return a new PineconeClientConfig with new PineconeClientConfig and PineconeConnectionConfig but the values (apiKey, env, projectName, and indexName) will not be set. So it'll throw invalidException.

src/main/java/io/pinecone/PineconeConfigs.java
```diff
@@ -0,0 +1,60 @@
+package io.pinecone;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class PineconeConfigs {
+    private PineconeClientConfig clientConfig;
+    private PineconeConnectionConfig connectionConfig;
+
+    public PineconeConfigs() {
```
rohanshah18, line 10: I can initialize both config objects as a part of default constructor but if the 3 values are not passed, there's no need for new config objects since the user will instantiate both configs using .withClientConfig and .withConnectionConfig methods. But please validate this logic.

jhamon:
I have another meeting starting so I can't go through this line by line to give final approval just yet, but I'm happy to see you adding a good number of unit tests. For something like this, reliability is the most important thing so we should be very thorough in testing this.

jhamon (approved):
Approving based on our in-person group review today

austin-denoble (approved):
Nice work here, I appreciate the detailed walkthrough you gave us earlier around the approach. Overall this looks good to me, I think for now wrapping things inside `PineconeConfigs` makes a lot of sense.

Approving, I'm also going to walk through the steps to get things setup locally to validate the approach there.

src/main/java/io/pinecone/PineconeConfigs.java
```diff
@@ -17,20 +17,20 @@ public PineconeConfigs(PineconeClientConfig clientConfig, PineconeConnectionConf
...
             String indexName = matcher.group(1);
             String projectName = matcher.group(2);
             String environment = matcher.group(3);
             clientConfig = new PineconeClientConfig().withApiKey(apiKey).withProjectName(projectName).withEnvironment(environment);
-            connectionConfig = new PineconeConnectionConfig().withIndexName(indexName);
+            connectionConfig = new PineconeConnectionConfig().withIndexName(indexName).withConnectionUrl(connectionUrl);
         }
         else {
-            throw new PineconeValidationException("Unable to parse connection url");
+            throw new PineconeValidationException("Unable to parse connection url " + connectionUrl);
```
jhamon, line 33: I think we're missing an if-else condition where we check for the v4 url pattern. In that case we want to save the connecutionUrl and not throw an error here.

PR #19, opened 2023-08-29
Add support for connection url for data plane operations only

rohanshah18:
## Problem
Connection url contains indexName, environment, and projectName but currently these values are manually set by the users in configuration, so the use the connection url string instead of asking for 3 different variables for data plane operations only.

## Solution
To reduce user errors, parse the connection url string to get indexName, environment, and projectName and add them to the configurations. The grpc calls for data plane were using these variables for reconstructing the endpoint again so added an if-else loop where if a valid connection url is found, it'll be used instead of reconstructing the endpoint. Although, if no connection url is passed, it'll be constructing the endpoint using the three variables mentioned before which is done as of today.

## Type of Change

## Test Plan
Created unit tests and performed manual integration tests locally.

### DIFF
--- a/src/main/java/io/pinecone/PineconeClient.java
+++ b/src/main/java/io/pinecone/PineconeClient.java
@@ -37,8 +37,8 @@ public final class PineconeClient {
         return new PineconeConnection(config, connectionConfig);
     }
 
-    public PineconeConnection connect(String apiKey, String connectionUrl) {
-        return new PineconeConnection(new PineconeClientConfig().withApiKey(apiKey),
-                new PineconeConnectionConfig().withConnectionUrl(connectionUrl));
+    public PineconeConnection connectWithUrl(String connectionUrl) {
+        return connect(new PineconeConnectionConfig()
+                .withConnectionUrl(connectionUrl));
     }
 }
--- a/src/main/java/io/pinecone/PineconeClient.java
+++ b/src/main/java/io/pinecone/PineconeClient.java
@@ -36,4 +36,9 @@ public final class PineconeClient {
     public PineconeConnection connect(PineconeConnectionConfig connectionConfig) {
         return new PineconeConnection(config, connectionConfig);
     }
+
+    public PineconeConnection connect(String apiKey, String connectionUrl) {
+        return new PineconeConnection(new PineconeClientConfig().withApiKey(apiKey),
+                new PineconeConnectionConfig().withConnectionUrl(connectionUrl));
+    }
 }
--- a/src/main/java/io/pinecone/PineconeClientConfig.java
+++ b/src/main/java/io/pinecone/PineconeClientConfig.java
@@ -100,10 +100,6 @@ public class PineconeClientConfig {
     void validate() {
         if (apiKey == null)
             throw new PineconeValidationException("Invalid Pinecone config: missing apiKey");
-        if (environment == null)
-            throw new PineconeValidationException("Invalid Pinecone config: missing environment");
-        if (projectName == null)
-            throw new PineconeValidationException("Invalid Pinecone config: missing projectName ");
     }
 
     @Override
--- a/src/main/java/io/pinecone/PineconeConnection.java
+++ b/src/main/java/io/pinecone/PineconeConnection.java
@@ -35,9 +35,10 @@ public class PineconeConnection implements AutoCloseable {
     private VectorServiceGrpc.VectorServiceStub asyncStub;
 
     public PineconeConnection(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig) {
-        connectionConfig.validate();
         this.connectionConfig = connectionConfig;
         this.clientConfig = clientConfig;
+        validateConfigs();
+
         channel = connectionConfig.getCustomChannelBuilder() != null
                 ? connectionConfig.getCustomChannelBuilder().apply(clientConfig, connectionConfig)
                 : buildChannel(clientConfig, connectionConfig);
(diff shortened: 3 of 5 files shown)


src/main/java/io/pinecone/PineconeClientConfig.java
```diff
@@ -100,10 +100,6 @@ public PineconeClientConfig withServerSideTimeoutSec(int serverSideTimeout) {
     void validate() {
         if (apiKey == null)
             throw new PineconeValidationException("Invalid Pinecone config: missing apiKey");
-        if (environment == null)
```
rohanshah18, line 103: Removed these conditions from validate() since environment and projectName will be empty when connection url is used for data plane operations.

src/main/java/io/pinecone/PineconeConnection.java
```diff
@@ -113,15 +112,16 @@ private VectorServiceGrpc.VectorServiceBlockingStub applyDefaultBlockingStubConf
                 .withMaxOutboundMessageSize(DEFAULT_MAX_MESSAGE_SIZE);
     }
 
-    private static String getEndpoint(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig) {
-        String endpoint = String.format("%s-%s.svc.%s.pinecone.io",
-                connectionConfig.getIndexName(),
-                clientConfig.getProjectName(),
-                clientConfig.getEnvironment());
+    static String getEndpoint(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig) {
+        String endpoint = (connectionConfig.getConnectionUrl() != null) ?
```
rohanshah18, line 116: Use connection url when entered by the user, otherwise follow the existing flow of building it using environment, projectName, and indexName.

src/main/java/io/pinecone/PineconeClient.java
```diff
@@ -36,4 +36,9 @@ public PineconeConnection connect(String indexName) {
     public PineconeConnection connect(PineconeConnectionConfig connectionConfig) {
         return new PineconeConnection(config, connectionConfig);
     }
+
+    public PineconeConnection connectWithUrl(String connectionUrl) {
```
rohanshah18, line 40: Approach 1: 
connectWithUrl(String apiKey, String connectionUrl) {
// Initialize clientConfig.withApiKey()
// Initialize connection.withConnectionUrl()
}

Approach 1 requires user to enter api key for every new indexes which doesn't make sense since apiKey doesnt change for each connectionUrl/indexes. Rather, use the existing clientConfig with apiKey and initialize a new PineconeConnection object which internally initializes PineconeConfigClient with connectionUrl. Please let me know your thoughts on this if you've a better approach or different opinion.

src/main/java/io/pinecone/PineconeClient.java
```diff
@@ -36,4 +36,9 @@ public PineconeConnection connect(String indexName) {
     public PineconeConnection connect(PineconeConnectionConfig connectionConfig) {
         return new PineconeConnection(config, connectionConfig);
```
jhamon, line 37: I see we call `config.validate()` in the constructor, but that validation is now minimal (asserting API key only). And since these are all mutable objects, there is a risk that config state could change after a `PineconeConnection` object is instantiated. We can address both of these concerns by doing some validation at the moment these fields are about to be used, i.e. in the `connect()` method. Otherwise we run the risk of uncaught NullPointerExceptions.

We removed the parts of the `PineconeClientConfig`'s `validate()` that assert environment and projectId must be set, since there are now valid situations where that is not that case. But from this PineconeConnection class, we have all the information to know whether we are in that situation or not. So we could add something like this:

```java
void validateConfigs() throws PineconeValidationError {
  if (this.config == null) {
    throw new PineconeValidationError('PineconeClientConfiguration may not be null')
  }
  if (this.connectionConfig == null) {
    throw new PineconeValidationError('PineconeConnectionConfig may not be null')
  }

  config.validate(); // always make sure that API key is still present 

  if (connectionConfig.indexName != null) {
      if (this.config.getEnvironment() == null || this.config.getProjectName() == null) {
        throw new PineconeValidationError('Cannot connect with indexName ' + indexName + ' unless PineconeClientConfig contains projectName and environment')
      }
  }
}
```

and call it from within `connect()`

jhamon (approved):

src/main/java/io/pinecone/PineconeClient.java
austin-denoble: Agreed, this feels like it makes sense here. 

**nit:** Also wondering if we should be adding a little documentation type comment here similar to the other `connect()` methods.

austin-denoble (approved):
Looks good to me, I think @jhamon's point around improving some of the validation of needed fields at the time of calling `connect()` would be ideal.

I'm also curious if it makes sense to add / update any of the existing `/java/pineconeexamples/*` files to demonstrate the added functionality, but maybe that's not necessary quite yet.

src/main/java/io/pinecone/PineconeConnectionConfig.java
```diff
@@ -54,15 +67,16 @@ public PineconeConnectionConfig withCustomChannelBuilder(BiFunction<PineconeClie
 
     void validate() {
         String messagePrefix = "Invalid Pinecone config: ";
-        if (indexName == null)
-            throw new PineconeValidationException(messagePrefix + "indexName must be specified");
+        if (indexName == null && connectionUrl == null)
+            throw new PineconeValidationException(messagePrefix + "indexName or connection url must be specified");
```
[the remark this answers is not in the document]
rohanshah18: Fixed!

rohanshah18:
> Looks good to me, I think @jhamon's point around improving some of the validation of needed fields at the time of calling `connect()` would be ideal.
> 
> I'm also curious if it makes sense to add / update any of the existing `/java/pineconeexamples/*` files to demonstrate the added functionality, but maybe that's not necessary quite yet.



> Looks good to me, I think @jhamon's point around improving some of the validation of needed fields at the time of calling `connect()` would be ideal.
> 
> I'm also curious if it makes sense to add / update any of the existing `/java/pineconeexamples/*` files to demonstrate the added functionality, but maybe that's not necessary quite yet.

Yes, my plan is to have those examples updated and we have a follow up ticket. Thanks for pointing out.

src/main/java/io/pinecone/PineconeConnectionConfig.java
```diff
@@ -42,6 +45,16 @@ public PineconeConnectionConfig withIndexName(String indexName) {
...
+    public String getConnectionUrl() {
+        return connectionUrl;
+    }
+
+    public PineconeConnectionConfig withConnectionUrl(String connectionUrl) {
+        PineconeConnectionConfig config = new PineconeConnectionConfig(this);
+        config.connectionUrl = connectionUrl;
+        return config;
+    }
+
```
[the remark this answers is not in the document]
rohanshah18: A lot of these comments are auto generated, so something on wish list where I'll like to clean up and add relevant comments throughout the sdk. Thanks for pointing it out!!

PR #23, opened 2023-09-18
Update http client and improve error handling

rohanshah18:
## Problem

The java sdk currently relies on AsyncHttpClient but keeping the long term goal in mind, will like to replace it with OkHttpClient. Also the errors raised are not clear enough for the users.

## Solution

Replaced OkHttpClient for create and delete index functionalities. Also added httpErrorMapper which will provide details on why the exception occurred rather than throwing a high level IOException.

## Type of Change

This is neither a bug fix nor a new feature but knowing that in future we'll like to rely on OkHttpClient, it is best to replace it now before implementing rest of the control plane operations.

## Test Plan

Added unit tests and ran integration locally to make sure the client works as it used to before and is backwards compatible.


src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -1,52 +1,67 @@
...
-        System.out.println("Sending delete index request:");
-        client.prepare("DELETE", url + indexName)
-                .setHeader("accept", "text/plain")
-                .setHeader("Api-Key", clientConfig.getApiKey())
-                .execute()
-                .toCompletableFuture()
-                .join();
-
-        client.close();
+        Request request = new Request.Builder()
```
rohanshah18, line 26: I'd a choice to chose between async vs sync calls. The older implementation used to use async calls but the join() was called right after the completableFuture() which I believe is similar to sync calls since it blocks and waits for the completion of the asynchronous request. So I decided to go with sync calls to keep things simpler.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -1,52 +1,67 @@
...
+                .addHeader(ACCEPT_HEADER, TEXT_PLAIN)
+                .addHeader(API_KEY, clientConfig.getApiKey())
+                .build();
+
+        try (Response response = client.newCall(request).execute()) {
+            if (!response.isSuccessful()) {
+                throw new IOException(response.message());
+            }
+        } finally {
+            close(client);
```
rohanshah18, line 38: I don't like adding this close(client) here but keeping backwards compatibility in mind i.e. the older implementation used to close client within each method after the call was executed, I added it here. If I remove the close() call, it'll shift responsibility to the user to close the client after the usage.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -1,52 +1,77 @@
...
     public PineconeIndexOperationClient(PineconeClientConfig clientConfig) {
-        this.clientConfig = clientConfig;
-        client = new DefaultAsyncHttpClient();
-        this.url = "https://controller." + clientConfig.getEnvironment() + ".pinecone.io/databases/";
+        this(clientConfig, new OkHttpClient());
+    }
+
+    private static String createUrl(PineconeClientConfig clientConfig) {
+        if (clientConfig.getApiKey() == null || clientConfig.getEnvironment() == null) {
+            throw new PineconeValidationException("Both API key and environment name are required for index operations.");
```
rohanshah18, line 29: The validation is required to prevent null pointer exception since if PineconeClientConfig is created without either apikey or environment, then the index operations cannot happen

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -1,52 +1,77 @@
...
 
+        Request request = new Request.Builder()
+                .url(url)
+                .post(requestBody)
+                .addHeader(ACCEPT_HEADER, TEXT_PLAIN)
+                .addHeader(CONTENT_TYPE, CONTENT_TYPE_JSON)
+                .addHeader(API_KEY, clientConfig.getApiKey())
+                .build();
 
+        try (Response response = client.newCall(request).execute()) {
```
rohanshah18, line 64: Note that the way I've designed currently is if the response is unsuccessful, i.e. if the index already exists or some invalid parameters caused server side exception, then the client will result in IOException with the cause of the message. Please let me know if you prefer raising PineconeException rather than IOException (same applies to delete functionality).

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -1,52 +1,77 @@
...
+            if (!response.isSuccessful()) {
+                throw new IOException(response.message());
+            }
+        } finally {
+            close(client);
+        }
+    }
+
+    public void close(OkHttpClient client) {
```
jhamon, line 73: From the OkHttpClient docs

> Shutdown the dispatcher’s executor service with [shutdown()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#shutdown--). This will also cause future calls to the client to be rejected.

This is another reason we shouldn't automatically call close(). I haven't done hands-on testing, but my reading of this makes me think the way we are calling close after every operation will render the http client unable to be used again. So you won't be able to do two or more operations in a row without creating a new instance of `PineconeIndexOperationClient` (to get a new okHttpClient instance) which is definitely not what we want.

src/main/java/io/pinecone/utils/Constants.java
```diff
@@ -0,0 +1,11 @@
+package io.pinecone.utils;
+
+public class Constants {
+    public static final String ACCEPT_HEADER = "accept";
+    public static final String API_KEY = "Api-Key";
```
jhamon, line 5:
```suggestion
    public static final String API_KEY_HEADER_NAME = "Api-Key";
```

src/main/java/io/pinecone/utils/Constants.java
```diff
@@ -0,0 +1,11 @@
+package io.pinecone.utils;
+
+public class Constants {
```
jhamon, line 3: IMO these constants should all probably live with the `PineconeIndexOperationClient`. I don't see any benefit or removing these values from that context and putting them in a different class because they're not being used elsewhere. `Constants` is kind of a vague term that over time can end up collecting a lot of unrelated stuff in it.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
jhamon: Definitely `IOException` is too generic. I think the more specific we can make these exceptions the better. Take a look at how I'm approaching this in the [ts client](https://github.com/pinecone-io/pinecone-ts-client/blob/main/src/errors/http.ts#L96): 

```
export const mapHttpStatusError = (failedRequestInfo: FailedRequestInfo) => {
  switch (failedRequestInfo.status) {
    case 400:
      return new PineconeBadRequestError(failedRequestInfo);
    case 401:
      return new PineconeAuthorizationError(failedRequestInfo);
    case 404:
      return new PineconeNotFoundError(failedRequestInfo);
    case 500:
      return new PineconeInternalServerError(failedRequestInfo);
    case 501:
      return new PineconeNotImplementedError(failedRequestInfo);
    default:
      throw new PineconeUnmappedHttpError(failedRequestInfo);
  }
};
```

My recommendation would be to create a `HttpErrorMapper` class (or similar name) that knows how to map status codes to the appropriate error classes. Putting this logic into a different class allows us to have a nicer experience without duplicating logic for errors that can happen anywhere (e.g. 401 Authorization errors). And also greatly simplifies documentation when people come asking about what errors can occur.

jhamon (approved):
In general I like the direction, `OkHttpClient` seems quite nice. But I do think we need some more iteration on this before merging (see comments below). I'm going to mark this as "Approve" so you can not be blocked while we're out of town if you make the requested changes and feel confident merging. But if you want another round of reviews, let us know.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: Great, that was the original thought of having the close() in the class and not calling it at the end of every method. I agree with all of the points you've made, so I'll remove the close(client) calls from each method and leave the close() as a part of the class.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: Thanks for providing me with the example, adding a mapper is the ideal goal, so I'll implement this change here.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: Yes, that's the reason why I didn't like it in the first place but this will be removed.

src/main/java/io/pinecone/utils/Constants.java
rohanshah18: Good point, I'll move them within PineconeIndexOperationClient class.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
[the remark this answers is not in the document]
rohanshah18: I like `PineconeConfigurationException` name better, I'll add this change as a part of this PR.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -1,52 +1,88 @@
...
-                .setHeader("accept", "text/plain")
-                .setHeader("Api-Key", clientConfig.getApiKey())
-                .execute()
-                .toCompletableFuture()
-                .join();
+        Request request = new Request.Builder()
+                .url(url + indexName)
+                .delete()
+                .addHeader(ACCEPT_HEADER, TEXT_PLAIN)
+                .addHeader(API_KEY_HEADER_NAME, clientConfig.getApiKey())
```
jhamon, line 50: It seems like a bummer that you end up writing a lot of similar boilerplate for each request. What do you think about making a helper method that encapsulates some of these steps that happen in every request we make?

Something like this (??? is a placeholder for the return type, which I can't know without some help from the docs or IDE)

```
private ??? initRequest(String path) {
  return new Request.Builder()
         .url(url + path)
         .addHeader(ACCEPT_HEADER, TEXT_PLAIN)
         .addHeader(API_KEY_HEADER_NAME, clientConfig.getApiKey())
}
```

and then usage would be something like

```
Request request = initRequest(indexName)
                                   .delete()
                                   .build()
```

src/main/java/io/pinecone/exceptions/HttpErrorMapper.java
```diff
@@ -0,0 +1,22 @@
...
+        int statusCode = failedRequestInfo.getStatus();
+        switch (statusCode) {
+            case 400:
+                throw new PineconeBadRequestException(failedRequestInfo.getMessage());
+            case 401:
+                throw new PineconeAuthorizationException(failedRequestInfo.getMessage());
+            case 404:
+                throw new PineconeNotFoundException(failedRequestInfo.getMessage());
+            case 409:
+                throw new PineconeAlreadyIndexExistsException(failedRequestInfo.getMessage());
```
jhamon, line 15:
```suggestion
                throw new PineconeConflictException(failedRequestInfo.getMessage());
```

I think a 409 error could also result from trying to create other named resources such as Collections (and maybe more in the future. So we shouldn't put `Index` in the exception name IMO.

jhamon (approved):
I left a few more suggestions, but overall it's looking great. Nice work.

src/main/java/io/pinecone/PineconeClientConfig.java
```diff
@@ -17,9 +19,6 @@ public class PineconeClientConfig {
      */
     private String apiKey;
 
-    /**
-     * Required project name
-     */
     private String projectName;
 
     private String environment;
```
austin-denoble, line 24: **nit:** Just curious why we're removing the comment here. Should we maybe add one for `environment`?

src/main/java/io/pinecone/PineconeIndexOperationClient.java
austin-denoble: Makes sense, thanks for the explanation.

src/main/java/io/pinecone/exceptions/PineconeAlreadyIndexExistsException.java
```diff
@@ -0,0 +1,12 @@
+package io.pinecone.exceptions;
+
+public class PineconeAlreadyIndexExistsException extends PineconeException {
+
+    public PineconeAlreadyIndexExistsException(String message) {
```
austin-denoble, line 5: I was going to mention we might want to swap "Already" and "Index" here as a nit. Given @jhamon's feedback above around making this more general to handle other resource conflicts I think that would be best: https://github.com/pinecone-io/pinecone-java-client/pull/23/files#r1333964728

austin-denoble (approved):
Great work here swapping out `AsyncHttpClient` for `OkHttpClient`. Really appreciate your attention to detail and improving our error handling. 🎉

src/main/java/io/pinecone/exceptions/HttpErrorMapper.java
rohanshah18: Agreed!

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: I agree with you completely, so I've analyzed all index and collection api calls, and encapsulated the request builder method to support all data plane operations.

src/main/java/io/pinecone/PineconeClientConfig.java
rohanshah18: This was supposed to removed for a previous PR but we made projectName and environment optional if connection url is present.

src/main/java/io/pinecone/exceptions/PineconeAlreadyIndexExistsException.java
rohanshah18: Yeah Index will be unnecessary going forward, so removed it.

PR #24, opened 2023-09-22
Add support for describeIndex functionality

rohanshah18:
## Problem

The java sdk lacks describeIndex functionality.

## Solution

Added synchronous http call to describe index.

## Type of Change

## Test Plan

Added unit test as well performed integration tests locally.

### DIFF
--- a/src/main/java/io/pinecone/PineconeIndexOperationClient.java
+++ b/src/main/java/io/pinecone/PineconeIndexOperationClient.java
@@ -71,10 +71,10 @@ public class PineconeIndexOperationClient {
         }
     }
 
-    private Request buildRequest(String method, String path, String header, RequestBody requestBody) {
+    private Request buildRequest(String method, String path, String acceptHeader, RequestBody requestBody) {
         Request.Builder builder = new Request.Builder()
                 .url(url + path)
-                .addHeader(ACCEPT_HEADER, header)
+                .addHeader(ACCEPT_HEADER, acceptHeader)
                 .addHeader(API_KEY_HEADER_NAME, clientConfig.getApiKey());
 
         if (HTTP_METHOD_POST.equals(method)) {
--- a/src/main/java/io/pinecone/PineconeIndexOperationClient.java
+++ b/src/main/java/io/pinecone/PineconeIndexOperationClient.java
@@ -4,6 +4,7 @@ import io.pinecone.exceptions.FailedRequestInfo;
 import io.pinecone.exceptions.HttpErrorMapper;
 import io.pinecone.exceptions.PineconeConfigurationException;
 import io.pinecone.model.CreateIndexRequest;
+import io.pinecone.model.IndexMeta;
 import okhttp3.*;
 
 import java.io.IOException;
@@ -18,9 +19,10 @@ public class PineconeIndexOperationClient {
     public static final String BASE_URL_SUFFIX = ".pinecone.io/databases/";
     public static final String CONTENT_TYPE = "content-type";
     public static final String CONTENT_TYPE_JSON = "application/json";
-    public static final String DELETE = "DELETE";
     public static final String EMPTY_RESOURCE_PATH = "";
-    public static final String POST = "POST";
+    public static final String HTTP_METHOD_DELETE = "DELETE";
+    public static final String HTTP_METHOD_GET = "GET";
+    public static final String HTTP_METHOD_POST = "POST";
     public static final String TEXT_PLAIN = "text/plain";
 
     private PineconeIndexOperationClient(PineconeClientConfig clientConfig, OkHttpClient client, String url) {
@@ -46,7 +48,8 @@ public class PineconeIndexOperationClient {
     }
 
     public void deleteIndex(String indexName) throws IOException {
-        try (Response response = client.newCall(buildRequest(DELETE, indexName, TEXT_PLAIN,null)).execute()) {
+        Request request = buildRequest(HTTP_METHOD_DELETE, indexName, TEXT_PLAIN,null);
+        try (Response response = client.newCall(request).execute()) {
             handleResponse(response);
         }
     }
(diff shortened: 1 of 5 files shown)


src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -46,17 +48,26 @@ private static String createUrl(PineconeClientConfig clientConfig) {
     }
 
     public void deleteIndex(String indexName) throws IOException {
-        try (Response response = client.newCall(buildRequest(DELETE, indexName, TEXT_PLAIN,null)).execute()) {
+        Request request = buildRequest(HTTP_METHOD_DELETE, indexName, TEXT_PLAIN,null);
+        try (Response response = client.newCall(request).execute()) {
```
rohanshah18, line 52: Refactored into the variables, functionality is still the same.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -46,17 +48,26 @@ private static String createUrl(PineconeClientConfig clientConfig) {
...
             handleResponse(response);
         }
     }
 
     public void createIndex(CreateIndexRequest createIndexRequest) throws IOException {
         MediaType mediaType = MediaType.parse("application/json; charset=utf-8");
         RequestBody requestBody = RequestBody.create(createIndexRequest.toJson(), mediaType);
+        Request request = buildRequest(HTTP_METHOD_POST, EMPTY_RESOURCE_PATH, TEXT_PLAIN, requestBody);
```
rohanshah18, line 60: Refactored into variables, functionality is still the same

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -46,17 +48,26 @@ private static String createUrl(PineconeClientConfig clientConfig) {
...
+        try (Response response = client.newCall(request).execute()) {
+            handleResponse(response);
+        }
+    }
 
-        try (Response response = client.newCall(buildRequest(POST, EMPTY_RESOURCE_PATH, TEXT_PLAIN, requestBody)).execute()) {
+    public IndexMeta describeIndex(String indexName) throws IOException {
```
rohanshah18, line 66: Returning a user defined object "IndexMeta" as we've discussed in the past that it is preferable to do so.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -66,10 +77,10 @@ private Request buildRequest(String method, String path, String header, RequestB
                 .addHeader(ACCEPT_HEADER, header)
```
austin-denoble, line 77: **nitpick / question:** What do you think about updating the `buildRequest()` signature to rename `header` to something more specific like `acceptsHeader`?

Very minor but given that we're setting multiple headers when building the request, I feel like it might be nice to be specific.

src/main/java/io/pinecone/model/IndexMeta.java
```diff
@@ -0,0 +1,47 @@
+package io.pinecone.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class IndexMeta {
+    @JsonProperty("database")
+    private IndexMetaDatabase metaDatabase;
```
austin-denoble, line 9: **nit:** Maybe worth just calling this `database`? I think that's consistent with what we do in the node client as well. Just feels a little confusing to me at first glance. We're already in a class called `IndexMeta` so maybe just `database` is sufficient. We may not need the `@JsonProperty()` transform either.

src/main/java/io/pinecone/model/IndexStatus.java
```diff
@@ -0,0 +1,63 @@
+package io.pinecone.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
austin-denoble, line 5: **question:** Is this needed on `IndexMetaDatabase` and `IndexMeta` as well?

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -170,4 +170,56 @@ public void testCreateIndexWithAllFields() throws IOException {
...
+                .withReady(true)
+                .withPort("433");
+
+
+        IndexMeta expectedIndexMeta = new IndexMeta()
+                .withMetaDatabase(metaDatabase)
+                .withStatus(indexStatus);
+
+        expectedIndexMeta.withMetaDatabase(metaDatabase);
+
```
austin-denoble, line 204: Are we calling `expectedIndexMeta.withMetaDatabase(metaDatabase)` twice here for a specific reason?

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -170,4 +170,56 @@ public void testCreateIndexWithAllFields() throws IOException {
         verify(mockClient, times(1)).newCall(any(Request.class));
         verify(mockCall, times(1)).execute();
     }
+
+    @Test
+    public void testDescribeIndex() throws IOException {
+        String indexName = "testIndex";
+        String indexJsonString = "{\"database\":{\"name\":\"testIndex\",\"metric\":\"cosine\",\"dimension\":3," +
```
austin-denoble, line 177: Did you have to create this by hand? Yikes. 😬

austin-denoble (approved):
Overall this looks good to me, nice work. Left some nitpicks and suggestions but I think this is good to go otherwise. 🚢

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: No, there's no need for second call. Thanks for pointing it out.

src/main/java/io/pinecone/model/IndexStatus.java
rohanshah18: Discussed internally on why it's needed.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
rohanshah18: Discussed internally on slack!!

src/main/java/io/pinecone/PineconeIndexOperationClient.java
rohanshah18: Done!

src/main/java/io/pinecone/model/IndexMeta.java
rohanshah18: Thanks for the feedback, fixed!

PR #34, opened 2023-10-03
Expand CI coverage to Java 17, add integration tests

jhamon:
## Problem

Goals for CI, for all supported versions of Java:
- Ensure project builds
- Run unit tests
- Run integration tests

## Solution

**Behold**: [Build passing](https://github.com/pinecone-io/pinecone-java-client/actions/runs/6407066991) 🎉 

- **Workflow configuration**: The hard part of this diff was getting different java versions, gradle, and build.gradle all working well together.
    - Add a configuration matrix that covers Java 8, 11, 16, and 17. These are all the major versions supported by the setup-java github action at the moment.
    - For each of these Java versions, run with a compatible gradle version. Gradle 6.8 is needed for Java 8/11, but is unable to build more modern versions of Java.
- **Unit tests**: These are run automatically as part of the `gradle build` command. The build command is very complex and covers a large number of things.
- **Integration tests**: 
    - Refactor integration tests into their own package, so they can be enabled but run separately from unit tests. We may want to run unit tests frequently in development on local machines, but only run integration tests in CI. Having a different package (and corresponding gradle sourceSet / configuration) is the easiest way to accomplish this.
    - Setup a new project in app.pinecone.io called [sdk-java-testing](https://app.pinecone.io/organizations/-NF9kuQiPWth9QLPg5NS/projects/us-east-1-aws:e6dddb2/indexes) and gave it its own quota of 5 pods
    - Created an API key and added these secrets to github so the github actions workflow can access values for `PINECONE_API_KEY` and `PINECONE_ENVIRONMENT`
- **Run on every commit** Now that these builds are passing, enable them for all PRs and pushes to main by specifiying additional workflow triggers.

To develop this, I pushed changes to my branch and triggered manual builds using the github CLI: `gh workflow run pr.yml --ref jhamon/ci-iteration`

## Type of Change

- [x] Infrastructure change (CI configs, etc)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,6 +24,7 @@ jobs:
       uses: gradle/gradle-build-action@v2
       with:
         gradle-version: ${{ matrix.gradleVersion }}
+        cache-disabled: true
 
     - name: Setup gradle.properties
       if: matrix.javaVersion == 16 || matrix.javaVersion == 17
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -8,8 +8,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
         javaVersion: [8, 11, 16, 17]
+        os: [ubuntu-latest, windows-latest]
         include:
           - javaVersion: 8
             gradleVersion: 6.8
@@ -42,7 +42,7 @@ jobs:
     - name: Run tests
       run: ./gradlew test
 
-    - name: Build examples
-      run: |
-        cd examples
-        ./../gradlew build
+    # - name: Build examples
+    #   run: |
+    #     cd examples
+    #     ./../gradlew build
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         javaVersion: [8, 11, 16, 17]
         os: [ubuntu-latest]
-        gradleVersion: [6.8, 7.6, 8.3]
+        gradleVersion: [7.3, 8.3]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
@@ -45,7 +45,6 @@ jobs:
         javaVersion: [8, 17]
         os: [ubuntu-latest]
         gradleVersion: [8.3]
-
     steps:
     - uses: actions/checkout@v4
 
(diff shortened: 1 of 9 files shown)


.github/workflows/pr.yml
```diff
@@ -2,27 +2,86 @@ name: Pull Request
...
-        java-version: ${{ matrix.javaVersion}}
+        java-version: ${{ matrix.versions.java}}
         
     - name: Setup Gradle
       uses: gradle/gradle-build-action@v2
       with:
-        gradle-version: 6.8
+        gradle-version: ${{ matrix.versions.gradle }}
+
+    - name: Setup gradle.properties
```
jhamon, line 36: I should check again whether this is still needed. I changed up a bunch of stuff since I added this, but at one point it was needed in order to get java 16+17 building.

.github/workflows/pr.yml
```diff
@@ -2,27 +2,86 @@ name: Pull Request
...
+  
+  integration-test:
+    runs-on: ubuntu-latest
+    needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        versions: [
+            { java: 8, gradle: 6.8 },
```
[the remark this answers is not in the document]
austin-denoble: Agreed, I feel like it gives us a lot of coverage and keeps things a bit more lean for running the actual integration tests. Were only two versions selected to save us a bit of time and not bloat the integration pipeline? Just curious.

.github/workflows/pr.yml
```diff
@@ -2,27 +2,86 @@ name: Pull Request
...
+            { java: 17, gradle: 7.3.1 }
```
rohanshah18, line 59: Optional: Can we add java versions 18, 19, and 20.

src/integration/java/io/pinecone/PineconeClientLiveIntegTest.java
```diff
@@ -156,7 +155,7 @@ public void sanity() {
         //        Query by id example
```
rohanshah18, line 155: Nit: Can we fix the indentation?

rohanshah18:
Thanks for adding integration tests and other CI components. I have left some optional comments but overall it looks good to me.

.github/workflows/pr.yml
jhamon: Let's handle this in a follow-up.

rohanshah18 (approved):

PR #42, opened 2023-10-30
Add support to configure index

rohanshah18:
## Problem

The sdk lacks the ability to configure an index.

## Solution

Added support to configure `podtype` and the `number of replicas` for an index.

## Type of Change

## Test Plan

Added unit and integrations tests.

### DIFF
--- a/src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
+++ b/src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
@@ -1,6 +1,7 @@
 package io.pinecone;
 
 import io.pinecone.helpers.RandomStringBuilder;
+import io.pinecone.model.ConfigureIndexRequest;
 import io.pinecone.model.CreateIndexRequest;
 import io.pinecone.model.IndexMeta;
 import org.junit.jupiter.api.BeforeEach;
@@ -45,6 +46,18 @@ public class PineconeIndexOperationsClientIntegrationTest {
         // List the index
         List<String> indexList = pinecone.listIndexes();
         assert !indexList.isEmpty();
+
+        // Configure the index
+        ConfigureIndexRequest configureIndexRequest = new ConfigureIndexRequest()
+                .withPodType("p1.x2")
+                .withReplicas(3);
+        pinecone.configureIndex(indexName, configureIndexRequest);
+
+        // Get the index description
+        indexMeta = pinecone.describeIndex(indexName);
+        assertNotNull(indexMeta);
+        assertEquals(3, indexMeta.getDatabase().getReplicas());
+        assertEquals("p1.x2", indexMeta.getDatabase().getPodType());
 
         // Cleanup
         pinecone.deleteIndex(indexName);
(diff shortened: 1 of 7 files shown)


src/test/java/io/pinecone/PineconeConnectionTest.java
```diff
@@ -1,8 +1,5 @@
 package io.pinecone;
 
-import io.pinecone.PineconeClientConfig;
```
rohanshah18, line 3: Removed unnecessary imports.

src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
```diff
@@ -46,6 +47,18 @@ public void createAndDelete() throws IOException {
         List<String> indexList = pinecone.listIndexes();
         assert !indexList.isEmpty();
 
+        // Configure the index
```
rohanshah18, line 50: This is integration test where I'm configuring podtype and replicas which are further verified by describeIndex call.

src/test/java/io/pinecone/PineconeIndexOperationClientTest.java
```diff
@@ -265,4 +265,31 @@ public void testListIndexes() throws IOException {
         assertNotNull(emptyIndexList);
         assertTrue(emptyIndexList.isEmpty());
     }
+
+    @Test
+    public void testConfigureIndex() throws IOException {
```
rohanshah18, line 270: Added unit test here

src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
jhamon: In tests like this, I like to assert what the state was before applying a change. Just to make sure my test is actually doing what I think. For example, [here](https://github.com/pinecone-io/pinecone-ts-client/blob/main/src/integration/control/configureIndex.test.ts#L89), in the typescript integration tests.

src/main/java/io/pinecone/PineconeIndexOperationClient.java
```diff
@@ -96,9 +107,11 @@ private Request buildRequest(String method, String path, String acceptHeader, Re
         if (HTTP_METHOD_POST.equals(method)) {
             builder.post(requestBody);
             builder.addHeader(CONTENT_TYPE, CONTENT_TYPE_JSON);
-        } else if (HTTP_METHOD_DELETE.equals(method)) {
+        } else if(HTTP_METHOD_PATCH.equals(method)) {
+            builder.patch(requestBody);
+            builder.addHeader(CONTENT_TYPE, CONTENT_TYPE_JSON);
+        } else if (HTTP_METHOD_DELETE.equals(method))
             builder.delete();
-        }
```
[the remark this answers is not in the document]
rohanshah18: I actually skipped the brackets completely by mistake which is fine for java since there's only one line after the else-if but definitely not consistent, so let me add it back

austin-denoble (approved):
Nice work! Agreed with Jen, would be nice to add some error-case tests if it's not too much trouble.

src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
rohanshah18: Added integration tests with edge cases!

PR #43, opened 2023-11-02
Add user agent

rohanshah18:
## Problem

The sdk currently lacks the ability to track usage

## Solution

Add a User-Agent header to each request. The header includes sdk version and usage context set by the consumer.

## Type of Change

## Test Plan

Ran Integration tests that populates datadog metric (please see the ticket).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
 }
 
 group = 'io.pinecone'
-version = '0.6.0' // [pc:VERSION_NEXT]
+version = pineconeClientVersion // [pc:VERSION_NEXT]
 description = 'The Pinecone.io Java Client'
 
 sourceCompatibility = JavaVersion.VERSION_1_8
--- a/src/main/java/io/pinecone/PineconeClientConfig.java
+++ b/src/main/java/io/pinecone/PineconeClientConfig.java
@@ -142,7 +142,7 @@ public class PineconeClientConfig {
         Properties properties = new Properties();
         String userAgentLanguage = "lang=java";
         String version = (this.getUsageContext() != null) ?
-                userAgentLanguage + "; usageContext= " + this.getUsageContext()
+                userAgentLanguage + "; usageContext=" + this.getUsageContext()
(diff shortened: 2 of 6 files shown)


src/main/java/io/pinecone/PineconeClientConfig.java
```diff
@@ -121,4 +137,20 @@ private String maskedApiKey() {
             return String.join("-", splits);
         }
     }
+
+    public String getUserAgent() {
+        Properties properties = new Properties();
+        String userAgentLanguage = "lang=java";
+        String version = (this.getUsageContext() != null) ?
+                userAgentLanguage + "; " + this.getUsageContext()
```
austin-denoble, line 145: Do we need a trailing `;` after `this.getUsageContext()` here? Is there an expected shape that needs to be adhered to when providing custom context?

build.gradle
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 group = 'io.pinecone'
-version = '0.6.0' // [pc:VERSION_NEXT]
+version = pineconeClientVersion // [pc:VERSION_NEXT]
```
austin-denoble, line 12: For my own learning: this file has implicit access to fields in `gradle.properties`?

austin-denoble (approved):
Nice job, this looks good to me.

Is there any testing we'd want to add around some of this stuff? The generation of the user agent itself is fairly simple, but I'm wondering if we'd want to validate that anywhere.

build.gradle
rohanshah18: Yes!

rohanshah18: Also I tested it by releasing on local and making sure I was getting correct jars, so the versioning works!

src/main/java/io/pinecone/PineconeClientConfig.java
rohanshah18: Yes, its a format we decided to follow. I'll send you the slack thread where we discussed this.

PR #44, opened 2023-11-07
Add integration tests for future stub 

rohanshah18:
## Problem

The sdk currently lacks integration tests for future stub which is used for data plane operations.

## Solution

Added integration test for all data plane operations for future stub. As a part of this change, I have organized the integration test folder better by moving and renaming files:
1. Moved `ConfigureIndexTest.java`:
src/integration/java/io/pinecone/integration/ConfigureIndexTest.java -> src/integration/java/io/pinecone/ControlPlane/ConfigureIndexTest.java
2. Renamed `PineconeIndexOperationsClientIntegrationTest.java` to `CreateAndDeleteIndexTest.java` and moved it from: src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java -> src/integration/java/io/pinecone/ControlPlane/CreateAndDeleteIndexTest.java
3. Renamed `PineconeClientLiveIntegTest.java` to `SyncStubTest.java` and moved it from:
src/integration/java/io/pinecone/PineconeClientLiveIntegTest.java -> 
src/integration/java/io/pinecone/DataPlane/SyncStubTest.java

I have also updated some weird assert calls with the correct assert functionality in SyncStubTest.java so the only files to review are:
1. AsyncStubTest.java
2. SyncStubTest.java

## Type of Change

## Test Plan

Ran integration tests.

### DIFF
--- a/src/integration/java/io/pinecone/ControlPlane/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/ControlPlane/ConfigureIndexTest.java
@@ -1,7 +1,6 @@
 package io.pinecone.ControlPlane;
 
 import io.pinecone.PineconeClientConfig;
-import io.pinecone.DataPlane.SyncStubTest;
 import io.pinecone.PineconeIndexOperationClient;
 import io.pinecone.exceptions.PineconeBadRequestException;
 import io.pinecone.exceptions.PineconeNotFoundException;
@@ -20,7 +19,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 public class ConfigureIndexTest {
     private PineconeIndexOperationClient pinecone;
     private String indexName;
-    private static final Logger logger = LoggerFactory.getLogger(SyncStubTest.class);
+    private static final Logger logger = LoggerFactory.getLogger(ConfigureIndexTest.class);
 
     @BeforeEach
     public void setUp() throws IOException {
--- a/src/integration/java/io/pinecone/ControlPlane/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/ControlPlane/ConfigureIndexTest.java
@@ -1,7 +1,7 @@
-package io.pinecone.integration.ControlPlane;
+package io.pinecone.ControlPlane;
 
 import io.pinecone.PineconeClientConfig;
-import io.pinecone.integration.DataPlane.SyncStubTest;
+import io.pinecone.DataPlane.SyncStubTest;
 import io.pinecone.PineconeIndexOperationClient;
 import io.pinecone.exceptions.PineconeBadRequestException;
 import io.pinecone.exceptions.PineconeNotFoundException;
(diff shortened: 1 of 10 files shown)


src/integration/java/io/pinecone/ControlPlane/ConfigureIndexTest.java
```diff
@@ -20,7 +20,7 @@
 public class ConfigureIndexTest {
     private PineconeIndexOperationClient pinecone;
     private String indexName;
-    private static final Logger logger = LoggerFactory.getLogger(PineconeClientLiveIntegTest.class);
+    private static final Logger logger = LoggerFactory.getLogger(SyncStubTest.class);
```
jhamon, line 23: Should this be `ConfigureIndexTest.class`?

src/integration/java/io/pinecone/DataPlane/FutureStubTest.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Test
+    public void checkIndexSetup() throws Exception {
+        IndexMeta indexMeta = controlPlaneClient.describeIndex(indexName);
+        assertEquals(3, indexMeta.getDatabase().getDimension());
+    }
+
+    @Test
+    public void sanity() throws Exception {
```
jhamon, line 49: I think this test is doing way too much. When a test like this fails, it takes a lot of investigation to figure out what is actually breaking.

Some operations (e.g. create & describe/list, upsert & fetch, upsert & query) make sense to use and test together, but you could still break this down into many smaller test cases.

I would have test cases for:
- Create index (with minimal params) and describe index
- Create index (passing all optional params) and describe index
- Upsert and fetch
- Upsert hybrid and fetch 
- Upsert and fetch by id
- Upsert and query by id
- Upsert and query with vector values
  - with `includeMetadata`
  - with `includeValues` 
- Upsert hybrid and query
- Edge cases (missing params, invalid names, wrong vector dimension, etc)

src/integration/java/io/pinecone/DataPlane/SyncStubTest.java
```diff
@@ -185,7 +187,7 @@ public void sanity() throws Exception {
         conn.getBlockingStub().delete(deleteRequest);
         fetchRequest = FetchRequest.newBuilder().addAllIds(ids).setNamespace(namespace).build();
         fetchResponse = conn.getBlockingStub().fetch(fetchRequest);
-        assert (fetchResponse.getVectorsCount() == ids.size() - 1);
+        assertEquals (fetchResponse.getVectorsCount(), ids.size() - 1);
```
jhamon, line 190:
```suggestion
        assertEquals(fetchResponse.getVectorsCount(), ids.size() - 1);
```

jhamon (approved):
I really think we should split up this integration test into many smaller and more focused tests.

src/integration/java/io/pinecone/DataPlane/FutureStubTest.java
rohanshah18: This is great, thank you for providing me with this list. It has been like this since I picked up and always thought it was okay haha.

PR #47, opened 2023-11-22
Add describeIndexStats integration test

rohanshah18:
## Problem

The java sdk lacks integration test for describeIndexStats functionality

## Solution

Added integration test for both future and sync client to describeIndexStats. Note this test does include Upsert operation but I'll be testing it rigorously in UpsertAndFetch integration test.

## Type of Change

## Test Plan

Ran integration tests

Note: PineconeClientLiveIntegTest will go away once I add all of the integration tests for data plane. PineconeIndexOperationsClientIntegrationTest will be moved under control plane integration test sub-folder and renamed to CreateAndDeleteIndex, but I'll be adding more tests in their if needed.


src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
```diff
@@ -0,0 +1,60 @@
...
+import io.pinecone.proto.Vector;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class BuildUpsertRequest {
+    private static final float[][] upsertData = {{1.0F, 2.0F, 3.0F}, {4.0F, 5.0F, 6.0F}, {7.0F, 8.0F, 9.0F}};
+
+    public static UpsertRequest buildRequiredUpsertRequest() {
```
rohanshah18, line 17: This class has two buildUpsertRequest: One with required parameters only and the second one is with optional parameters

src/integration/java/io/pinecone/helpers/GetOrCreateIndex.java
```diff
@@ -0,0 +1,39 @@
+package io.pinecone.helpers;
+
+import io.pinecone.*;
+import io.pinecone.model.CreateIndexRequest;
+import io.pinecone.model.IndexMeta;
+
+import java.io.IOException;
+import java.util.List;
+
+public class GetOrCreateIndex {
```
rohanshah18, line 10: This is the logic for getting an existing index or creating one. This needs to be updated in future when I rigorously test upsert since chances of having an index of dimension d != dimension y for other upserts is high.

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDescribeIndexStatsTest.java
```diff
@@ -0,0 +1,106 @@
...
+    private static VectorServiceGrpc.VectorServiceBlockingStub blockingStub;
+    private static VectorServiceGrpc.VectorServiceFutureStub futureStub;
+    private static final int dimension = 3;
+
+    @BeforeAll
+    public static void setUp() throws IOException {
+        PineconeConnection connection = getOrCreateIndex(dimension);
+        blockingStub = connection.getBlockingStub();
+        futureStub = connection.getFutureStub();
+    }
```
austin-denoble, line 24: I think it may be good to set up an `@AfterAll` that uses `listIndexes` to handle deleting all indexes that may have been created during testing. We do that in integration tests for TS and I believe Python.

I think in this case given the usage of `getOrCreateIndex()` we expect to only be creating one index for use in this suite, but it may be good to go ahead and clean things up.

src/integration/java/io/pinecone/helpers/GetOrCreateIndex.java
```diff
@@ -0,0 +1,39 @@
...
+    public static PineconeConnection getOrCreateIndex(int dimension) throws IOException {
```
austin-denoble, line 11: Could this helper function also be reused in the `ConfigureIndexTest` file?

austin-denoble (approved):
Overall this looks good to me, nice work. Left some questions and feedback.

I was also curious, are we expecting the existing `PineconeClientLiveIntegTest` and `PineconeIndexOperationsClientIntegrationTest` files are going to be going away as integration tests are added?

It might also be nice to move the existing `ConfigureIndexTest` file into `integration/controlplane` to line up with the new dataplane tests, if that makes sense.

austin-denoble:
Also looks like the tests are failing in CI when validating dimension: `assertEquals(describeIndexStatsResponse.getDimension(), dimension);`

Also looks like `ConfigureIndexTest` is failing when creating an index in Java 17. 🤔 

Not sure exactly why.

src/integration/java/io/pinecone/helpers/GetOrCreateIndex.java
```diff
@@ -0,0 +1,39 @@
...
+
+        PineconeIndexOperationClient controlPlaneClient = new PineconeIndexOperationClient(config);
+
+        List<String> indexList = controlPlaneClient.listIndexes();
+        String requiredParamsIndexName;
+        if(!indexList.isEmpty()) {
+            requiredParamsIndexName = indexList.get(0);
+        }
+        else {
+            requiredParamsIndexName = RandomStringBuilder.build("index-name", 8);
```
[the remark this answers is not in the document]
rohanshah18: Yes!

src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
rohanshah18: Yes I'll need it for future upsert test.

src/integration/java/io/pinecone/helpers/GetOrCreateIndex.java
rohanshah18: Imagine a scenario where index an index with a different dimension d exists but my test is expecting dimension y. So I have to now call describeIndexStats to confirm the dimension.

rohanshah18: Which really makes me wonder if I should create a new index every-time I start a new test so I'm not relying on the previous state.

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDescribeIndexStatsTest.java
rohanshah18: The goal is to reuse the existing index for all data plane integration tests so I'm going to add a cron job to delete all indexes at every 24 hours

src/integration/java/io/pinecone/helpers/GetOrCreateIndex.java
rohanshah18: For any integration test, we would like to make the sub test independent and not reply on the previous sub test. To achieve this, we are going to make use of BeforeEach and setting a unique namespace in data plane but for configure index, we must create a new index.

src/integration/java/io/pinecone/helpers/GetOrCreateIndex.java
rohanshah18: I have updated the logic to check for dimension of the index, otherwise create a new index.

PR #48, opened 2023-11-29
Add integration test for deleting vectors

rohanshah18:
## Problem

The sdk currently lacks integration tests for deleting vectors.

## Solution

Vectors can be deleted in multiple ways, so added an integration tests that covers all of the different possibilities.

## Type of Change

- [X] None of the above: Integration tests

## Test Plan

Ran integration tests

### DIFF
--- a/src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
+++ b/src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
@@ -15,17 +15,22 @@ public class BuildUpsertRequest {
     private static final float[][] upsertData = {{1.0F, 2.0F, 3.0F}, {4.0F, 5.0F, 6.0F}, {7.0F, 8.0F, 9.0F}};
 
     public static UpsertRequest buildRequiredUpsertRequest() {
-        String namespace = RandomStringBuilder.build("ns", 8);
-        List<String> upsertIds = Arrays.asList("v1", "v2", "v3");
-        List<Vector> upsertVectors = new ArrayList<>();
+        return buildRequiredUpsertRequest(new ArrayList<>(), "");
+    }
 
+    public static UpsertRequest buildRequiredUpsertRequest(String namespace) {
+        return buildRequiredUpsertRequest(new ArrayList<>(), namespace);
+    }
+
+    public static UpsertRequest buildRequiredUpsertRequest(List<String> upsertIds, String namespace) {
+        if (upsertIds.isEmpty()) upsertIds = Arrays.asList("v1", "v2", "v3");
+        if (namespace.isEmpty()) namespace = RandomStringBuilder.build("ns", 8);
+
+        List<Vector> upsertVectors = new ArrayList<>();
         for (int i = 0; i < upsertData.length; i++) {
             upsertVectors.add(Vector.newBuilder()
                     .addAllValues(Floats.asList(upsertData[i]))
-                    .setMetadata(Struct.newBuilder()
-                            .putFields("some_field", Value.newBuilder().setNumberValue(i).build())
-                            .build())
-                    .setId(upsertIds.get(i))
+                    .setId(upsertIds.get(i % upsertData.length))
                     .build());
         }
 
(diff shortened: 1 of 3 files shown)


src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
```diff
@@ -15,17 +15,22 @@ public class BuildUpsertRequest {
     private static final float[][] upsertData = {{1.0F, 2.0F, 3.0F}, {4.0F, 5.0F, 6.0F}, {7.0F, 8.0F, 9.0F}};
 
     public static UpsertRequest buildRequiredUpsertRequest() {
-        String namespace = RandomStringBuilder.build("ns", 8);
-        List<String> upsertIds = Arrays.asList("v1", "v2", "v3");
-        List<Vector> upsertVectors = new ArrayList<>();
+        return buildRequiredUpsertRequest(new ArrayList<>(), "");
+    }
+
+    public static UpsertRequest buildRequiredUpsertRequest(String namespace) {
```
austin-denoble, line 21: **praise:** Nice, I like the flexibility with the overloaded methods. 👍

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
```diff
@@ -0,0 +1,240 @@
...
+                .addAllIds(Arrays.asList(idsToDelete))
+                .setDeleteAll(false)
+                .build();
+        blockingStub.delete(deleteRequest);
+        Thread.sleep(3500);
+
+        // Call describeIndexStats to get updated counts
+        describeIndexStatsResponse = blockingStub.describeIndexStats(describeIndexRequest);
+        // Verify the updated vector count
+        assertEquals(describeIndexStatsResponse.getNamespacesMap().get(namespace).getVectorCount(), startVectorCount - idsToDelete.length);
```
austin-denoble, line 75: Does the `startVectorCount - idsToDelete.length` actually work here? Up above on ln.48 we're deleting `upsertIds.get(0)`. The `startVectorCount - idsToDelete.length` count works in this case as we've deleted 1 of 3 vectors.

Here though we're resetting `idsToDelete` and reusing `upsertIds.get(0)` on ln.63. So while `startVectorCount - idsToDelete.length` still validates, we're really only deleting 1 additional vector since we already deleted `upsertIds[0]`.

I was expecting ln.63 to be `idsToDelete = new String[]{upsertIds.get(1), upsertIds.get(2)};`, in which case `startVectorCount - idsToDelete.length` would equal 1, but we'd deleted 3 vectors. I'm wondering if it makes sense to hard-code things to keep legibility of the tests a bit clearer, but maybe I'm over thinking.

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+        // Delete by filtering
+        DeleteRequest deleteRequest = DeleteRequest.newBuilder()
+                .setNamespace(namespace)
+                .setDeleteAll(false)
+                .setFilter(Struct.newBuilder()
+                        .putFields("genre", Value.newBuilder()
+                                .setStructValue(Struct.newBuilder()
+                                        .putFields("$eq", Value.newBuilder()
+                                                .setStringValue("drama")
```
austin-denoble, line 124: A little unfortunate that you have to have knowledge of `BuildUpsertRequest`/`buildOptionalUpsertRequest` for the metadata fields we're using to delete. I think that's fine ultimately, just something to keep in mind with helpers and where they're being used in tests.

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
```diff
@@ -0,0 +1,240 @@
...
+                .addAllIds(Arrays.asList(idsToDelete))
+                .setDeleteAll(false)
+                .build();
+        blockingStub.delete(deleteRequest);
+        Thread.sleep(3500);
+
+        // Call describeIndexStats to get updated counts
+        describeIndexStatsResponse = futureStub.describeIndexStats(describeIndexRequest).get();
+        // Verify the updated vector count
+        assertEquals(describeIndexStatsResponse.getNamespacesMap().get(namespace).getVectorCount(), startVectorCount - idsToDelete.length);
```
austin-denoble, line 178: Same possible issue with `startVectorCount - idsToDelete.length` I mentioned above, unless I'm misunderstanding things.

austin-denoble (approved):
Overall looks good to me, thanks for continuing to get the integration tests off the ground!

Most of my questions and comments are around readability of tests and aren't necessarily blocking. We could probably bike shed how we mock things for quite a while.

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+        // Delete 1 vector
+        String[] idsToDelete = new String[]{upsertIds.get(0)};
+        DeleteRequest deleteRequest = DeleteRequest.newBuilder()
+                .setNamespace(namespace)
+                .addAllIds(Arrays.asList(idsToDelete))
+                .setDeleteAll(false)
+                .build();
+        blockingStub.delete(deleteRequest);
+        Thread.sleep(3500);
```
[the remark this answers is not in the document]
rohanshah18: Deletion is not atomic, so I don't want to take a chance where the deletion was not completed and the verification using describeIndexStats starts. Dont remember if I saw it failed atleast once but this can happen when integration tests are ran on every commit. This is where our internal conversation on making integration test bullet proof is related (instead of a thread.sleep, retry with exponential back-off).

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
rohanshah18: You're right, its a decision I made where the test is expecting the fields but this is only one instance I have seen where metadata is set, so I'll make the change now but I dont know how future patterns will look like. So a refactor might be needed in future.

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
rohanshah18: You are right, the reason this test was working is because I was using `idsToDelete = new String[]{upsertIds.get(1), upsertIds.get(2)}` instead of `idsToDelete = new String[]{upsertIds.get(1), upsertIds.get(2)}`. Good catch!!

src/integration/java/io/pinecone/integration/dataplane/UpsertAndDeleteTest.java
rohanshah18: Fixed!

PR #50, opened 2023-12-06
Refactor configure index test

rohanshah18:
## Problem

After recent changes in integration test to use the existing index if it exists, the configure index test was required to be updated to consume less resources.

## Solution

1. Refactored checkIfIndexExists code: 
    a. To support both data and control plane integration tests
    b. Check if any index matches the criteria compared to old approach of checking the first index of the list and otherwise create a new index.
2. For configure test: 
    a. Updated the beforeAll logic to check if an index exist 
    b. Deleted the afterEach block to delete each index and save them for future test

## Type of Change

Added Integration tests

## Test Plan

Ran integration tests locally.

### DIFF
--- a/src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
+++ b/src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
@@ -25,7 +25,7 @@ public class PineconeIndexOperationsClientIntegrationTest {
     }
 
     @Test
-    public void createAndDelete() throws IOException {
+    public void createAndDelete() throws IOException, InterruptedException {
         String indexName = RandomStringBuilder.build("index-name", 8);
 
         // Create an index
@@ -48,5 +48,6 @@ public class PineconeIndexOperationsClientIntegrationTest {
 
         // Cleanup
         pinecone.deleteIndex(indexName);
+        Thread.sleep(3500);
     }
 }
--- a/src/integration/java/io/pinecone/helpers/IndexManager.java
+++ b/src/integration/java/io/pinecone/helpers/IndexManager.java
@@ -8,31 +8,15 @@ import java.io.IOException;
 import java.util.List;
 
 public class IndexManager {
-    public PineconeConnection createIndexIfNotExists(int dimension) throws IOException, InterruptedException {
-        boolean createNewIndex = false;
-        String indexName = "";
+    public PineconeConnection createIndexIfNotExistsDataPlane(int dimension) throws IOException, InterruptedException {
         PineconeClientConfig config = new PineconeClientConfig()
                 .withApiKey(System.getenv("PINECONE_API_KEY"))
                 .withEnvironment(System.getenv("PINECONE_ENVIRONMENT"));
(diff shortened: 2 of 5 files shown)


src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
```diff
@@ -48,5 +48,6 @@ public void createAndDelete() throws IOException {
 
         // Cleanup
         pinecone.deleteIndex(indexName);
+        Thread.sleep(3500);
```
rohanshah18, line 51: We must have a timeout after index deletion is called so the next test doesn't pick up this indexName, since deletion is not an atomic process.

src/integration/java/io/pinecone/integration/controlPlane/ConfigureIndexTest.java
```diff
@@ -163,13 +133,17 @@ public void sizeIncrease() {
             // Get the index description to verify the new pod type
             indexMeta = indexOperationClient.describeIndex(indexName);
             assertEquals("p1.x2", indexMeta.getDatabase().getPodType());
+
+            // Delete this index since it'll be unused for future tests
+            indexOperationClient.deleteIndex(indexName);
```
rohanshah18, line 138: Index of pod-type p1.x2 i.e. scaled up is useless for future tests, so delete them right away.

src/integration/java/io/pinecone/integration/controlPlane/ConfigureIndexTest.java
```diff
@@ -1,13 +1,12 @@
-package io.pinecone.integration;
+package io.pinecone.integration.controlPlane;
```
rohanshah18, line 1: Moved this file into controlPlane sub-package.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -43,6 +27,40 @@ public PineconeConnection createIndexIfNotExists(int dimension) throws IOExcepti
...
+        List<String> indexList = controlPlaneClient.listIndexes();
+        String indexName = checkIfIndexExists(indexList, dimension, controlPlaneClient);
+
+        return (indexName.isEmpty()) ? createNewIndex(dimension, controlPlaneClient) : indexName;
+    }
+
+    private static String checkIfIndexExists(List<String> indexList, int dimension, PineconeIndexOperationClient controlPlaneClient)
+            throws IOException, InterruptedException {
+        int i = 0;
+        while (i < indexList.size()) {
```
rohanshah18, line 41: Compare to former approach of checking for the first index of the index list only, check for all indexes in the list if there's a potential index that can be used for the test. If not, create a new index. This has helped me cut down total time of the integration tests by 67% (from 10 mins to less than 3 mins).

jhamon (approved):
Nice work!

src/integration/java/io/pinecone/PineconeIndexOperationsClientIntegrationTest.java
rohanshah18: Discussed this internally on why its needed!

PR #58, opened 2024-01-12
Add source_collection to indexMetaDatabase object and ignore newly added fields

rohanshah18:
## Problem

`source_collection` is a newly added field in `indexMetaDatabase` object of describeIndex response body. But without adding it to the response body of describe index call, it breaks. 

## Solution

Added the field as well as added support to ignore newly added fields.

## Type of Change

## Test Plan
Ran integration tests and added unit test for the same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [comment]: <> (When bumping [pc:VERSION_LATEST_RELEASE] create a new entry below)
 ### Unreleased version
+### v0.7.4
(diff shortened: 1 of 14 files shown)


jhamon (approved):
Ah, yes. `ignoreUnknown = true` is very important. I'm pretty surprised that isn't the default behavior. Sorry it caused an issue in this case, but this will be much better going forward.

We should add this onto every model we deserialize into, not just the `IndexMeta` case. So we can be resilient to additive changes.

PR #59, opened 2024-01-24
Add global control plane code

rohanshah18:
## Problem

Global control plane support needs to be added to the Java SDK.

## Solution

Add open api generated code for the global control plane.
1. Files under org.openapitools.client are auto generated and can be ignored during the review
2. Updated `gradle build` to `gradle clean build` for the github actions
3. Rearranged the test folders for adding collection tests
4. Updated `build.gradle` to add dependencies for the open api generated code

## Type of Change

## Test Plan

Ran integration tests for the existing code base but future PR's will be adding unit and integration tests for each operations.


austin-denoble (approved):
Overall code looks good to me, we'll need to integrate all of the generated control plane logic into `PineconeIndexOperationClient` to really test things out, but the passing builds should give us some confidence here.

It seems one of the integration test runs is failing. Specifically `Java 17` with `Gradle 7.3.1`. Looks like the `ConfigureIndexTest` failed a bunch of different times on `setUp()`. The other failure is `UpdateAndQueryTest` where it looks like there was an assertion failure, which may be due to flakiness or freshness. Not sure if you want to try and address these before merging.

rohanshah18:
@austin-denoble That's correct! As we discussed internally, this PR is meant to add openAPI changes only and the future index operations tickets are meant to add integration tests for index operations.
Also you're currently working on the collections, those tickets will help us test the open api generated code for collections.

PR #65, opened 2024-02-14
Add collections operations to `PineconeControlPlaneClient` with integration tests

austin-denoble:
## Problem
The Java SDK is currently missing collections functionality. Now that we've added generated code for the OpenAPI spec, we can hook up collections operations in `PineconeControlPlaneClient`.

We'd also like to add integration tests to cover collections.

## Solution
- Update `PineconeControlPlaneClient` to include collections operations for create, list, delete, and describe.
- Add new integration test files: `CollectionTest` and `CollectionErrorTest`.
- Add new helpers to `IndexManager` for creating and connecting to an index by `indexName`, creating a collection and waiting for it to be ready, and polling until an index is ready.
- Add new helper to `BuildUpsertRequest` for generating vectors by dimension.

Bonuses:

- I added some logging configs in `build.gradle` for `test` and the `integrationTest` task. This was primarily to help me debug things in CI and log a bit more info to the console. We can tweak as needed, but I think having something like this will be really helpful as opposed to the generic output.
- Set `max-parallel: 1` in the `integration-test` job in the `pr.yml` workflow. Because of the way `findIndexWithDimensionAndType` works we cannot run integration tests in parallel without flakiness.
- Fixed an issue in `AssertRetry`. We were incrementing `delay` at the class level which is a static variable, meaning if you re-used the function in one place it would continue increasing the delay for all subsequent calls based on the `backOff`.

## Type of Change

## Test Plan
Run integration tests to make sure `CollectionTest` and `CollectionErrorTest` pass as expected. Integration tests should also be passing in CI barring any flakiness.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -58,6 +58,7 @@ jobs:
             { java: 8, gradle: 6.8 },
             { java: 17, gradle: 7.3.1 }
         ]
+      max-parallel: 1
     steps:
     - uses: actions/checkout@v4
 
@@ -83,5 +84,5 @@ jobs:
       env:
         PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
         PINECONE_ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
-        
+
 
--- a/build.gradle
+++ b/build.gradle
@@ -97,7 +97,6 @@ tasks.named('build') {
 test {
     useJUnitPlatform()
 
-
     testLogging {
         events = [TestLogEvent.FAILED, TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.STANDARD_OUT]
         exceptionFormat = TestExceptionFormat.FULL
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,3 @@
-import org.gradle.api.tasks.testing.logging.TestExceptionFormat;
-import org.gradle.api.tasks.testing.logging.TestLogEvent;
-
 plugins {
     id 'com.github.johnrengelman.shadow' version '6.1.0'
     id 'java-library'
--- a/build.gradle
+++ b/build.gradle
@@ -96,35 +96,6 @@ tasks.named('build') {
 
 test {
     useJUnitPlatform()
-
-    testLogging {
-        events = [TestLogEvent.FAILED, TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.STANDARD_OUT]
-        exceptionFormat = TestExceptionFormat.FULL
-        showExceptions = true
-        showCauses = true
-        showStackTraces = true
-
-        debugOptions {
-            events = [
-                    TestLogEvent.STARTED,
-                    TestLogEvent.FAILED,
-                    TestLogEvent.PASSED,
-                    TestLogEvent.SKIPPED,
-                    TestLogEvent.STANDARD_ERROR,
-                    TestLogEvent.STANDARD_OUT
-            ]
-            exceptionFormat = TestExceptionFormat.FULL
-        }
-
-        afterSuite { desc, result ->
-            if (!desc.parent) {
-                def output = "Results: ${result.resultType} (${result.testCount}) tests, ${result.successfulTestCount} successful, ${result.failedTestCount} failed, ${result.skippedTestCount} skipped."
-                def startItem = "| ", endItem = " |"
-                def repeatLength = startItem.length() + output.length() + endItem.length()
-                println('\n' + ('-' * repeatLength) + '\n' + startItem + output + endItem + '\n' + ('-' * repeatLength))
-            }
-        }
-    }
 }
 
 task integrationTest(type: Test) {
@@ -132,37 +103,6 @@ task integrationTest(type: Test) {
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
     classpath = sourceSets.integrationTest.runtimeClasspath
     outputs.upToDateWhen { false }
-
-
-    testLogging {
-        events = [TestLogEvent.FAILED, TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.STANDARD_OUT]
-        exceptionFormat = TestExceptionFormat.FULL
-        showExceptions = true
-        showCauses = true
-        showStackTraces = true
-
-        debugOptions {
-            events = [
-                    TestLogEvent.STARTED,
-                    TestLogEvent.FAILED,
-                    TestLogEvent.PASSED,
-                    TestLogEvent.SKIPPED,
-                    TestLogEvent.STANDARD_ERROR,
-                    TestLogEvent.STANDARD_OUT
-            ]
-            exceptionFormat = TestExceptionFormat.FULL
-
-        }
-
-        afterSuite { desc, result ->
-            if (!desc.parent) {
-                def output = "Results: ${result.resultType} (${result.testCount}) tests, ${result.successfulTestCount} successful, ${result.failedTestCount} failed, ${result.skippedTestCount} skipped."
-                def startItem = "| ", endItem = " |"
-                def repeatLength = startItem.length() + output.length() + endItem.length()
-                println('\n' + ('-' * repeatLength) + '\n' + startItem + output + endItem + '\n' + ('-' * repeatLength))
-            }
-        }
-    }
 }
 
 // Configure Auto Relocation
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,6 @@
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat;
+import org.gradle.api.tasks.testing.logging.TestLogEvent;
+
 plugins {
     id 'com.github.johnrengelman.shadow' version '6.1.0'
     id 'java-library'
@@ -93,6 +96,36 @@ tasks.named('build') {
 
 test {
     useJUnitPlatform()
+
+
+    testLogging {
+        events = [TestLogEvent.FAILED, TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.STANDARD_OUT]
+        exceptionFormat = TestExceptionFormat.FULL
+        showExceptions = true
+        showCauses = true
+        showStackTraces = true
+
+        debugOptions {
+            events = [
+                    TestLogEvent.STARTED,
+                    TestLogEvent.FAILED,
+                    TestLogEvent.PASSED,
+                    TestLogEvent.SKIPPED,
+                    TestLogEvent.STANDARD_ERROR,
+                    TestLogEvent.STANDARD_OUT
+            ]
+            exceptionFormat = TestExceptionFormat.FULL
+        }
+
+        afterSuite { desc, result ->
+            if (!desc.parent) {
+                def output = "Results: ${result.resultType} (${result.testCount}) tests, ${result.successfulTestCount} successful, ${result.failedTestCount} failed, ${result.skippedTestCount} skipped."
+                def startItem = "| ", endItem = " |"
+                def repeatLength = startItem.length() + output.length() + endItem.length()
+                println('\n' + ('-' * repeatLength) + '\n' + startItem + output + endItem + '\n' + ('-' * repeatLength))
+            }
+        }
+    }
 }
 
 task integrationTest(type: Test) {
@@ -100,6 +133,37 @@ task integrationTest(type: Test) {
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
     classpath = sourceSets.integrationTest.runtimeClasspath
     outputs.upToDateWhen { false }
+
+
+    testLogging {
+        events = [TestLogEvent.FAILED, TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.STANDARD_OUT]
+        exceptionFormat = TestExceptionFormat.FULL
+        showExceptions = true
+        showCauses = true
+        showStackTraces = true
+
+        debugOptions {
+            events = [
+                    TestLogEvent.STARTED,
+                    TestLogEvent.FAILED,
+                    TestLogEvent.PASSED,
+                    TestLogEvent.SKIPPED,
+                    TestLogEvent.STANDARD_ERROR,
+                    TestLogEvent.STANDARD_OUT
+            ]
+            exceptionFormat = TestExceptionFormat.FULL
+
+        }
+
+        afterSuite { desc, result ->
+            if (!desc.parent) {
+                def output = "Results: ${result.resultType} (${result.testCount}) tests, ${result.successfulTestCount} successful, ${result.failedTestCount} failed, ${result.skippedTestCount} skipped."
+                def startItem = "| ", endItem = " |"
+                def repeatLength = startItem.length() + output.length() + endItem.length()
+                println('\n' + ('-' * repeatLength) + '\n' + startItem + output + endItem + '\n' + ('-' * repeatLength))
+            }
+        }
+    }
 }
 
 // Configure Auto Relocation
--- a/src/integration/java/io/pinecone/helpers/AssertRetry.java
+++ b/src/integration/java/io/pinecone/helpers/AssertRetry.java
@@ -13,16 +13,21 @@ public class AssertRetry {
         assertWithRetry(assertionRunnable, 2);
     }
 
-    public static void assertWithRetry(AssertionRunnable assertionRunnable, int backOff) throws InterruptedException, PineconeException {
+    public static void assertWithRetry(AssertionRunnable assertionRunnable, int backOff) throws AssertionError, InterruptedException {
         int retryCount = 0;
         int delayCount = delay;
         boolean success = false;
 
-        while (retryCount < maxRetry && !success) {
+        while (!success) {
             try {
                 assertionRunnable.run();
                 success = true;
             } catch (AssertionError | ExecutionException | IOException e) {
+                // If we've hit the max number of retries throw and abort
+                if (retryCount == maxRetry) {
+                    throw new AssertionError(e.getLocalizedMessage());
+                }
+
                 retryCount++;
                 Thread.sleep(delayCount);
                 delayCount*=backOff;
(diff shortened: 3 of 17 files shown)


.github/workflows/pr.yml
```diff
@@ -58,6 +58,7 @@ jobs:
             { java: 8, gradle: 6.8 },
             { java: 17, gradle: 7.3.1 }
         ]
+      max-parallel: 1
```
austin-denoble, line 61: I bumped this down to 1 but I'm still seeing issues in CI.

build.gradle
```diff
@@ -93,13 +96,73 @@ tasks.named('build') {
 
 test {
     useJUnitPlatform()
+
+    testLogging {
```
austin-denoble, line 100: These configs add logging to our test runs, which has been helpful in trying to debug. We can leave them in or modify them or rip them out, but I think we should look at improving what we output when tests are running.

src/integration/java/io/pinecone/helpers/AssertRetry.java
```diff
@@ -21,14 +24,14 @@ public static void assertWithRetry(AssertionRunnable assertionRunnable, int back
                 success = true;
             } catch (AssertionError | ExecutionException | IOException e) {
                 retryCount++;
-                delay*=backOff;
```
austin-denoble, line 24: Previously this was applying the `backOff` before the first sleep.

src/integration/java/io/pinecone/helpers/AssertRetry.java
```diff
@@ -1,18 +1,21 @@
...
 
-    public static void assertWithRetry(AssertionRunnable assertionRunnable) throws InterruptedException {
+    public static void assertWithRetry(AssertionRunnable assertionRunnable) throws InterruptedException, PineconeException {
         assertWithRetry(assertionRunnable, 2);
     }
 
-    public static void assertWithRetry(AssertionRunnable assertionRunnable, int backOff) throws InterruptedException {
+    public static void assertWithRetry(AssertionRunnable assertionRunnable, int backOff) throws InterruptedException, PineconeException {
         int retryCount = 0;
+        int delayCount = delay;
```
austin-denoble, line 18: I moved to this because I noticed that if you re-used this function in the same class / test, it would continue incrementing the static `delay` value without resetting it, leading to really long wait times for certain things.

src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
```diff
@@ -44,6 +41,23 @@ public static UpsertRequest buildRequiredUpsertRequest(List<String> upsertIds, S
                 .build();
     }
 
+    public static UpsertRequest buildRequiredUpsertRequestByDimension(List<String> upsertIds, int dimension, String namespace) {
```
austin-denoble, line 44: I wanted to be able to generate vectors with a specific `dimension` rather than relying on the hard-coded `upsertData` in this class.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -45,9 +48,11 @@ private static String findIndexWithDimensionAndType(IndexList indexList, int dim
         List<IndexModel> indexModels = indexList.getIndexes();
         while (i < indexModels.size()) {
             IndexModel indexModel = isIndexReady(indexModels.get(i).getName(), controlPlaneClient);
-            // ToDo: add pod type support
             if (indexModel.getDimension() == dimension
-                    && ((indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null && indexModel.getSpec().getPod().getReplicas() == 1 && indexModel.getSpec().getPod().getPodType().equalsIgnoreCase("p1.x1"))
+                    && ((indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD)
```
austin-denoble, line 52: I was having trouble reading this if block so I moved some things.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -79,6 +84,76 @@ private static String createNewIndex(PineconeControlPlaneClient controlPlaneClie
         return indexName;
     }
 
+    public static IndexModel waitUntilIndexIsReady(PineconeControlPlaneClient controlPlaneClient, String indexName, Integer totalMsToWait) throws InterruptedException {
```
austin-denoble, line 87: I added this because using `isIndexReady` wasn't sufficient for waiting on indexes that are created from collections.

They seemingly can take a long time, sometimes longer than 2 minutes. `isIndexReady` only allows you to control the backoff, which is clunky when trying to specify a more specific amount of time to wait.

We do something similar to this for polling in the other clients.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -20,16 +20,12 @@
...
+    private static String indexName;
     private static final Logger logger = LoggerFactory.getLogger(PineconeClientLiveIntegTest.class);
 
     @BeforeAll
-    public static void defineControlPlaneClient() {
+    public static void setUp() throws InterruptedException, IOException {
         controlPlaneClient = new PineconeControlPlaneClient(System.getenv("PINECONE_API_KEY"));
-    }
-
-    @BeforeEach
```
austin-denoble, line 31: This didn't seem like it needed to be called before each test, I think doing it before everything is fine, but let me know if not.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -83,7 +79,7 @@ public void scaleUpAndDown() {
             });
 
             // Scaling down
-            pod = new ConfigureIndexRequestSpecPod().replicas(3);
+            pod = new ConfigureIndexRequestSpecPod().replicas(1);
```
austin-denoble, line 82: This didn't seem correct as it's validating for `replicas = 1` below in the `assertWithRetry`. I'm not sure how that was passing previously.

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,4 +1,4 @@
-package io.pinecone.integration.controlPlane.index.pod;
+package io.pinecone.integration.controlPlane.pod;
```
austin-denoble, line 1: I don't think we need the extra `index` here, it would only be index and collections anyways.

src/integration/java/io/pinecone/integration/dataplane/PineconeClientLiveIntegTest.java
```diff
@@ -24,14 +25,21 @@
...
     private static VectorServiceGrpc.VectorServiceBlockingStub blockingStub;
 
     @BeforeAll
     public static void defineConfig() throws IOException, InterruptedException {
-        PineconeConnection connection = createIndexIfNotExistsDataPlane(3, IndexModelSpec.SERIALIZED_NAME_POD);
+        connection = createIndexIfNotExistsDataPlane(3, IndexModelSpec.SERIALIZED_NAME_POD);
         blockingStub = connection.getBlockingStub();
     }
 
+    @AfterAll
```
austin-denoble, line 38: I added `@AfterAll` in a bunch of these tests to make sure we're closing the `PineconeConnection` once we're done.

src/main/java/io/pinecone/PineconeControlPlaneClient.java
```diff
@@ -29,17 +27,17 @@ public PineconeControlPlaneClient(String apiKey, OkHttpClient okHttpClient) {
         manageIndexesApi.setApiClient(apiClient);
     }
 
-    public IndexModel createIndex(CreateIndexRequest createIndexRequest) {
+    public IndexModel createIndex(CreateIndexRequest createIndexRequest) throws PineconeException {
         IndexModel indexModel = new IndexModel();
         try {
-            manageIndexesApi.createIndex(createIndexRequest);
+            indexModel = manageIndexesApi.createIndex(createIndexRequest);
```
austin-denoble, line 33: This was never being returned.

src/main/java/io/pinecone/PineconeControlPlaneClient.java
```diff
@@ -67,14 +65,52 @@ public IndexList listIndexes() {
...
-    public void deleteIndex(String indexName) {
+    public void deleteIndex(String indexName) throws PineconeException  {
         try {
             manageIndexesApi.deleteIndex(indexName);
         } catch (ApiException apiException) {
             handleApiException(apiException);
         }
     }
 
+    public CollectionModel createCollection(CreateCollectionRequest createCollectionRequest) throws PineconeException {
```
austin-denoble, line 76: Here's the implementation for all the collection operations.

src/main/java/io/pinecone/PineconeControlPlaneClient.java
```diff
@@ -29,17 +27,17 @@ public PineconeControlPlaneClient(String apiKey, OkHttpClient okHttpClient) {
         manageIndexesApi.setApiClient(apiClient);
     }
 
-    public IndexModel createIndex(CreateIndexRequest createIndexRequest) {
+    public IndexModel createIndex(CreateIndexRequest createIndexRequest) throws PineconeException {
```
austin-denoble, line 30: These should all throw `PineconeException` from the catch blocks when calling `handleApiException`.

src/integration/java/io/pinecone/integration/dataplane/PineconeClientLiveIntegTest.java
rohanshah18: I like this!!

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -79,6 +84,76 @@ private static String createNewIndex(PineconeControlPlaneClient controlPlaneClie
...
+        IndexModel index = controlPlaneClient.describeIndex(indexName);
+        int waitedTimeMs = 0;
+        int intervalMs = 1500;
+
+        while (!index.getStatus().getReady()) {
+            index = controlPlaneClient.describeIndex(indexName);
+            if (waitedTimeMs >= totalMsToWait) {
+                System.out.println("Index " + indexName + " not ready after " + waitedTimeMs + "ms");
```
rohanshah18, line 95: Let's replace this with logger.info

build.gradle
rohanshah18: Nice, we can leave it for now and if it gets noisy or creates any issue, we can reconsider it later

.github/workflows/pr.yml
rohanshah18: Great, I like this change!

src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
rohanshah18: Okay, I can cleanup and refactor when dealing with data plane tests

src/integration/java/io/pinecone/helpers/IndexManager.java
rohanshah18: Adding specific amount of time is like hard coding (if there's some platform issue, it'll fail), so can we pass a higher backoff instead?

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionErrorTest.java
```diff
@@ -0,0 +1,160 @@
...
+        Thread.sleep(2500);
+    }
+
+    @Test
+    public void testCreateCollectionFromInvalidIndex() {
+        try {
+            CreateCollectionRequest createCollectionRequest = new CreateCollectionRequest().name(RandomStringBuilder.build("coll1", 8)).source("invalid-index");
+            controlPlaneClient.createCollection(createCollectionRequest);
+        } catch (PineconeException exception) {
+            System.out.println("Exception: " + exception.getMessage());
```
rohanshah18, line 76: These print statements can be removed right?

src/integration/java/io/pinecone/helpers/IndexManager.java
rohanshah18: Example: 
Define this at the class level:

Within the method where we need to log:
logger.info("logging statement");

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
rohanshah18: Oh I meant to have two directories:
1. index tests
2. collection tests

So I created a placeholder but no big deal

src/main/java/io/pinecone/PineconeControlPlaneClient.java
rohanshah18: Good catch!!

rohanshah18 (approved):
Great work!! I have left some comments but one area especially I'd like to think about is the hard-coded time outs added for collection tests in indexManager's waitUntilReady methods.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionErrorTest.java
austin-denoble: I actually added them because they're helpful for debugging if the test fails. With only the assertion, if the `exception` doesn't contain what we expect we don't see the actual error message in the output, so I feel like this is probably fine to leave in for when/if these tests fail and we need to debug.

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: Good point, let me take a look at updating all the println statements.

.github/workflows/pr.yml
rohanshah18: Also you said you were seeing issues in CI so what change resolved those issues? I see that the existing tests have been kept as it is.

build.gradle
rohanshah18: Actually we can debug the tests by adding flags --debug and/or --stacktrace on local or in CI, so is it okay if we remove this for now and keep the focus of this PR on adding tests for collections only?

rohanshah18: We can't add --debug to the CI though because it can leak information. So if the logging is required, we can add this block but for now, the CI will tell us which test failed with what condition.

src/integration/java/io/pinecone/helpers/AssertRetry.java
```diff
@@ -21,14 +24,14 @@ public static void assertWithRetry(AssertionRunnable assertionRunnable, int back
                 success = true;
```
rohanshah18, line 24: Can we update this function to the following:
```java
    public static void assertWithRetry(AssertionRunnable assertionRunnable, int backOff) throws InterruptedException {
        int retryCount = 0;
        boolean success = false;
        String errorMessage = null;

        while (retryCount < maxRetry && !success) {
            try {
                assertionRunnable.run();
                success = true;
            } catch (AssertionError | ExecutionException | IOException e) {
                retryCount++;
                delay*=backOff;
                Thread.sleep(delay);
                errorMessage = e.getLocalizedMessage();
            }
        }

        if (!success) {
            throw new AssertionError(errorMessage);
        }
    }
```

I dont have strong opinions about adding backOff before or after and adding the PineconeException to the signature but this will make sure the assert throws an error after all of the retries.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
rohanshah18: The reason we want this is to make sure each test starts by checking if the starting state is correct. We already have a condition in each of the tests and previously, we used to have an extra test which was removed, so it should safe to remove this.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -83,7 +79,7 @@ public void scaleUpAndDown() {
...
             spec = new ConfigureIndexRequestSpec().pod(pod);
             configureIndexRequest = new ConfigureIndexRequest().spec(spec);
             controlPlaneClient.configureIndex(indexName, configureIndexRequest);
```
rohanshah18, line 85: Not sure why I can't comment on line#98 but can we update it to the following:
`throw new PineconeException("Test failed: " + exception.getLocalizedMessage());`

Currently the error will show a hashcode and not the message itself.

src/main/java/io/pinecone/PineconeControlPlaneClient.java
rohanshah18: This will improve readability!

src/main/java/io/pinecone/PineconeControlPlaneClient.java
```diff
@@ -67,14 +65,52 @@ public IndexList listIndexes() {
...
+
+    public void deleteCollection(String collectionName) throws PineconeException {
+        try {
+            manageIndexesApi.deleteCollection(collectionName);
+        } catch (ApiException apiException) {
+            handleApiException(apiException);
+        }
+    }
+
     private void handleApiException(ApiException apiException) {
```
rohanshah18, line 114: We can also add the throws statement here to improve readability

austin-denoble: Good callout, I missed this one but will add it in.

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: Right, yeah I think we just don't need the extra level given collections could be classified as index operations. I think we can just group everything under `pod`/`serverless` for now.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
rohanshah18: Line#94 *

PR #68, opened 2024-02-22
Update changelogs, sdk version, and user-agent for v0.8.0 release

rohanshah18:
## Problem

Release control plane operations for java sdk and cleanup configure index test

## Solution

Update changelogs, sdk version, and user-agent for v0.8.0 release
Configure index test had an incorrect condition and the catch() wasn't throwing the exception.

## Type of Change

## Test Plan

Ran integration tests and published the jar locally to test all operations in a maven project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 [comment]: <> (When bumping [pc:VERSION_LATEST_RELEASE] create a new entry below)
 ### Unreleased version
+### v0.8.0
+- Add support for control plane operations for serverless indexes
+- Add support for collections for pod indexes
+
 ### v0.7.4
 - Add source_collection and support to ignore newly added fields to the response body for describeIndex's indexMetaDatabase object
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 [comment]: <> (When bumping [pc:VERSION_LATEST_RELEASE] create a new entry below)
 ### Unreleased version
+### v0.8.0
+- Add support for control plane operations for serverless indexes
+- Add support for collections for pod indexes
```
austin-denoble, line 7: I think it's important that we add specifics about the primary class changes here for control plane, what do you think?

- Rename of `PineconeIndexOperationClient` -> `PineconeControlPlaneClient`.
- The change in constructor signatures for this class: `PineconeClientConfig` is no longer required, you now pass `apiKey` as a string directly. However, `PineconeClientConfig` is still needed when connecting to the data plane.

Calling these things out will be important as updating to this version will break users and we want to give them some heads up beyond looking at the integration test implementations.

austin-denoble (approved):
LGTM, thanks for getting this prepped.

CHANGELOG.md
rohanshah18: The release notes will have all of the details

PR #70, opened 2024-02-26
Add data plane wrapper and accept sparse indices as unsigned 32-bit integer

rohanshah18:
## Problem

1. The SDK is missing a wrapper for data plane operations that helps improve user interaction.
2. The sparse indices in Pinecone are `unsigned 32-bit integer` but since Java doesn't offer the support for it natively, so there is a datatype mismatch issue.

## Solution
1. Add a wrapper for data plane operations and broke down the input request objects to java native datatypes except for Struct. The wrapper contains operations for both blocking and the async stub. Also I have added overloaded methods so its easier for users to call the data plane operations.
Example:
`public UpsertResponse upsert(UpsertRequest request)`
vs. 
```
public UpsertResponse upsert(String id,
            List<Float> values,
            List<Long> sparseIndices,
            List<Float> sparseValues,
            com.google.protobuf.Struct metadata,
            String namespace)
```

2. The datatype for sparse indices in Pinecone db is `unsigned 32-bit integer` so as a part of this PR, the sparse indices will be now accepted as Java `long` instead of Java `int` with the input range of `unsigned 32-bit integer` [0, 2^32 - 1] and everything outside of this range will throw `PineconeValidationException`. As a part of this change, I have added 3 classes relating to the Query Response in the `unsigned_indices_model` package since we don't want to edit the java generated classes:

- QueryResponseWithUnsignedIndices.java (reflects QueryResponse)
- ScoredVectorWithUnsignedIndices.java (reflects ScoredVector)
- SparseValuesWithUnsignedIndices.java (reflects SparseValues)

## Type of Change

## Test Plan

Added unit and integration tests.


src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+
+    public PineconeDataPlaneClient(VectorServiceGrpc.VectorServiceBlockingStub blockingStub) {
+        if (blockingStub == null) {
+            throw new PineconeValidationException("BlockingStub cannot be null.");
+        }
+
+        this.blockingStub = blockingStub;
+    }
+
+    public UpsertResponse upsert(String id,
```
rohanshah18, line 21: Tested for each constructor

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+        upsertRequest.addVectors(vectorBuilder.build());
+
+        if (namespace != null && !namespace.isEmpty()) {
+            upsertRequest.setNamespace(namespace);
+        }
+
+        return blockingStub.upsert(upsertRequest.build());
+    }
+
+    public QueryResponse query(String id,
```
rohanshah18, line 72: I haven't added a separate query() since topk is the only required parameter and passing only top k while setting everything else to null is not going to be a widely used method

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+                               Struct filter,
+                               boolean includeValues,
+                               boolean includeMetadata,
+                               List<Float> vectors,
+                               List<Long> sparseIndices,
+                               List<Float> sparseValues) {
+        QueryRequest.Builder queryRequest = QueryRequest.newBuilder();
+
+        if (id != null && !id.isEmpty() && vectors != null && !vectors.isEmpty()) {
+            throw new PineconeValidationException("Invalid Query Request. Both id and vectors cannot be present");
```
rohanshah18, line 84: Will update this to:
Invalid Query Request. Both Vector ID and vectors cannot be present.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+        DescribeIndexStatsRequest.Builder describeIndexStatsRequest = DescribeIndexStatsRequest.newBuilder();
+
+        if (filter != null) {
+            describeIndexStatsRequest.setFilter(filter);
+        }
+
+        return blockingStub.describeIndexStats(describeIndexStatsRequest.build());
+    }
+
+    // ToDo: Add tests
```
rohanshah18, line 211: Going to be a separate PR to add tests and convert int to long for response.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -1,145 +1,145 @@
...
-        } catch (Exception exception) {
-            throw new PineconeException("Test failed: " + exception.getLocalizedMessage());
-        } finally {
-            // Delete this index since it'll be unused for future tests
-            controlPlaneClient.deleteIndex(indexName);
-            Thread.sleep(3500);
-        }
-    }
-}
+//package io.pinecone.integration.controlPlane.pod;
```
rohanshah18, line 1: Will re-enable this after data plane work is concluded

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndDescribeIndexStatsTest.java
```diff
@@ -33,91 +36,104 @@ public static void cleanUp() {
...
-            DescribeIndexStatsResponse describeIndexStatsResponse2 = futureStub.describeIndexStats(describeIndexRequest).get();
-
-            // verify updated vector and namespace counts
-            assertEquals(describeIndexStatsResponse2.getTotalVectorCount(), startVectorCount + upsertResponse.getUpsertedCount());
-            assertEquals(describeIndexStatsResponse2.getNamespacesCount(), startNamespaceCount + 1);
-        });
-    }
+    // ToDo: Update when future stub is added
+//    @Test
+//    public void UpsertRequiredVectorsAndDescribeIndexStatsFutureTest() throws ExecutionException, InterruptedException {
```
rohanshah18, line 97: All async tests are going to be a part of the next PR

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+
+    public QueryResponse query(String id,
+                               String namespace,
+                               int topK,
+                               Struct filter,
+                               boolean includeValues,
+                               boolean includeMetadata,
+                               List<Float> vectors,
+                               List<Long> sparseIndices,
+                               List<Float> sparseValues) {
```
jhamon, line 80: I would make some changes here:

- Put topK first, since it is always required
- Change `vectors` to `vector`. This aligns with the [python arg naming.](https://github.com/pinecone-io/pinecone-python-client/blob/main/pinecone/data/index.py#L330)
- Prioritize `values` over `id` because I think that is the more commonly used way of querying.
- Make `includeValues` and `includeMetadata` nullable, and implement defaults of `false` inside this method. This is less important for the others, but I think `null` is a nice shorthand for `I don't care, use a default value` that we can easily implement.

These changes will make common usage look like `query(10, Arrays.asList(0.1f, 0.2f, 0.3f), null, null, null, null, null, null, null)` grouping all nulls toward the end.

```suggestion
    public QueryResponse query(
                               int topK,
                               List<Float> vector,
                               List<Long> sparseIndices,
                               List<Float> sparseValues
                               String id,
                               String namespace,
                               Struct filter,
                               Boolean includeValues,
                               Boolean includeMetadata,
) {
            // todo implementation
    }
 ```


Then I think this would be better if we added some helpers such as `queryByVectorId` and `queryByVectorValues` methods. I think this makes the process of adding overloaded methods much easier, since id & vector you can ignore some fields that aren't relevant to that scenario.

```java
    public QueryResponse queryByVectorId(
                               int topK,
                               String id,
                               String namespace,
                               Struct filter,
                               Boolean includeValues,
                               Boolean includeMetadata,) {
            // todo implementation
    }

    public QueryResponse queryByVectorId(
                               int topK,
                               String id,
                               String namespace,
                               Struct filter
    ) {
            // todo implementation where includeValues, includeMetadata are null
    }

    public QueryResponse queryByVectorId(
                               int topK,
                               String id,
                               String namespace,
    ) {
            // todo implementation where includeValues, includeMetadata, filter are null
    }

    public QueryResponse queryByVectorId(
                               int topK,
                               String id
    ) {
            // todo implementation where includeValues, includeMetadata, filter, namespace are null
    }
```

jhamon (commented):
I'm not done going through this, but thought I would submit some initial feedback for you to work on.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionErrorTest.java
```diff
@@ -1,163 +1,163 @@
...
-            logger.info("Exception: " + exception.getMessage());
-            assert (exception.getMessage().contains("Source index is not ready"));
-        } finally {
-            // Wait for index to initialize so it can be deleted in @AfterAll
-            waitUntilIndexIsReady(controlPlaneClient, notReadyIndexName);
-        }
-    }
-}
\ No newline at end of file
+//package io.pinecone.integration.controlPlane.pod;
```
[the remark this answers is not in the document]
rohanshah18: Good point about the @Ignore. Also the configureIndexTest fails for InternalServerException which can be ignored and the collectionsTest is failing for some assertion because collections can take variable amount of time to initialize, so to speed up the process, I have disabled it for now. I can enable it and clean it later

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+                                 List<Float> sparseValues,
+                                 com.google.protobuf.Struct metadata,
+                                 String namespace) {
+        UpsertRequest.Builder upsertRequest = UpsertRequest.newBuilder();
+
+        if (id == null || id.isEmpty() || values == null || values.isEmpty()) {
+            throw new PineconeValidationException("Invalid Upsert Request. Please ensure that both id and values are provided.");
+        }
+
+        Vector.Builder vectorBuilder = Vector.newBuilder()
```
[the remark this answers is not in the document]
rohanshah18: Good point, let me add a test and verify the behavior

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+                .setId(id)
+                .addAllValues(values);
+
+        if (sparseIndices != null && sparseValues != null && sparseIndices.size() != sparseValues.size()) {
+            throw new PineconeValidationException("Invalid Upsert Request. Please ensure that both sparse indices and values are of the same length.");
+        }
```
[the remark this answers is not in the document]
rohanshah18: Sure, will add the test

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: This is really great feedback, thanks so much!!

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+
+        if (id != null && !id.isEmpty() && vectors != null && !vectors.isEmpty()) {
+            throw new PineconeValidationException("Invalid Query Request. Both Vector ID and vectors cannot be present.");
+        }
+
+        if (id != null && !id.isEmpty()) {
+            queryRequest.setId(id);
+        }
+
+        if (namespace != null && !namespace.isEmpty()) {
```
jhamon, line 91: `""` is a valid namespace that people may want to pass. It's also the default in the backend though so I don't think it makes much difference whether we send it or not.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+                               Struct filter,
+                               boolean includeValues,
+                               boolean includeMetadata,
+                               List<Float> vectors,
+                               List<Long> sparseIndices,
+                               List<Float> sparseValues) {
+        QueryRequest.Builder queryRequest = QueryRequest.newBuilder();
+
+        if (id != null && !id.isEmpty() && vectors != null && !vectors.isEmpty()) {
+            throw new PineconeValidationException("Invalid Query Request. Both Vector ID and vectors cannot be present.");
```
jhamon, line 84:
```suggestion
            throw new PineconeValidationException("Invalid Query Request. Cannot query with both vector id and vector values.");
```

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+    private Iterable<? extends Integer> convertUnsigned32IntToSigned32Int(Iterable<? extends Long> unsigned32IntValues) {
+        List<Integer> int32Values = new ArrayList<>();
+        for (Long value : unsigned32IntValues) {
+            if (value < 0 || value > 0xFFFFFFFFL) {
+                throw new PineconeValidationException("Sparse indices are out of range for unsigned 32-bit integers.");
+            }
+            int32Values.add(value.intValue());
+        }
+        return int32Values;
+    }
```
jhamon, line 221: Even though it's private, this feels out of place here since this class is mainly about defining the public interface for our data plane functionality. I would make it a static method on some kind of small utility class with thorough unit testing. Also some comments to explain the purpose of this will be useful in the future after we forget why this was necessary.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+            queryRequest.setSparseVector(SparseValues.newBuilder()
+                    .addAllIndices(convertUnsigned32IntToSigned32Int(sparseIndices))
+                    .addAllValues(sparseValues)
+                    .build());
+        }
+
+        return blockingStub.query(queryRequest.build());
+    }
+
+    public FetchResponse fetch(List<String> ids,
```
jhamon, line 121: I would add an overload method for `fetch(List<String> ids)` without namespace so that people who aren't using namespaces will not have to pass the default namespace value `""`.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+    public UpdateResponse update(String id,
+                                 List<Float> values,
+                                 Struct metadata,
+                                 String namespace,
+                                 List<Long> sparseIndices,
+                                 List<Float> sparseValues) {
+        UpdateRequest.Builder updateRequest = UpdateRequest.newBuilder();
+
+        if (id == null) {
+            throw new PineconeValidationException("Invalid Update Request. Vector ID must be present");
```
jhamon, line 147: Not sure how I feel about all the capitalization throughout these errors. I'd lean toward not capitalizing things unless you are referring to a specific Java class or something. I'm not going to clutter this review by  commenting on every occurrence of this, but would be good to handle this consistently across all exceptions.

```suggestion
            throw new PineconeValidationException("Invalid update request. Vector id must be present.");
```

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+
+        if (metadata != null) {
+            updateRequest.setSetMetadata(metadata);
+        }
+
+        if (namespace != null && !namespace.isEmpty()) {
+            updateRequest.setNamespace(namespace);
+        }
+
+        if (sparseIndices != null && sparseValues != null) {
```
jhamon, line 163: Similar to query, we may also need to worry about the error cases where only one of `sparseIndices` or `sparseValues` is set. This looks like if `sparseIndices == null` you would skip this check, then hit a NPE when trying to do the int/long conversion.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,222 @@
...
+                    .addAllValues(sparseValues)
+                    .build());
+        }
+
+        return blockingStub.update(updateRequest.build());
+    }
+
+    public DeleteResponse delete(List<String> ids,
+                                 boolean deleteAll,
+                                 String namespace, Struct filter) {
```
jhamon, line 179: Since most of the arguments don't make sense when used together, what do you think about this method breakdown:

- `delete(List<String> ids, Struct filter, boolean deleteAll, String namespace)`
- `deleteByIds(List<String> ids, String namespace)`
- `deleteByIds(List<String> ids)` for default namespace
- `deleteByFilter(Struct filter, String namespace)`
- `deleteByFilter(Struct filter)` for default namespace
- `deleteAll(String namespace)`

And as an additional convenience:
- `deleteById(String id, String namespace)` for one single id
- `deleteById(String id)` for one single id in default namespace

Thoughts @austin-denoble @haruska?

Maybe I'm misreading the API docs, but `deleteAll` is about clearing out vectors in a namespace which seems incompatible with specifying ids/filters, so that should also be checked as an error case.

jhamon (approved):
Here's some more feedback. Something to keep in mind in general is that namespaces are not used by everyone. We should have tests exercising the default and non-default namespace cases to make sure the usage looks reasonably concise for both use cases. Otherwise when we go to show how to do things in docs, it's going to look poor from a UX perspective.

I didn't have a chance to go line-by-line through all the tests, but since I made a lot of suggestions in the non-test code I'm expecting that to change. I don't love how many tests are commented in this PR... as a general rule, we want to keep main in a shippable state at all times and disabling a large amount of tests seems like moving away from that goal but I leave it up to you. I'm sure you already know, but there's a substantial amount of testing needed to cover all the combinations of optional parameters and make sure we aren't going to NPE anywhere when nulls are being passed around.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: No strong opinion on this, will update all of them.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: Okay, will remove isEmpty() on namespace.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: Yes, I'll be moving it out of this class because of the reasons you mentioned and another reason is if I want to test this function, I'll need to instantiate blockingStub.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: Currently, I'm skipping them so I'll add a validation that if sparseIndices is not null, then the sparseValues should be not null and of the same length

src/main/java/io/pinecone/PineconeDataPlaneClient.java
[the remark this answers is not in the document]
rohanshah18: Updated the logic and added a test. Will check for if either is not null, then throw an error.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: Updated the logic and added a test. Will check for if either one is not null, then throw an error. If both are null, continue. Lastly, continue with both not null for checking same size() and then setting it up.

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: Done!

src/main/java/io/pinecone/PineconeDataPlaneClient.java
rohanshah18: Added different options

rohanshah18:
@jhamon Thanks for the details feedback. I have made all of the changes and my intention was to get to the level where our expectations are same about the thin-wrapper. I 100% agree that the tests should not be commented so instead of a separate PR for future stub, I'll push a commit tomorrow with which will re-enable all of the tests.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -80,7 +80,7 @@ private static String createNewIndex(PineconeControlPlaneClient controlPlaneClie
         CreateIndexRequest createIndexRequest = new CreateIndexRequest()
                 .name(indexName)
                 .dimension(dimension)
-                .metric(IndexMetric.EUCLIDEAN)
+                .metric(IndexMetric.DOTPRODUCT)
```
rohanshah18, line 83: Needed to query with sparse values

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeleteTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+        assertWithRetry(() -> {
+            // Call describeIndexStats to get updated vector count
+            DescribeIndexStatsResponse describeIndexStatsResponse = dataPlaneClient.describeIndexStats(nullFilterStruct);
+            // Verify the updated vector count
+            assertEquals(describeIndexStatsResponse.getNamespacesMap().get(namespace).getVectorCount(), updatedVectorCount);
+        });
+    }
+
+    // ToDo: Update when future stub changes are in
```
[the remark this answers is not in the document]
rohanshah18: Wont need this since future stub changes are now part of this PR

rohanshah18:
The original plan was to split this PR into 3 individual PR's (data plane changes for blocking and future stubs, and converting sparseIndices to unsigned 32 bit int) so they are focused and easy to review but I have now merged them all into 1 PR.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionErrorTest.java
```diff
@@ -5,6 +5,7 @@
 import io.pinecone.exceptions.PineconeException;
 import io.pinecone.helpers.RandomStringBuilder;
 import io.pinecone.proto.VectorServiceGrpc;
+import org.junit.Ignore;
```
austin-denoble, line 8: You need to use the annotation on the class / specific functions. The tests are still running in CI.

I used `org.junit.jupiter.api.Disabled` in one of my tests in this file, ln 129 in this file. You can also pass a string to it with an explanation for why it's disabled/ignored.

You're disabling these tests because of flakiness, would you like me to look at cleaning them up and trying to make them a bit more consistent?

austin-denoble (approved):
Thanks for implementing all of @jhamon's feedback, nice progress on getting the data wrapper in place.

Is there anything you'd like me to look at in relation to the integration tests? You unassigned the control plane integration ticket from me so I'm not sure, but it's clear we need some clean-up and hardening, specifically the collection tests if you're disabling them all. Let me know if there's anything there you'd like me to work on.

See my comments on the `PineconeFutureDataPlaneClient` file around having a super class for these two classes to consolidate some of the validation logic and let me know if that makes sense.

src/main/java/io/pinecone/PineconeBlockingDataPlaneClient.java
```diff
@@ -0,0 +1,280 @@
...
+            vectorBuilder.setMetadata(metadata);
+        }
+
+        upsertRequest.addVectors(vectorBuilder.build());
+
+        if (namespace != null) {
+            upsertRequest.setNamespace(namespace);
+        }
+
+        return blockingStub.upsert(upsertRequest.build());
```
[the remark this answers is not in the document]
rohanshah18: Yes, error handling is part of my next PR

src/main/java/io/pinecone/unsigned_indices_model/QueryResponseWithUnsignedIndices.java
```diff
@@ -0,0 +1,54 @@
+package io.pinecone.unsigned_indices_model;
```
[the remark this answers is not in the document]
rohanshah18: just java convention

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeleteTest.java
```diff
@@ -0,0 +1,271 @@
...
+            // verify the updated vector count
+            assertEquals(describeIndexStatsResponse.getNamespacesMap().get(namespace).getVectorCount(), actualVectorCount);
+        });
+
+        // Delete 1 vector
+        List<String> idsToDelete = new ArrayList<>(3);
+        idsToDelete.add(upsertIds.get(0));
+        dataPlaneClient.delete(idsToDelete, false, namespace, nullFilterStruct);
+        vectorCount -= idsToDelete.size();
+        int testSingleDeletedVectorCount = vectorCount;
```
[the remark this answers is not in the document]
rohanshah18: Yeah I think my last commit cleaned up things

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionErrorTest.java
rohanshah18: I did enable all of them back but yes, let's talk offline on how we can clean it up. I have a few tickets in the roadmap that will help us clean this up.

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    null,
+                    null,
+                    null,
+                    generateVectorValuesByDimension(dimension));
+        } catch (PineconeValidationException validationException) {
+            assertEquals(validationException.getLocalizedMessage(), "Invalid upsert request. Please ensure that both sparse indices and values are present.");
+        }
+    }
+
+    // ToDo: @Test public void queryWithIncorrectVectorDimensionFuture()
```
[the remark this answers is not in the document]
rohanshah18: Yeah my last commit took care of it, thanks!!

PR #72, opened 2024-03-01
Refactor to improve user-experience and delete support for endpoint construction for data plane operations using projectId, env, and indexName 

rohanshah18:
## Problem

Currently the user has to create multiple config objects before they can call data plane operations.

## Solution

I have added PineconeConnection() constructor that takes apiKey and indexName as input parameters, and internally creates the config objects to provide a connection object directly without having the need to create the config objects. I have also refactored the classes especially PineconeConnection class as a part of this PR. Lastly, the endpoint construction for gRPC calls using projectName, indexName, and environment is discontinued, and will only use host parameter to get the endpoint. The host parameter is obtained by calling describeIndex().

Refactored the following:
1. Removed PineconePackageInfo since it wasn't used anywhere
2. Moved PineconeClient, PineconeConnection, PineconeConnectionConfig, and PineconeClientConfig under configs. I'll eventually like to cut down all dependencies to PineconeClient class and delete it but currently some of the test classes do rely on it and we can take care of it in future PR's. Also I'll like to keep PineconeClient out of clients package since I don't want users to think that it's one of the top level clients that will be useful to them. 
3. Moved PineconeControlPlaneClient, PineconeBlockingDataPlaneClient, and PineconeFutureDataPlaneClient under clients package.
4. Removed ServerSideTimeout field from PineconeClientConfig since the values were not set correctly.

## Type of Change

Refactored existing codebase so no new functionalities were added. 

## Test Plan


src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -29,13 +30,7 @@ public static PineconeConnection createIndexIfNotExistsDataPlane(int dimension,
...
         // Do not proceed until the newly created index is ready
         isIndexReady(indexName, controlPlaneClient);
-        // ToDo: Update the constructor by removing dependency on PineconeClientConfig
-        PineconeClient dataPlaneClient = new PineconeClient(config);
-        String host = controlPlaneClient.describeIndex(indexName).getHost();
-
-        return dataPlaneClient.connect(
-                new PineconeConnectionConfig()
-                        .withConnectionUrl("https://" + host));
+        return new PineconeConnection(apiKey, indexName);
```
rohanshah18, line 33: Testing the new PineconeConnection(String apiKey, String indexName) constructor

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -36,27 +37,47 @@ public class PineconeConnection implements AutoCloseable {
 
     private VectorServiceGrpc.VectorServiceFutureStub futureStub;
 
+    public PineconeConnection(String apiKey, String indexName) {
```
rohanshah18, line 40: Added the new constructor

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -116,17 +135,18 @@ private static Metadata assembleMetadata(PineconeClientConfig clientConfig) {
...
-    static String getEndpoint(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig) {
-        String endpoint = (connectionConfig.getConnectionUrl() != null) ?
-                connectionConfig.getConnectionUrl().replaceFirst("https?://", "") :
-                String.format("%s-%s.svc.%s.pinecone.io",
-                        connectionConfig.getIndexName(),
-                        clientConfig.getProjectName(),
-                        clientConfig.getEnvironment());
-
-        logger.debug("Pinecone endpoint is: " + endpoint);
+    public static String getEndpoint(String host) {
```
rohanshah18, line 138: Updated the getEndpoint() by removing the old method of constructing the gRPC endpoint using indexname, projectName, and environment.

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -116,17 +135,18 @@ private static Metadata assembleMetadata(PineconeClientConfig clientConfig) {
...
+        }
+        else {
+            throw new PineconeValidationException("Index host cannot be null or empty");
+        }
+    }
 
-        return endpoint;
+    private static String getHost(String apiKey, String indexName) {
+        PineconeControlPlaneClient controlPlaneClient = new PineconeControlPlaneClient(apiKey);
+        return controlPlaneClient.describeIndex(indexName).getHost();
```
austin-denoble, line 149: I feel like this may need to be a bit more robust in terms of possible failure of the `describeIndex` call. If that call throws, it may not be apparent to the user why it happened if we're intrinsically trying to resolve the host when they instantiated `PineconeConnection`.

Have you thought about having a cache that tracks index host URLs? We do this in the other clients with a singleton to prevent unnecessary calls to `describeIndex` when we've already resolved a specific host value. I think something like that backing this class may be useful as we could skip the calls to `describeIndex` if we've already retrieved it elsewhere.

austin-denoble (approved):
Overall I feel like we're trending in the right direction here. I definitely think we should look at removing `PineconeClient` entirely if that's not something we expect users to interact with. Is our intention to maintain the existing config classes to allow easier backwards compatibility on upgrading to v1.0.0?

I've left some thoughts in `PineconeConnection`, please let me know if there's anything you'd like to discuss.

src/main/java/io/pinecone/configs/PineconeConnection.java
rohanshah18: Sure!

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -141,12 +161,10 @@ void validateConfigs() throws PineconeValidationException {
         this.clientConfig.validate();
         this.connectionConfig.validate();
 
-        if (this.connectionConfig.getIndexName() != null) {
-            if (this.clientConfig.getEnvironment() == null || this.clientConfig.getProjectName() == null) {
-                throw new PineconeValidationException("Cannot connect with indexName "
-                        + this.connectionConfig.getIndexName()
-                        + " unless PineconeClientConfig contains projectName and environment");
-            }
+        if(connectionConfig.getIndexName() != null && !connectionConfig.getIndexName().isEmpty()
```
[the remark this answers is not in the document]
rohanshah18: Got it, thanks!

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -36,27 +37,47 @@ public class PineconeConnection implements AutoCloseable {
 
     private VectorServiceGrpc.VectorServiceFutureStub futureStub;
 
+    public PineconeConnection(String apiKey, String indexName) {
+        this.clientConfig = new PineconeClientConfig().withApiKey(apiKey);
+        String host = getHost(apiKey, indexName);
```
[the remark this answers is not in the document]
rohanshah18: Replied on the other thread

src/main/java/io/pinecone/configs/PineconeConnection.java
rohanshah18: As we discussed it internally, I'll handle this as a part of a ticket where I'm trying to decouple ManagedChannels with indexes.

PR #76, opened 2024-03-08
Add PineconeDataPlaneClient Interface

austin-denoble:
## Problem
Currently, we have two flavors of classes for dealing with `blocking` and `future` behavior for gRPC data plane calls:

- `PineconeBlockingDataPlaneClient`
- `PineconeFutureDataPlaneClient`

Ultimately, both of these classes conform to the same interface for wrapping data plane calls, and both classes have internal logic for validating the arguments passed to these requests that's duplicated across both classes:

PineconeBlockingDataPlaneClient | PineconeFutureDataPlaneClient
-- | --
upsert(String id,                                 List<Float> values) | upsert(String id,                                                   List<Float> values)
upsert(String id,                                 List<Float> values,                                 String namespace) | upsert(String id,                                                   List<Float> values,                                                   String namespace)
upsert(String id,                                 List<Float> values,                                 List<Long> sparseIndices,                                 List<Float> sparseValues,                                 com.google.protobuf.Struct metadata,                                 String namespace) | upsert(String id,                                                   List<Float> values,                                                   List<Long> sparseIndices,                                                   List<Float> sparseValues,                                                   com.google.protobuf.Struct metadata,                                                   String namespace)
queryByVectorId(int topK,                                         String id,                                         String namespace,                                         Struct filter,                                         boolean includeValues,                                         boolean includeMetadata) | queryByVectorId(int topK,                                                           String id,                                                           String namespace,                                                           Struct filter,                                                           boolean includeValues,                                                           boolean includeMetadata)
queryByVectorId(int topK,                                         String id,                                         String namespace,                                         Struct filter) | queryByVectorId(int topK,                                                           String id,                                                           String namespace,                                                           Struct filter)
queryByVectorId(int topK,                                         String id,                                         String namespace) | queryByVectorId(int topK,                                                           String id,                                                           String namespace)
queryByVectorId(int topK,                                         String id) | queryByVectorId(int topK,                                                           String id)
query(int topK,                               List<Float> vector,                               List<Long> sparseIndices,                               List<Float> sparseValues,                               String id,                               String namespace,                               Struct filter,                               boolean includeValues,                               boolean includeMetadata) | query(int topK,                                                 List<Float> vector,                                                 List<Long> sparseIndices,                                                 List<Float> sparseValues,                                                 String id,                                                 String namespace,                                                 Struct filter,                                                 boolean includeValues,                                                 boolean includeMetadata)
fetch(List<String> ids) | fetch(List<String> ids)
fetch(List<String> ids,                               String namespace) | fetch(List<String> ids,                                                 String namespace)
update(String id,                                 List<Float> values) | update(String id,                                                   List<Float> values)
update(String id,                                 List<Float> values,                                 String namespace) | update(String id,                                                   List<Float> values,                                                   String namespace)
update(String id,                                 List<Float> values,                                 Struct metadata,                                 String namespace,                                 List<Long> sparseIndices,                                 List<Float> sparseValues) | update(String id,                                                   List<Float> values,                                                   Struct metadata,                                                   String namespace,                                                   List<Long> sparseIndices,                                                   List<Float> sparseValues)
deleteByIds(List<String> ids, String namespace) | deleteByIds(List<String> ids, String namespace)
deleteByIds(List<String> ids) | deleteByIds(List<String> ids)
deleteByFilter(Struct filter, String namespace) | deleteByFilter(Struct filter, String namespace)
deleteByFilter(Struct filter) | deleteByFilter(Struct filter)
deleteAll(String namespace) | deleteAll(String namespace)
delete(List<String> ids,                                 boolean deleteAll,                                 String namespace,                                 Struct filter) | delete(List<String> ids,                                 boolean deleteAll,                                 String namespace,                                 Struct filter)
describeIndexStats(Struct filter) | describeIndexStats(Struct filter)

## Solution
Unify the two data plane wrapper classes with a `PineconeDataPlaneClient` interface.

This interface has default methods for dealing with validation steps for each core dataplane operation:
- `upsert`
- `query`
- `fetch`
- `update`
- `delete`
- `describeIndexStats`

The interface also has stubs for the various overloaded functions that are used to improve the development experience for consumers since Java 8 doesn't support optional arguments. Keeping this contract in an interface allows us to define the expected shape in one place, and avoid needing to validate consistency across the two files.

The interface takes in 6 generic types to represent the return types for each "group" of data plane operations. The reason for this is `PineconeBlockingDataPlaneClient` and `PineconeFutureDataPlaneClient` have different return types where the blocking class returns the response objects directly, and the future class returns `ListenableFuture<T>`. Using generics, this gives us the flexibility to define our own return types for each group of operations:
- `upsert` - `T`
- `query` - `U`
- `fetch` - `V`
- `update` - `W`
- `delete` - `X`
- `describeIndexStats` - `Z`

## Type of Change
- [X] None of the above: refactoring / abstraction

## Test Plan
The classes remain the same, so integration tests and builds should pass as normal.

### DIFF
--- a/src/main/java/io/pinecone/clients/PineconeDataPlaneClient.java
+++ b/src/main/java/io/pinecone/clients/PineconeDataPlaneClient.java
@@ -1,0 +1,241 @@
+package io.pinecone.clients;
+
+import com.google.protobuf.Struct;
+import io.pinecone.exceptions.PineconeValidationException;
+import io.pinecone.proto.*;
+
+import java.util.List;
+
+import static io.pinecone.utils.SparseIndicesConverter.convertUnsigned32IntToSigned32Int;
+
+public interface PineconeDataPlaneClient<T, U, V, W, X, Y> {
+    // default methods
+    default UpsertRequest validateUpsertRequest(String id,
+                                                List<Float> values,
+                                                List<Long> sparseIndices,
+                                                List<Float> sparseValues,
+                                                Struct metadata,
+                                                String namespace) {
+        UpsertRequest.Builder upsertRequest = UpsertRequest.newBuilder();
+
+        if (id == null || id.isEmpty() || values == null || values.isEmpty()) {
+            throw new PineconeValidationException("Invalid upsert request. Please ensure that both id and values are " +
+                    "provided.");
+        }
+
+        Vector.Builder vectorBuilder = Vector.newBuilder()
+                .setId(id)
+                .addAllValues(values);
+
+        if ((sparseIndices != null && sparseValues == null) || (sparseIndices == null && sparseValues != null)) {
+            throw new PineconeValidationException("Invalid upsert request. Please ensure that both sparse indices and" +
+                    " values are present.");
+        }
+        if (sparseIndices != null) {
+            if (sparseIndices.size() != sparseValues.size()) {
+                throw new PineconeValidationException("Invalid upsert request. Please ensure that both sparse indices" +
+                        " and values are of the same length.");
+            }
+            vectorBuilder.setSparseValues(SparseValues.newBuilder()
+                    .addAllIndices(convertUnsigned32IntToSigned32Int(sparseIndices))
+                    .addAllValues(sparseValues)
+                    .build());
+        }
+
+        if (metadata != null) {
+            vectorBuilder.setMetadata(metadata);
+        }
+
+        upsertRequest.addVectors(vectorBuilder.build());
+
+        if (namespace != null) {
+            upsertRequest.setNamespace(namespace);
+        }
+        return upsertRequest.build();
+    }
+
+    default QueryRequest validateQueryRequest(int topK,
+                                              List<Float> vector,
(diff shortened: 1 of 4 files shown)


src/main/java/io/pinecone/clients/PineconeDataPlaneClient.java
```diff
@@ -0,0 +1,241 @@
+package io.pinecone.clients;
```
rohanshah18, line 1: Can we move this under a new package named "commons" and rename this class to PineconeDataPlaneInterface? This will help the end-user to know that this is not a client where client is a class that the end-user will use.

rohanshah18 (approved):

src/main/java/io/pinecone/clients/PineconeDataPlaneClient.java
austin-denoble: Great call, added `.commons` and updated the naming for the interface.

PR #77, opened 2024-03-08
Refactor configs and disable collections and configure tests

rohanshah18:
## Problem

Currently, the Java SDK has PineconeClientConfig and PineconeConnectionConfig which are required to be instantiated before a user can call any dataplane operations. Although, recently we did create a constructor that allowed users to call any data plane operations without having the need of instantiating both configs. But in order to improve the user-experience and align it with other SDK's, we would like to combine the two config classes. 

Secondly, the CollectionsTest and ConfigureIndexTest are failing consistently causing inconvenience in CI pipeline.

## Solution

This PR removes PineconeClientConfig, PineconeConnectionConfig, and PineconeClient classes, and instead introduces PineconeConfig which has a required field of apiKey along with optional members such as host, usage-context, and customManagedChannel, where the customManagedChannel will allow the user to input custom gRPC channel for data plane operations.
As a part of this PR, I have disabled ConfigureIndexTest and CollectionsTest to improve testing pipeline.

## Type of Change

## Test Plan

Updated integration tests.


src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -38,27 +37,28 @@ public class PineconeConnection implements AutoCloseable {
...
-    public PineconeConnection(PineconeClientConfig clientConfig, PineconeConnectionConfig connectionConfig) {
-        this.connectionConfig = connectionConfig;
-        this.clientConfig = clientConfig;
-        validateConfigsAndGetHostIfEmpty();
-
-        channel = connectionConfig.getCustomChannelBuilder() != null
-                ? connectionConfig.getCustomChannelBuilder().apply(clientConfig, connectionConfig)
-                : buildChannel(connectionConfig.getConnectionUrl());
+    public PineconeConnection(PineconeConfig config, String indexName) {
+        this.config = config;
```
rohanshah18, line 47: Validating here

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -148,23 +147,4 @@ private static String getHost(String apiKey, String indexName) {
         PineconeControlPlaneClient controlPlaneClient = new PineconeControlPlaneClient(apiKey);
         return controlPlaneClient.describeIndex(indexName).getHost();
     }
-
-    void validateConfigsAndGetHostIfEmpty() throws PineconeValidationException {
```
rohanshah18, line 152: Deleting two configs removes the need for validations like this

src/test/java/io/pinecone/PineconeConnectionTest.java
```diff
@@ -10,20 +9,16 @@ public class PineconeConnectionTest {
 
     @Test
     void testGetEndpointWithConnectionUrlWithHttps() {
-        PineconeConnectionConfig connectionConfig = new PineconeConnectionConfig()
-                .withConnectionUrl("https://steps-784-123-eqasas0aaaa1213aasasc-1223-f1eea9.svc.production.pinecone.io");
-
-        String endpoint = PineconeConnection.formatEndpoint(connectionConfig.getConnectionUrl());
+        String hostWithHttps = "https://steps-784-123-eqasas0aaaa1213aasasc-1223-f1eea9.svc.production.pinecone.io";
```
rohanshah18, line 12: remove dependency on connectionConfig which wasn't needed in the first place

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndDescribeIndexStatsTest.java
```diff
@@ -157,9 +157,12 @@ public void upsertOptionalVectorsAndQueryIndexFutureTest() throws InterruptedExc
                     true).get();
 
             ScoredVectorWithUnsignedIndices scoredVectorV1 = null;
-            for (int i = 0; i < topK; i++) {
-                if (upsertIds.get(0).equals(queryResponse.getMatches(i).getId())) {
-                    scoredVectorV1 = queryResponse.getMatches(i);
+            // if the sizes are not equal, let the following assertions fail and retry again
```
rohanshah18, line 160: Should have made this change as a part of a separate PR, sorry :( but if there's any latency in querying the vectors, the test will fail with IndexOutOfBounds exception so instead, if the lists sizes are not equal, let the assertions fail and assertWithRetry will handle it.

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -0,0 +1,63 @@
...
+        ManagedChannel buildChannel();
+    }
+
+    void validate() {
+        if (apiKey == null)
+            throw new PineconeValidationException("Invalid PineconeConfig: missing apiKey");
+    }
+
+    public String getUserAgent() {
+        String userAgentLanguage = "lang=java; pineconeClientVersion = v0.8.0";
```
jhamon, line 59: We shouldn't be hard-coding this version string like this. We will definitely forget to update it when releasing.

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -0,0 +1,63 @@
...
+        return (this.getUsageContext() != null) ?
+                userAgentLanguage + "; usageContext=" + this.getUsageContext() : userAgentLanguage;
```
jhamon, line 61: This is a nit but I find this hard to read, mainly because of the `;` inside the string concatenation being so visually similar to `:`. 

Would be nice for readability to write this out fully.

```suggestion
        if (this.getUsageContext() == null) {
            return userAgentLanguage;
        } else {
            return userAgentLanguage + "; usageContext=" + this.getUsageContext();
        }
```

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -0,0 +1,63 @@
...
+        return this.customChannelBuilder;
+    }
+
+    public void setCustomChannelBuilder(ManagedChannel customChannelBuilder) {
+        this.customChannelBuilder = customChannelBuilder;
+    }
+
+    public interface CustomChannelBuilder {
+        ManagedChannel buildChannel();
+    }
```
jhamon, line 51: I'm confused about what's going on here. Shouldn't the return type of `getCustomChannelBuilder` be a `CustomChannelBuilder` instance which implements `buildChannel`? Seems like the interface you defined here is not being used for anything, which makes the method names quite confusing.

jhamon (commented):
Sharing a couple questions now. Have to jump into a meeting before finalizing this review.

src/main/java/io/pinecone/configs/PineconeConfig.java
rohanshah18: We shouldn't be hardcoding and we already have a ticket for this. We used to extract the sdk version at runtime but we had a bug so going to tackle this after the java reference docs. 

As a part of release process, it is one of the manual steps that I follow every time I release. In the README file, we also have to update it manually which is something I'll like to tackle later.

src/main/java/io/pinecone/configs/PineconeConfig.java
rohanshah18: We used to offer this functionality in PineconeConnectionConfig before. It allows the user to pass in their own gRPC ManagedChannel. If they don't provide it, we will create one for them within the PineconeConnection class. Also the reason why we have two constructors in PineconeConnection class.

src/main/java/io/pinecone/configs/PineconeConfig.java
rohanshah18: Sure, I can expand it.

src/main/java/io/pinecone/configs/PineconeConfig.java
rohanshah18: Good point, I'll rename it.

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -38,27 +37,28 @@ public class PineconeConnection implements AutoCloseable {
     private VectorServiceGrpc.VectorServiceFutureStub futureStub;
 
     public PineconeConnection(String apiKey, String indexName) {
-        this.clientConfig = new PineconeClientConfig().withApiKey(apiKey);
+        this.config = new PineconeConfig(apiKey);
```
jhamon, line 40: It seems weird to me that we would build a configuration object here; as we add features to the SDK, there could be a lot of other fields added to `PineconeConfig` that we would not have correct values of when building it here. For instance, if we implemented handling of custom SSL certs, you would not ever want to be newing up a fresh config object here in the middle of of the implementation. So I think we should strive to have the user always be the one to create that config object and pass it into our implementation.

So I think this should perhaps be changed to:

```java
public PineconeConnection(PineconeConfig config, String indexName) {
  this.config = config;
  String host = getHost(config.getApiKey(), indexName)
  channel = buildChannel(host);
  initialize();
}
```

I think `PineconeConnection` should probably also have a constructor that accepts `PineconeConfig` object as the sole parameter for cases where the user knows the host or has provided a ManagedChannel for us to use. Something like:

```java
public PineconeConnection(PineconeConfig config) {
  if (config.getManagedChannel() != null) {
    channel = config.getManagedChannel()
  } else {
    if (config.getHost() == null) {
       // throws exception
    }
    channel = buildChannel(config.getHost());
  }

   initialize();
}
```

jhamon (approved):
I'm surprised such a fundamental change didn't have a larger footprint in tests. It's hard to assess the impact to UX without seeing usage written out in a straightforward way in tests, so adding some more tests asap would be a big help.

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -0,0 +1,63 @@
+package io.pinecone.configs;
+
+import io.grpc.ManagedChannel;
+import io.pinecone.exceptions.PineconeValidationException;
+
+public class PineconeConfig {
+
+    private String apiKey;
+    private String host;
+    private String usageContext;
```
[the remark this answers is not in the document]
rohanshah18: We decided to add this for detecting user-agent string coming from spark connector. Initially it was added this in the PineconeClientConfig class.
https://github.com/pinecone-io/pinecone-java-client/pull/43

src/main/java/io/pinecone/configs/PineconeConnection.java
rohanshah18: We already have the constructor of type `PineconeConnection(PineconeConfig config, String indexName)` so what you're suggesting is to delete the constructor `PineconeConfig(String apiKey, String indexName)`
The reason it was added in the first place is because I was trying to replicate the python and node sdk behavior of being able to make data plane calls with just apiKey and indexName. 

So as a next commit, I'll delete `PineconeConfig(String apiKey, String indexName)` constructor and add `PineconeConfig(PineconeConfig config)`.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -30,7 +28,10 @@ public static PineconeConnection createIndexIfNotExistsDataPlane(int dimension,
 
         // Do not proceed until the newly created index is ready
         isIndexReady(indexName, controlPlaneClient);
-        return new PineconeConnection(apiKey, indexName);
+
+        // Adding to test PineconeConnection(pineconeConfig, indexName) constructor
+        PineconeConfig config = new PineconeConfig(apiKey);
+        return new PineconeConnection(config, indexName);
```
rohanshah18, line 34: @jhamon this is where we are using the new constructor for all of the data plane tests. If the indexes don't exist, we create it using the new PineconeConfig by passing config object and indexName.

src/main/java/io/pinecone/configs/PineconeConnection.java
jhamon: Yeah, I see now. Sorry I overlooked the `PineconeConnection(PineconeConfig config, String indexName)` that's already implemented. The way github renders the diff is a bit hard to follow.

src/main/java/io/pinecone/configs/PineconeConfig.java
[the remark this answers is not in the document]
austin-denoble: What do you think of naming this `integrationId`? That's what I went with when I implemented the same thing on the TypeScript side, although it hasn't landed yet: https://github.com/pinecone-io/pinecone-ts-client/pull/197

I think `usageContext` ends up feeling more broad, or certain users may see `context` and make assumptions about usage. We should probably also align the value in user-agents across clients. You'd have to update the `getUserAgent()` call as well, but everything else could stay the same.

austin-denoble (approved):
Thanks very much for cleaning up the `PineconeClient`, `PineconeClientConfig`, and `PineconeConnectionConfig` classes! I feel like things are going to be a lot easier to reason about and work with going forward.

You can keep the collection and configure index tests skipped for now, I'll pick those up for improving and fixing in a follow-up PR.

src/main/java/io/pinecone/configs/PineconeConfig.java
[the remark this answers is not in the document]
rohanshah18: I agree, given usage now means something, its not a good variable name. But doesn't integrationId refer to some sort of Id?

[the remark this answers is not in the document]
austin-denoble: I feel like `integration` should be represented in the name, since that's the intent of the field. `userContext` could still be misconstrued by users, and from my understanding this was meant to support passing specific integrations (`spark-connector`, `langchain`, etc) for us to evaluate usage.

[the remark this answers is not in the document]
jhamon: I think I also prefer `integrationId`. This is really aimed at use cases where partners like `langchain` or `llamaindex` want to build something on top of our stuff. Or high-level tools we choose to build ourselves on top of SDKs (e.g. canopy). `userContext` makes it seem like it's something for the typical end-user, but it's really not.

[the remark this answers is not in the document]
rohanshah18: Okay I'll make the change in next PR. I'll start an internal conversation regarding renaming clients as well.

PR #78, opened 2024-03-08
Regenerate gRPC classes after removing grpc-gateway / grpc -> OpenAPI spec metadata

austin-denoble:
## Problem
Currently, the [Java classes we generate](https://github.com/pinecone-io/pinecone-java-client/tree/main/src/main/java/io/pinecone/proto) from `vector_service.proto` include several lines of code which cause builds to fail, and are currently commented out.

Example of build errors after generation:
Screenshot 2024-03-08 at 4 16 45 PM

The gist is it seems like the `grpc-gateway` package is unable to be resolved during generation.

## Solution
We generate the Java gRPC classes using `protoc` and a plugin: `protoc-gen-grpc-java`. You can see the general approach here: https://github.com/pinecone-io/pinecone-protos/blob/main/scripts/sdk-grpc-generation/java/gen.sh

I talked with @haruska as I was having trouble getting the `grpc-gateway` to resolve, even though it seems to be provided in `/thirdparty/`. I tried pulling in the thirdparty protos manually and updating the paths, but it seems like the Java plugin we use may not support grpc-gateway, although I haven't been able to confirm this.

After talking with @haruska, he mentioned he had pulled out all of the grpc -> openapi spec information from the proto file before generation in the Go client: https://github.com/pinecone-io/go-pinecone/blob/main/apis/proto/pinecone/data/v1/vector_service.proto. 

I went with re-generating the Java classes using this same approach. The grpc-gateway bits should not be needed for the client code as we don't care about the REST -> gRPC conversion.

Jason mentioned we may want to look at generating this version of the spec pre-codegen for Java and Go.

## Type of Change

## Test Plan
Build + Integration Tests via CI


src/main/java/io/pinecone/proto/DeleteResponse.java
```diff
@@ -23,28 +23,28 @@ private DeleteResponse(com.google.protobuf.GeneratedMessageV3.Builder<?> builder
...
 
   public static final com.google.protobuf.Descriptors.Descriptor
       getDescriptor() {
-    return VectorServiceOuterClass.internal_static_DeleteResponse_descriptor;
+    return io.pinecone.proto.VectorServiceOuterClass.internal_static_DeleteResponse_descriptor;
   }
 
-  @Override
-  protected FieldAccessorTable
+  @java.lang.Override
```
rohanshah18, line 38: When I generate it locally, the Override is not replaced with java.lang.Override. I'm going to test it locally against serverless indexes and approve this PR soon.

rohanshah18: Also generating the code locally, there are no changes in the the following files: 
1. DeleteResponse
2. DeleteResponseOrBuilder
3. DescribeIndexStatsRequestOrBuilder
4. RequestUnion
5. RequestUnionOrBuilder
6. UpdateResponseOrBuilder

We can ignore files like DeleteResponseOrBuilder since you are using different protoc version but I'm not sure about the other changes.

rohanshah18: Okay I have tested with serverless indexes and things look good. Since Override is part of Java.lang, I'm not concerned why its expanding when you generate the code.

rohanshah18 (approved):
Thanks for handling this code generation issue. Tested it against serverless indexes as well, so we are good to go.

jhamon (approved):

PR #79, opened 2024-03-12
Fix Collection and ConfigureIndex tests

austin-denoble:
## Problem
Currently, `CollectionTest`, and `ConfigureIndexTest` are very flakey. Collections tests take a long while to run due to the need to spin up a collection, wait until it's ready, and then creating an index and waiting until that's ready.

`ConfigureIndexTest` deletes the index as a part of the final `@Test` in the file, but it doesn't seem like the order of execution is guaranteed, so deleting an index like this may cause issues: https://www.baeldung.com/junit-5-test-order

## Solution
- Update `CollectionTest` to create a single collection from the initial index that's shared across both tests. Previously we were creating a collection for each test, which is not ultimately necessary and is very time-consuming. Also, use `assertWithRetry()` to retry upserting and describing index stats. These calls seemingly have a tendency to flake when made against a newly created index. Retrying a few times if we have an error on the initial pass seems to help here.
- `ConfigureIndexTest` has a `finally` clause in the last `@Test` in the file deletes the index. Given we set up with `@BeforeAll` and the execution order is not guaranteed unless we specify, it feels like it makes more sense to clean things up in `@AfterAll`, which seems to resolve the issue with the test itself.

## Type of Change

- [x] None of the above: (integration tests)

## Test Plan
Validate integration runs in CI. Runs should overall be a few minutes faster, and less prone to failing on these specific tests.

### DIFF
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -23,7 +23,6 @@ import static io.pinecone.helpers.BuildUpsertRequest.*;
 import static io.pinecone.helpers.AssertRetry.assertWithRetry;
 import static org.junit.jupiter.api.Assertions.*;
 
-@Disabled("Disable the entire class")
 public class CollectionTest {
     private static final String indexName = RandomStringBuilder.build("collection-test", 8);
     private static final String collectionName = RandomStringBuilder.build("collection-test", 8);
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -20,65 +20,81 @@ import static io.pinecone.helpers.IndexManager.createNewIndexAndConnect;
 import static io.pinecone.helpers.IndexManager.waitUntilIndexIsReady;
 import static io.pinecone.helpers.IndexManager.createCollection;
 import static io.pinecone.helpers.BuildUpsertRequest.*;
+import static io.pinecone.helpers.AssertRetry.assertWithRetry;
 import static org.junit.jupiter.api.Assertions.*;
 
 @Disabled("Disable the entire class")
 public class CollectionTest {
-
-    private static PineconeControlPlaneClient controlPlaneClient;
     private static final String indexName = RandomStringBuilder.build("collection-test", 8);
+    private static final String collectionName = RandomStringBuilder.build("collection-test", 8);
     private static final ArrayList<String> indexes = new ArrayList<>();
-    private static final ArrayList<String> collections = new ArrayList<>();
     private static final IndexMetric indexMetric = IndexMetric.COSINE;
     private static final List<String> upsertIds = Arrays.asList("v1", "v2", "v3");
     private static final String namespace = RandomStringBuilder.build("ns", 8);
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
     private static final String environment = System.getenv("PINECONE_ENVIRONMENT");
     private static final int dimension = 4;
-
     private static final Logger logger = LoggerFactory.getLogger(CollectionTest.class);
+    private static PineconeControlPlaneClient controlPlaneClient;
+    private static CollectionModel collection;
 
     @BeforeAll
     public static void setUp() throws InterruptedException {
         controlPlaneClient = new PineconeControlPlaneClient(apiKey);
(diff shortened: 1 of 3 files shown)


src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -20,65 +20,81 @@
...
 import static io.pinecone.helpers.BuildUpsertRequest.*;
+import static io.pinecone.helpers.AssertRetry.assertWithRetry;
 import static org.junit.jupiter.api.Assertions.*;
 
 @Disabled("Disable the entire class")
 public class CollectionTest {
-
-    private static PineconeControlPlaneClient controlPlaneClient;
     private static final String indexName = RandomStringBuilder.build("collection-test", 8);
+    private static final String collectionName = RandomStringBuilder.build("collection-test", 8);
```
austin-denoble, line 29: Previously, each test in this file created a new collection, which was unnecessary and time consuming. We can create a single collection in our setup and share across both tests to increase speed.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -20,65 +20,81 @@
...
+                new CreateIndexRequestSpecPod().pods(1).podType("p1.x1").replicas(1).environment(environment);
         CreateIndexRequestSpec spec = new CreateIndexRequestSpec().pod(podSpec);
-        PineconeConnection dataPlaneConnection = createNewIndexAndConnect(controlPlaneClient, indexName, dimension, indexMetric, spec);
+        PineconeConnection dataPlaneConnection = createNewIndexAndConnect(controlPlaneClient, indexName, dimension,
+                indexMetric, spec);
         VectorServiceGrpc.VectorServiceBlockingStub blockingStub = dataPlaneConnection.getBlockingStub();
         indexes.add(indexName);
 
-        blockingStub.upsert(buildRequiredUpsertRequestByDimension(upsertIds, dimension, namespace));
+        // Sometimes we see grpc failures when upserting so quickly after creating, so retry
```
austin-denoble, line 54: This seems to help some of the flake we see when trying to upsert to a fresh index.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -20,65 +20,81 @@
...
-    public void testIndexToCollectionHappyPath() throws InterruptedException {
-        String collectionName = RandomStringBuilder.build("collection-test", 8);
+        // Verify we can delete the collection
+        controlPlaneClient.deleteCollection(collectionName);
+        Thread.sleep(2500);
 
-        // Create collection from index
-        CollectionModel collection = createCollection(controlPlaneClient, collectionName, indexName, true);
-        collections.add(collectionName);
+        List<CollectionModel> collectionList = controlPlaneClient.listCollections().getCollections();
```
austin-denoble, line 80: Maybe this is overkill and we can just send the request to delete, but we used to validate the delete using list. I just moved that into the cleanup.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -137,10 +143,6 @@ public void sizeIncrease() throws InterruptedException {
             });
         } catch (Exception exception) {
             throw new PineconeException("Test failed: " + exception.getLocalizedMessage());
-        } finally {
```
austin-denoble, line 140: I don't think `@Test` methods in a class are guaranteed to run in the same order every time, so I don't think this is guaranteed to run last which could cause issues.

Moving the clean up to `@AfterAll` helps with this.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
rohanshah18: Great, thanks so much for adding this change!

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
rohanshah18: Is it necessary to upsert data here?

rohanshah18 (approved):
Thanks for fixing this, I have left one comment but otherwise it looks great!

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
austin-denoble: I'm creating a new pod index for these tests rather than using `createNewIndexAndConnect()` because I need to use the index name and metric in my tests, which was not easy to do with `createIndexIfNotExistsDataPlane()`.

PR #81, opened 2024-03-13
Refactor `CollectionTest`, `ConfigureIndexTest`, and `IndexManager` to improve integration test speed and reliability

austin-denoble:
## Problem
We've had continued issues with flakiness of integration tests, and tests taking a while to run (25 - 30 minutes for the entire `pr` workflow in some cases). A lot of this is waiting for indexes and collections to be `Ready` instead of `Initializing`. Specifically,  the `CollectionTest` and `CollectionErrorTest` can take a really long time due to setting up and waiting on multiple indexes and collections.

While working on the above issue, I also noticed test failures and repeated issues due to `findIndexWithDimensionAndType`, and tests running with indexes created by and belonging to other test runs. The `isIndexReady` function also tends to be inefficient as the polling isn't consistent, and we can end up waiting sometimes minutes longer than we may need to.

With the new `Pinecone` and `Index`/`AsyncIndex` class structure in place, we can refactor our integration tests around the new pattern. 

## Solution
Ultimately, I would like to move our tests away from the `createIndexIfNotExistsDataPlane` / `createIndexIfNotExistsControlPlane` functions and their reliance on the `findIndexWithDimensionAndType` function, which should be deprecated. We should come up with a pattern for sharing resources across all tests in a run, and setting up / tearing those down once. This can be worked on in subsequent PRs.

- Move `CollectionErrorTests` tests into `CollectionTest`. Share index and collection setup `@BeforeAll` across collections tests. Wait less time for the indexes created from the collection to be ready as this specifically can take a number of minutes. Relax some of the assertions on the `status` of created collections and indexes as I don't think we need to be that thorough here.
- Update `createIndexIfNotExistsDataPlane` to return a tuple (`AbstractMap.SimpleEntry<String, Pinecone>`) of the `indexName` and the `Pinecone` client to make things a bit easier to work with.
- Update all of the `dataPlane/` tests to use `pineconeClient.createIndexConnection()` and `pineconeClient.createAsyncIndexConnection()` for managing data plane operations.
- Deprecate `isIndexReady` in lieu of `waitUntilIndexIsReady`. The polling is more consistent with this function, and it most likely saves us time overall.
- Refactor `ConfigureIndexTest` to create its own index and clean up after.
- Fix issue with `findIndexByDimensionAndType` calling `isIndexReady()` on each index while iterating through the list.
- Fix issue in `IndexInterface.validateUpsertRequest` where we were trying to call `sparseValuesWithUnsignedIndices.getIndicesWithUnsigned32IntList()` and `sparseValuesWithUnsignedIndices.getValuesList()` on a possible `null` causing a `NullPointerException`.
- Talked with @ssmith-pc, and I think it's standard practice in tests to not `try`/`catch` yourself unless you need to assert on the result. We should be letting errors throw to the test runner and let it handle them so we're not clobbering logs and stack traces. I've cleaned up `try`/`catch` statements which don't seem to be needed.
- Running `gradle integrationTest --info` in `pr.yml` to get more detailed log output in the console for better troubleshooting of ongoing flapping.
- Adding `assertWithRetry` wrappers for specific actions which have been troublesome. Adding `Thread.sleep()` to a few places to avoid hammering an index too quickly / etc.
- Added a new `describeIndexStats()` overload to `IndexInterface` and `Index` / `AsyncIndex` to allow calling without needing to explicitly pass `null`.
 
I spent a lot of time running these locally and in CI to see how they perform. Overall, it seems like these changes improve overall reliability, although we still do see a few failures like the gRPC `no healthy upstream` on data plane operations with fresh indexes.

The total amount of time it takes to run both sets of integration tests has been cut significantly in most cases:

Screenshot 2024-03-13 at 6 09 36 PM

Next steps would be to create a Junit extension and possible `IndexManagerSingleton` to manage index resources across all tests directly. This would help make things more predictable and reliable when adding future tests. This would also allow us to handle concurrent integration test runs under the same API key, which is currently very difficult due to `findIndexByDimensionAndType`.

## Type of Change

## Test Plan
Run integration tests and compare over run length and


src/main/java/io/pinecone/commons/IndexInterface.java
```diff
@@ -38,13 +38,13 @@ default UpsertRequest validateUpsertRequest(List<VectorWithUnsignedIndices> vect
...
 
-            Vector vector = buildUpsertVector(vectorWithUnsignedIndices.getId(),
-                    vectorWithUnsignedIndices.getValuesList(),
-                    sparseValuesWithUnsignedIndices.getIndicesWithUnsigned32IntList(),
-                    sparseValuesWithUnsignedIndices.getValuesList(),
-                    vectorWithUnsignedIndices.getMetadata());
+                Vector vector = buildUpsertVector(vectorWithUnsignedIndices.getId(),
+                        vectorWithUnsignedIndices.getValuesList(),
+                        (sparseValuesWithUnsignedIndices != null) ? sparseValuesWithUnsignedIndices.getIndicesWithUnsigned32IntList() : null,
+                        (sparseValuesWithUnsignedIndices != null) ? sparseValuesWithUnsignedIndices.getValuesList() : null,
```
austin-denoble, line 44: We were getting `NullPointerExceptions` here if the user had passed `null` for these values in the `VectorWithUnsignedIndices` class.

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorPodTest.java
```diff
@@ -23,46 +25,44 @@
...
-    }
-
-    @AfterAll
-    public static void cleanUp() {
-        connection.close();
+        AbstractMap.SimpleEntry<String, Pinecone> indexAndClient = createIndexIfNotExistsDataPlane(dimension, IndexModelSpec.SERIALIZED_NAME_POD);
+        indexName = indexAndClient.getKey();
+        pineconeClient = indexAndClient.getValue();
+        indexClient = pineconeClient.createIndexConnection(indexName);
+        asyncIndexClient = pineconeClient.createAsyncIndexConnection(indexName);
```
austin-denoble, line 41: I made this same change consuming the returned tuple from `createIndexIfNotExistsDataPlane` across all of these tests in this folder. It feels a bit easier to reason about, and we test the helper functions on `Pinecone`.

src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
```diff
@@ -135,21 +136,15 @@ public static UpsertRequest buildRequiredUpsertRequest(List<String> upsertIds, S
                 .build();
     }
 
-    public static UpsertRequest buildRequiredUpsertRequestByDimension(List<String> upsertIds, int dimension, String namespace) {
+    public static List<VectorWithUnsignedIndices> buildRequiredUpsertRequestByDimension(List<String> upsertIds, int dimension) {
```
austin-denoble, line 139: Updated to handle generating vectors for the new `Index.upsert()` function.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -19,62 +20,46 @@
 public class IndexManager {
     private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
 
-    public static PineconeConnection createIndexIfNotExistsDataPlane(int dimension, String indexType) throws IOException, InterruptedException {
+    public static AbstractMap.SimpleEntry<String, Pinecone> createIndexIfNotExistsDataPlane(int dimension, String indexType) throws IOException, InterruptedException {
```
austin-denoble, line 23: This function now returns a tuple of `[indexName, pineconeClient]`.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -19,62 +20,46 @@
...
         String apiKey = System.getenv("PINECONE_API_KEY");
         Pinecone pinecone = new Pinecone(apiKey);
 
         String indexName = findIndexWithDimensionAndType(pinecone, dimension, indexType);
-        if (indexName.isEmpty()) indexName = createNewIndex(pinecone, dimension, indexType);
+        if (indexName.isEmpty()) indexName = createNewIndex(pinecone, dimension, indexType, true);
```
austin-denoble, line 28: Updated `createNewIndex` to allow passing a boolean for `waitUntilIndexIsReady`.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -19,62 +20,46 @@
...
+        for (IndexModel indexModel : indexModels) {
             if (indexModel.getDimension() == dimension
-                    && ((indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD)
-                        && indexModel.getSpec().getPod() != null
-                        && indexModel.getSpec().getPod().getReplicas() == 1
-                        && indexModel.getSpec().getPod().getPodType().equalsIgnoreCase("p1.x1"))
-                    || (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS)))) {
+                && (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null)
+                || (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS) && indexModel.getSpec().getServerless() != null)
+            ) {
```
austin-denoble, line 44: I was seeing issues related to this function finding indexes for the API key that were unrelated to the test that was running. I wanted to make this a bit easier to reason about, and I think we should eventually phase it out.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -19,62 +20,46 @@
...
 
         if (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD)) {
-            CreateIndexRequestSpecPod podSpec = new CreateIndexRequestSpecPod().environment(environment).podType("p1.x1");
+            CreateIndexRequestSpecPod podSpec = new CreateIndexRequestSpecPod().environment(environment).podType("p1" +
+                    ".x1");
             createIndexRequestSpec = new CreateIndexRequestSpec().pod(podSpec);
         } else {
-            ServerlessSpec serverlessSpec = new ServerlessSpec().cloud(ServerlessSpec.CloudEnum.AWS).region(environment);
+            ServerlessSpec serverlessSpec =
+                    new ServerlessSpec().cloud(ServerlessSpec.CloudEnum.AWS).region("us-west-2");
```
austin-denoble, line 62: This needs to be either `us-west-2` for now, we shouldn't be relying on the `environment` value because it's often not correct in runs or when running manually.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -85,46 +70,61 @@ public static String createNewIndex(Pinecone pinecone, int dimension, String ind
...
             Thread.sleep(intervalMs);
+            logger.info("Waited " + waitedTimeMs + "ms for " + indexName + " to get ready");
             waitedTimeMs += intervalMs;
         }
         return index;
     }
 
     public static IndexModel waitUntilIndexIsReady(Pinecone pinecone, String indexName) throws InterruptedException {
-        return waitUntilIndexIsReady(pinecone, indexName, 120000);
+        return waitUntilIndexIsReady(pinecone, indexName, 200000);
```
austin-denoble, line 105: Bumped the default time a bit.

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -85,46 +70,61 @@ public static String createNewIndex(Pinecone pinecone, int dimension, String ind
...
-    public static PineconeConnection createNewIndexAndConnect(Pinecone pinecone, String indexName, int dimension, IndexMetric metric, CreateIndexRequestSpec spec) throws InterruptedException, PineconeException {
-        String apiKey = System.getenv("PINECONE_API_KEY");
+    public static Pinecone createNewIndex(Pinecone pinecone, String indexName, int dimension, IndexMetric metric, CreateIndexRequestSpec spec, boolean waitUntilIndexIsReady) throws InterruptedException, PineconeException {
+        CreateIndexRequest createIndexRequest = new CreateIndexRequest().name(indexName).dimension(dimension).metric(metric).spec(spec);
+        pinecone.createIndex(createIndexRequest);
+
+        if (waitUntilIndexIsReady) {
+            waitUntilIndexIsReady(pinecone, indexName);
+        }
+        return pinecone;
```
austin-denoble, line 115: New `waitUntilIndexIsReady` `boolean` plus we return the `Pinecone` client now instead of a connection, the consumer can use `indexName` to create a connection.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -50,58 +56,51 @@ public void configureIndexExceedingQuota() {
...
-            // Verify replicas were scaled down
-            assertWithRetry(() -> {
-                PodSpec podSpec = controlPlaneClient.describeIndex(indexName).getSpec().getPod();
-                assert (podSpec != null);
-                assertEquals(podSpec.getReplicas(), 1);
-            });
-        } catch (Exception exception) {
-            throw new PineconeException("Test failed: " + exception.getLocalizedMessage());
-        }
+    public void scaleUpAndDown() throws InterruptedException {
```
austin-denoble, line 66: Removed `try`/`catch`.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -112,33 +111,27 @@ public void changingBasePodType() {
...
     @Test
-    public void sizeIncrease() {
-        try {
-            // Verify the starting state
-            IndexModel indexModel = isIndexReady(indexName, controlPlaneClient);
-            assert indexModel.getSpec().getPod() != null;
-            assertEquals("p1.x1", indexModel.getSpec().getPod().getPodType());
+    public void sizeIncrease() throws InterruptedException {
+        // Verify the starting state
+        IndexModel indexModel = controlPlaneClient.describeIndex(indexName);
```
austin-denoble, line 120: Removed `try`/`catch`.

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryPodTest.java
```diff
@@ -185,22 +185,16 @@ public void upsertOptionalVectorsAndQueryIndexFutureTest() throws InterruptedExc
...
         String id = RandomStringBuilder.build(3);
         try {
-            dataPlaneClient.upsert(id,
+            asyncIndexClient.upsert(id,
                     generateVectorValuesByDimension(dimension),
                     null,
                     generateVectorValuesByDimension(dimension),
                     null,
                     null).get();
-        } catch (PineconeValidationException validationException) {
```
austin-denoble, line 198: I don't think this was needed, I think we can just use one `catch` statement.

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryServerlessTest.java
```diff
@@ -74,7 +74,7 @@ public void upsertOptionalVectorsAndQueryIndexSyncTest() throws Exception {
         index.upsert(vectors, namespace);
 
         // wait sometime for the vectors to be upserted
-        Thread.sleep(5000);
+        Thread.sleep(7500);
```
austin-denoble, line 77: This and the change below was to give a bit more time for freshness, which can be an issue with serverless indexes.

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryServerlessTest.java
```diff
@@ -211,26 +208,21 @@ public void upsertOptionalVectorsAndQueryIndexFutureTest() throws InterruptedExc
...
         String id = RandomStringBuilder.build(3);
-        StringBuilder exceptionMessage = new StringBuilder();
         try {
             asyncIndex.upsert(id,
                     generateVectorValuesByDimension(dimension),
                     null,
                     generateVectorValuesByDimension(dimension),
                     null,
                     null).get();
-        } catch (PineconeValidationException validationException) {
```
austin-denoble, line 228: Same as the previous change similar to this, I think we can just use one `catch`.

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -192,6 +192,13 @@ public ListenableFuture<DeleteResponse> delete(List<String> ids,
         return asyncStub.delete(deleteRequest);
     }
 
+    @Override
+    public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats() {
```
austin-denoble, line 196: Just a convenience for allowing users to call it without passing an explicit `null`.

.github/workflows/pr.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       run: gradle clean build
 
     - name: Run integration tests
-      run: gradle integrationTest
+      run: gradle integrationTest --info
```
austin-denoble, line 83: I'd prefer to leave this level of logging on for now so we can root cause test failures with a bit more ease.

src/integration/java/io/pinecone/helpers/AssertRetry.java
```diff
@@ -10,7 +10,7 @@ public class AssertRetry {
     private static final int delay = 1500;
 
     public static void assertWithRetry(AssertionRunnable assertionRunnable) throws InterruptedException, PineconeException {
-        assertWithRetry(assertionRunnable, 2);
+        assertWithRetry(assertionRunnable, 3);
```
austin-denoble, line 13: Bumped the default backoff on this.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -50,58 +54,51 @@ public void configureIndexExceedingQuota() {
         ConfigureIndexRequestSpec spec = new ConfigureIndexRequestSpec().pod(pod);
         ConfigureIndexRequest configureIndexRequest = new ConfigureIndexRequest().spec(spec);
         try {
-            isIndexReady(indexName, controlPlaneClient);
             controlPlaneClient.configureIndex(indexName, configureIndexRequest);
         } catch (PineconeForbiddenException forbiddenException) {
             assert (forbiddenException.getLocalizedMessage().contains("Increase your quota or upgrade to create more indexes."));
```
ssmith-pc, line 59: In all these cases where we're *expecting* a failure, we need to make sure that the test fails if the exception is not thrown. Here's a general example:

```java
try {
    // make the call that should result in an exception 
    controlPlanClient.configureIndex(indexName, configureIndexRequest);

    // call Assert.fail if we get to this point the code should have thrown,
    // but didn't, so we need to fail the test.
    fail("Should have throw PineconeForbiddenException");

} (catch PineconeForbiddenException expected) { 
    // I like naming the exception as "expected" to make it obvious it's the test success path

    // Do nothing.

    // (Or possibly validate the contents of the exception, if it matters to the client.) 
    // But in this case, do we actually care if the string error message changes? Probably not
}
```

src/integration/java/io/pinecone/helpers/IndexManager.java
ssmith-pc: Can we leave a comment to explain the need to hard-code the environment

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
ssmith-pc: Alternatively, you can wrap the call in `Assert.assertThrows` like:

```java
assertThrows(PineconeForbiddenException.class,
     controlPlanClient.configureIndex(indexName, configureIndexRequest));
```

But that's a little less flexible especially when you'd like to do some validation on the exception itself.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -178,14 +172,96 @@ public void testIndexFromDifferentMetricCollection() throws InterruptedException
...
+            CreateIndexRequestSpecPod specPod = new CreateIndexRequestSpecPod().pods(1).podType("p1.x1").replicas(1).environment(environment);
+            CreateIndexRequestSpec spec = new CreateIndexRequestSpec().pod(specPod);
+            CreateIndexRequest createIndexRequest = new CreateIndexRequest().name(notReadyIndexName).dimension(dimension).metric(IndexMetric.COSINE).spec(spec);
+            pineconeClient.createIndex(createIndexRequest);
+            indexesToCleanUp.add(notReadyIndexName);
+
+            createCollection(pineconeClient, newCollectionName, notReadyIndexName, false);
+        } catch (PineconeException exception) {
+            logger.info("Exception: " + exception.getMessage());
+            assert (exception.getMessage().contains("Source index is not ready"));
```
ssmith-pc, line 264: In general using the built-in `assert` mostly maps directly to JUnit `Assert.assertTrue`, it's preferable to avoid using the built-in. 

The reason is b/c it's possible to disable assertion checking in the JVM as a performance optimization. This is unlikely to be disabled in test runs, but you can avoid the chance altogether by just avoiding built-in assert statements.

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorPodTest.java
ssmith-pc: Yeah, this is nice. In the near future I can see this moving to an extension that runs at the beginning of the test run and injected for all tests so we only pay the cost once

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorPodTest.java
```diff
@@ -23,46 +23,44 @@
...
-        DescribeIndexStatsResponse describeIndexStatsResponse1 = index.describeIndexStats(null);
+        DescribeIndexStatsResponse describeIndexStatsResponse1 = indexClient.describeIndexStats(null);
         assertEquals(describeIndexStatsResponse1.getDimension(), dimension);
 
-        StringBuilder exceptionMessage = new StringBuilder();
         // Query with incorrect dimensions
         try {
             List<Float> vector = Arrays.asList(100F);
-            index.query(5, vector, null, null, null, namespace, null, true, true);
+            indexClient.query(5, vector, null, null, null, namespace, null, true, true);
```
ssmith-pc, line 51: Need to fail if it doesn't throw

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorPodTest.java
```diff
@@ -76,30 +74,25 @@ public void QueryWithNullSparseIndicesNotNullSparseValuesSyncTest() {
...
-        DescribeIndexStatsResponse describeIndexStatsResponse1 = asyncIndex.describeIndexStats(null).get();
+        DescribeIndexStatsResponse describeIndexStatsResponse1 = asyncIndexClient.describeIndexStats(null).get();
         assertEquals(describeIndexStatsResponse1.getDimension(), dimension);
 
-        StringBuilder exceptionMessage = new StringBuilder();
         // Query with incorrect dimensions
         try {
             List<Float> vector = Arrays.asList(100F);
-            asyncIndex.query(5, vector, null, null, null, namespace, null, true, true).get();
+            asyncIndexClient.query(5, vector, null, null, null, namespace, null, true, true).get();
```
ssmith-pc, line 83: Need to fail if it doesn't throw

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryPodTest.java
ssmith-pc: It is a bit cleaner if we just declare the exceptions in the test method signature that we don't want to handle (i.e. we want them to cause a failure if they're thrown)

Example:
```java
public void upsertNullSparseIndicesNotNullSparseValuesFutureTest() throws ExecutionException, InterruptedException {
  /// ...
  try {
     // ...

    fail("should have thrown PVE")

  } catch (PVE expected) {
    // validate exception message
  }
```

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryServerlessTest.java
ssmith-pc: Same comment as prev

ssmith-pc (commented):
In general these changes look good and should improve the reliability and speed of the test runs. 

I added some various test structure comments throughout

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
austin-denoble: I think @rohanshah18 had also mentioned this to me previously. Going through and updating these.

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryPodTest.java
austin-denoble: Ah thank you! I wasn't very clear on how to properly represent this.

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: Yes, good call. Public preview launched for serverless with one region `us-west-2` which is mostly hard-coded in tests at the moment.

rohanshah18 (approved):
Thanks for going through all of the changes. Looks like the collections tests are still failing. Also updating the CI to output more logs will add more noise and its not easily clear what tests are failing. I think having the short summary of what assertions failed with line numbers is cleaner and we can reproduce them locally but if this is helpful to you, we can keep add more info temporarily.

.github/workflows/pr.yml
austin-denoble: I don't think running the integration suite locally is a good substitute for debugging problems that happen in CI, and given that a lot of the errors we see are transitory and are not likely to be immediately reproducible by re-running locally, I think logging additional information as a part of CI is important.

src/integration/java/io/pinecone/helpers/AssertRetry.java
austin-denoble: I understand what the second parameter does here, and I also understand this increases the amount of time the asserts will wait overall. I bumped this up because we are essentially using this utility to poll for freshness in some of our data plane tests, and I was concerned it may not be waiting long enough because I was seeing occasional failures that indicated this was the case.

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: I wanted this function to be easier to work with and allow us to use `Pinecone` rather than `PineconeConnection` in consuming tests. Looks like you found the reasoning though. 👍

austin-denoble:
> Looks like the collections tests are still failing. Also updating the CI to output more logs will add more noise and its not easily clear what tests are failing. I think having the short summary of what assertions failed with line numbers is cleaner and we can reproduce them locally but if this is helpful to you, we can keep add more info temporarily.

I left a comment about this, but I don't think running locally is a substitute for debugging what is happening in CI. We should be able to look at the CI logs, and have a somewhat clear idea what the issue is. You can pretty easily search through the logs for `FAILED` to see which specific tests are failing.

.github/workflows/pr.yml
ssmith-pc: What's the downside of enabling logging for CI runs? Do we pay more?

src/integration/java/io/pinecone/helpers/AssertRetry.java
austin-denoble: Moved this back down to 2, let's keep an eye on data plane tests needing to poll for freshness and whether this works. 👍

PR #82, opened 2024-03-19
Add Create Index Error + Optional-argument integration tests for Pod + Serverless

austin-denoble:
## Problem
The `createIndex` integration tests are somewhat minimal and we've wanted to add thorough testing for error-cases, and optional configuration arguments for both pod and serverless indexes.

## Solution
Update `integration.controlPlane.pod.CreateDescribeAndDeleteIndexTest` and `integration.controlPlane.serverless.CreateDescribeAndDeleteIndexTest` to include additional tests validating error-conditions, and creating indexes with optional arguments.

## Type of Change

## Test Plan
run `gradle integrationTest` or validate the tests are run as part of the `pr` workflow in GitHub CI.

### DIFF
--- a/src/integration/java/io/pinecone/helpers/IndexManager.java
+++ b/src/integration/java/io/pinecone/helpers/IndexManager.java
@@ -54,7 +54,7 @@ public class IndexManager {
             CreateIndexRequestSpecPod podSpec = new CreateIndexRequestSpecPod().environment(environment).podType("p1.x1");
             createIndexRequestSpec = new CreateIndexRequestSpec().pod(podSpec);
         } else {
-            // Serverless currently has limited availability in specific regions, hardcode us-west-2 for now
+            // Serverless currently has limited availability in specific regions, hard-code us-west-2 for now
             ServerlessSpec serverlessSpec =
                     new ServerlessSpec().cloud(ServerlessSpec.CloudEnum.AWS).region("us-west-2");
             createIndexRequestSpec = new CreateIndexRequestSpec().serverless(serverlessSpec);
--- a/src/integration/java/io/pinecone/helpers/IndexManager.java
+++ b/src/integration/java/io/pinecone/helpers/IndexManager.java
@@ -112,7 +112,7 @@ public class IndexManager {
         return pinecone;
     }
 
-    public static Index createNewIndexAndConnect(Pinecone pinecone, String indexName, int dimension, IndexMetric metric, CreateIndexRequestSpec spec) throws InterruptedException, PineconeException {
+    public static Index createNewIndexAndConnectSync(Pinecone pinecone, String indexName, int dimension, IndexMetric metric, CreateIndexRequestSpec spec) throws InterruptedException, PineconeException {
         CreateIndexRequest createIndexRequest = new CreateIndexRequest().name(indexName).dimension(dimension).metric(metric).spec(spec);
         pinecone.createIndex(createIndexRequest);
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
@@ -1,51 +1,181 @@
 package io.pinecone.integration.controlPlane.pod;
 
 import io.pinecone.clients.Pinecone;
+import io.pinecone.exceptions.PineconeBadRequestException;
+import io.pinecone.exceptions.PineconeUnmappedHttpException;
 import io.pinecone.helpers.RandomStringBuilder;
-import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.openapitools.client.model.*;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import java.util.Random;
+
+import static io.pinecone.helpers.IndexManager.waitUntilIndexIsReady;
+import static org.junit.jupiter.api.Assertions.*;
 
 public class CreateDescribeListAndDeleteIndexTest {
-    private Pinecone controlPlaneClient;
-    private final String apiKey = System.getenv("PINECONE_API_KEY");
-    private final String environment = System.getenv("PINECONE_ENVIRONMENT");
+    private static final String environment = System.getenv("PINECONE_ENVIRONMENT");
+    private static final String indexName = RandomStringBuilder.build("create-index", 8);
+    private static Pinecone controlPlaneClient = new Pinecone(System.getenv("PINECONE_API_KEY"));
 
-    @BeforeEach
-    public void setUp() {
-        controlPlaneClient = new Pinecone(apiKey);
-    }
-
-    @Test
-    public void createAndDelete() throws InterruptedException {
-        String indexName = RandomStringBuilder.build("index-name", 8);
+    @BeforeAll
+    public static void setUp() throws InterruptedException {
+        // Create the index
         CreateIndexRequestSpecPod podSpec = new CreateIndexRequestSpecPod().environment(environment).podType("p1.x1");
         CreateIndexRequestSpec createIndexRequestSpec = new CreateIndexRequestSpec().pod(podSpec);
-
-        // Create the index
         CreateIndexRequest createIndexRequest = new CreateIndexRequest()
                 .name(indexName)
                 .metric(IndexMetric.COSINE)
                 .dimension(10)
                 .spec(createIndexRequestSpec);
         controlPlaneClient.createIndex(createIndexRequest);
+        waitUntilIndexIsReady(controlPlaneClient, indexName);
+    }
 
+    @AfterAll
+    public static void cleanUp() {
+        // Delete the index
+        controlPlaneClient.deleteIndex(indexName);
+    }
+
+    @Test
+    public void describeAndListIndex() {
         // Describe the index
         IndexModel indexModel = controlPlaneClient.describeIndex(indexName);
         assertNotNull(indexModel);
         assertEquals(10, indexModel.getDimension());
         assertEquals(indexName, indexModel.getName());
         assertEquals(IndexMetric.COSINE, indexModel.getMetric());
+        assertNotNull(indexModel.getSpec().getPod());
+        assertEquals("p1.x1", indexModel.getSpec().getPod().getPodType());
 
         // List the index
         IndexList indexList = controlPlaneClient.listIndexes();
         assertNotNull(indexList.getIndexes());
+        assertTrue(indexList.getIndexes().stream().anyMatch(index -> indexName.equals(index.getName())));
+    }
 
-        // Delete the index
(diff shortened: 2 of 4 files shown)


src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -112,7 +112,7 @@ public static Pinecone createNewIndex(Pinecone pinecone, String indexName, int d
         return pinecone;
     }
 
-    public static Index createNewIndexAndConnect(Pinecone pinecone, String indexName, int dimension, IndexMetric metric, CreateIndexRequestSpec spec) throws InterruptedException, PineconeException {
+    public static Index createNewIndexAndConnectSync(Pinecone pinecone, String indexName, int dimension, IndexMetric metric, CreateIndexRequestSpec spec) throws InterruptedException, PineconeException {
```
aulorbe, line 115: Is the only change here the addition of the "Sync" part of the name?

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,51 +1,181 @@
...
+    private static final String indexName = RandomStringBuilder.build("create-index", 8);
+    private static Pinecone controlPlaneClient = new Pinecone(System.getenv("PINECONE_API_KEY"));
 
-    @Test
-    public void createAndDelete() throws InterruptedException {
-        String indexName = RandomStringBuilder.build("index-name", 8);
+    @BeforeAll
+    public static void setUp() throws InterruptedException {
+        // Create the index
         CreateIndexRequestSpecPod podSpec = new CreateIndexRequestSpecPod().environment(environment).podType("p1.x1");
```
aulorbe, line 25: Total nit, but would it be more idiomatic to have a constant that holds the `podType` instead of hardcoding it here & then again in the `describeAndListIndex` test below? (Not sure if Java has a concept of constants, tbf)

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,51 +1,181 @@
...
+                .metric(IndexMetric.COSINE)
+                .dimension(10)
+                .spec(createIndexRequestSpec);
+
+        try {
+            controlPlaneClient.createIndex(createIndexRequest);
+
+            fail("Expected to throw PineconeBadRequestException");
+        } catch (PineconeBadRequestException expected) {
+            assertTrue(expected.getLocalizedMessage().contains("must consist of lower case alphanumeric characters or" +
```
aulorbe, line 96: Nit: you might want to change the error msg here to say *what* must consist of lowercase (one word) chars... so something like "Index name must consist..." or similar

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,51 +1,181 @@
...
+                .metric(IndexMetric.COSINE)
+                .dimension(-1)
+                .spec(createIndexRequestSpec);
+
+        try {
+            controlPlaneClient.createIndex(createIndexRequest);
+
+            fail("Expected to throw PineconeUnmappedHttpException");
+        } catch (PineconeUnmappedHttpException expected) {
+            assertTrue(expected.getLocalizedMessage().contains("dimension: invalid value"));
```
aulorbe, line 116: Maybe this error message could be more helpful if it said what is invalid about the user's dimension. So something like "Dimension invalid, must be integer > 0" or something

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,51 +1,181 @@
...
+                .metric(IndexMetric.COSINE)
+                .dimension(10)
+                .spec(createIndexRequestSpec);
+
+        try {
+            controlPlaneClient.createIndex(createIndexRequest);
+
+            fail("Expected to throw PineconeBadRequestException");
+        } catch (PineconeBadRequestException expected) {
+            assertTrue(expected.getLocalizedMessage().contains("pods"));
```
aulorbe, line 136: This might be a totally normal Java convention, so apologies if this is obvious, but why do you have an assert stmt for bot the `"pods"` part of the message and then the rest of it (`"must be greater than 0"`)?

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,51 +1,181 @@
...
+                .dimension(10)
+                .spec(createIndexRequestSpec);
+
+        try {
+            controlPlaneClient.createIndex(createIndexRequest);
+
+            fail("Expected to throw PineconeBadRequestException");
+        } catch (PineconeBadRequestException expected) {
+            assertTrue(expected.getLocalizedMessage().contains("total pods must be divisible by number of shards"));
+        }
```
aulorbe, line 179: Maybe worth adding a test to confirm an invalid similarity metric also raises an error? Just since you're doing tests for all the other index-creation params

src/integration/java/io/pinecone/integration/controlPlane/serverless/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,52 +1,131 @@
...
-
-    @Test
-    public void createAndDelete() throws InterruptedException {
-        String indexName = RandomStringBuilder.build("index-name", 8);
+    private static Pinecone controlPlaneClient = new Pinecone(System.getenv("PINECONE_API_KEY"));
+    private static final String indexName = RandomStringBuilder.build("create-index", 8);
+    @BeforeAll
+    public static void setUp() throws InterruptedException {
+        // Create the index
         ServerlessSpec serverlessSpec = new ServerlessSpec().cloud(ServerlessSpec.CloudEnum.AWS).region("us-west-2");
```
aulorbe, line 23: Might want to have a constant (or similar Java concept) for the `region` here in anticipation of us expanding to more regions?

src/integration/java/io/pinecone/integration/controlPlane/serverless/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,52 +1,131 @@
...
+                .metric(IndexMetric.COSINE)
+                .dimension(10)
+                .spec(createIndexRequestSpec);
+
+        try {
+            controlPlaneClient.createIndex(createIndexRequest);
+
+            fail("Expected to throw PineconeBadRequestException");
+        } catch (PineconeBadRequestException expected) {
+            assertTrue(expected.getLocalizedMessage().contains("must consist of lower case alphanumeric characters or '-'"));
```
aulorbe, line 71: Total nit, but technically shouldn't the "or" here be "and/or"? Currently to me it reads like the name can consist of lowercase alphanumeric chars OR just a hyphen, which isn't correct (I mean, technically you could do that I guess?)

src/integration/java/io/pinecone/integration/controlPlane/serverless/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -1,52 +1,131 @@
...
+                .metric(IndexMetric.COSINE)
+                .dimension(1)
+                .spec(createIndexRequestSpec);
+
+        try {
+            controlPlaneClient.createIndex(createIndexRequest);
+
+            fail("Expected to throw PineconeNotFoundException");
+        } catch (PineconeNotFoundException expected) {
+            assertTrue(expected.getLocalizedMessage().contains("not found"));
```
aulorbe, line 109: Not specific to this test, but maybe we should log a backlog issue to make the error msg for an invalid cloud more helpful than `"not found"`? Maybe something even just mentioning the resource, so like `"cloud provider not found"` and then an outlink to our documentation about current cloud providers? Not sure where to put that feedback...

aulorbe: (same w/region)

aulorbe (approved):
Just nits from me! Looks good!

rohanshah18 (approved):
Thanks for adding these integration tests.

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: Yup, since it returns an `Index` I just wanted to make that clearer (as opposed to `AsyncIndex`).

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: I can set up some `static final` variables, I think that's primarily how Java deals with the idea of constants.

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: For the validations on error messages, I was trying to keep them somewhat minimal per this feedback: https://github.com/pinecone-io/pinecone-java-client/pull/81#discussion_r1527630404

So I'm using `contains` to validate that specific substrings are in the error message, which is why some of these look a little odd at first glance. My thinking was a minimal string may help prevent things failing if the error messages change, but I feel like that's somewhat unlikely and it may be beneficial to include more of the error in the assert.

I'll touch things up based on your feedback, thanks!

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: See my response above about the error assertions, I'm just checking for substrings in the larger error message.

src/integration/java/io/pinecone/integration/controlPlane/serverless/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: Most of these errors are being returned from the server, so any tickets for adjusting the messages on these would need to be handled from that perspective. We're not generating the vast majority of these strings in the client.

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: The argument type for `metric` is an enum: `IndexMetric`. Users are not able to pass `String` values or other values outside of this enum. I can add a test / some asserts around passing different similarity metrics though, or leaving it empty.

aulorbe: ah TIL!

PR #83, opened 2024-03-20
Add data plane tests for serverless index

rohanshah18:
## Problem

Not all of the dataplane operations for serverless index are covered as a part of current test suite.

## Solution

Add integration and error tests for data plane operations for serverless index. As a part of this PR, I have added: 
1. Error tests for query and upsert data plane operations.
2. Integration tests for all data plane operations for both async and blocking stubs. 

## Type of Change

- [X] None of the above: 
Added more tests

## Test Plan
Ran integration tests locally and on CI


src/test/java/io/pinecone/BuildUpsertVectorWithUnsignedIndicesTest.java
```diff
@@ -58,14 +58,14 @@ void testNullId() {
         List<Long> sparseIndices = Arrays.asList(0L, 2L);
         List<Float> sparseValues = Arrays.asList(1.5f, 3.5f);
         Struct metadata = Struct.newBuilder().build();
-        StringBuilder exceptionMessage = new StringBuilder();
 
         try {
             buildUpsertVectorWithUnsignedIndices(id, values, sparseIndices, sparseValues, metadata);
+
+            fail("Expecting PineconeValidationException");
```
rohanshah18, line 65: Replaced the finally with fail()

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
```diff
@@ -0,0 +1,164 @@
...
+        PineconeConnection connectionMock = mock(PineconeConnection.class);
+
+        VectorServiceGrpc.VectorServiceBlockingStub stubMock = mock(VectorServiceGrpc.VectorServiceBlockingStub.class);
+        VectorServiceGrpc.VectorServiceFutureStub asyncStubMock = mock(VectorServiceGrpc.VectorServiceFutureStub.class);
+
+        when(connectionMock.getBlockingStub()).thenReturn(stubMock);
+        when(connectionMock.getFutureStub()).thenReturn(asyncStubMock);
+
+        index = new Index(connectionMock);
+        asyncIndex = new AsyncIndex(connectionMock);
```
austin-denoble, line 47: I'm confused why we're mocking here. Should we be testing the actual calls for integration testing purposes?

You should still be able to call without mocking and assert on the expected errors, and then we wouldn't need to disable the tests around server-side validations.

src/integration/java/io/pinecone/integration/dataPlane/UpsertErrorTest.java
```diff
@@ -0,0 +1,197 @@
...
+        PineconeConnection connectionMock = mock(PineconeConnection.class);
+
+        VectorServiceGrpc.VectorServiceBlockingStub stubMock = mock(VectorServiceGrpc.VectorServiceBlockingStub.class);
+        VectorServiceGrpc.VectorServiceFutureStub asyncStubMock = mock(VectorServiceGrpc.VectorServiceFutureStub.class);
+
+        when(connectionMock.getBlockingStub()).thenReturn(stubMock);
+        when(connectionMock.getFutureStub()).thenReturn(asyncStubMock);
+
+        index = new Index(connectionMock);
+        asyncIndex = new AsyncIndex(connectionMock);
```
austin-denoble, line 39: Again, I'm confused why we're mocking here rather than just making the actual requests and validating the expected errors. Is this to speed up the tests?

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Test
+    public void queryWithVectorAndIdSyncTest() {
+        List<Float> values = generateVectorValuesByDimension(3);
```
austin-denoble, line 52: You can probably use the `dimension` value where we're hard-coding this.

austin-denoble (approved):
Thanks for adding additional data plane tests, these look good to me. However, I'm confused about the mocking in integration tests.

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
rohanshah18: For error tests, we want to verify client-side validations only, so there's no need to hit the server. One option for server side error tests I thought was to move it to the query integration test but since that test is meant for successful querying cases only, it wouldn't be ideal to place it there. Also the disabled test is verifying server side error logic so its not something client tests should ideally worry about.

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
rohanshah18: Actually I should try to isolate each tests as much as possible, so let me remove the class variable.

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
austin-denoble: The client-side validations will prevent the requests from being sent to the server anyways. If the disabled tests aren't even meant to be run then I wouldn't include them in here if you're keeping the mocking paradigm, it's sort of confusing otherwise.

We assert and validate on server responses in integration tests on the other client, I don't think there's anything wrong in doing that. I did a bunch of that in create index error tests in the control plane: https://github.com/pinecone-io/pinecone-java-client/blob/7a26bde5d5c67ec64476b1e3b4a2fa6f0049febc/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java#L135

rohanshah18: > The client-side validations will prevent the requests from being sent to the server anyways.
I'm actually running a test that queries with a vector of dimension 1 for an index with dimension 3, so that's not something we check on client side. That's why the above statement is not true. We should hit the server if its not needed. And in my case, I'll move the server-side validations error test to the query integration test, so I can still mock stubs and run the client side validation tests.

rohanshah18: Turns out this is not even an issue for serverless, so we are allowed to query an index of dimension d1 with input vector of dimension d2. Will remove these tests and follow up with the team.

PR #84, opened 2024-03-21
Add `user-agent` to control plane operations, add `sourceTag`, re-enable `Pinecone` client unit tests

austin-denoble:
## Problem
We're currently not sending `user-agent` along with HTTP calls for control plane operations. We also want to rename `integrationId` to `sourceTag` to conform with the new attribution paradigm.

## Solution

While working on these changes I felt like it made sense to expand the number of constructors `Pinecone` supports to allow for easier passing of `apiKey`, `PineconeConfig`, `sourceTag`, and `OkHttpClient`. See the comments on the PR for details.

I also wanted to get the unit tests for the control plane client running again, so I refactored the old commented file.

- Add `user-agent` to HTTP control plane operations using `ApiClient.setUserAgent(PineconeConfig.getUserAgent())`. We already apply the `user-agent` to the gRPC `metadata`: https://github.com/pinecone-io/pinecone-java-client/blob/4581d5ef8c4db9709c7287f9cdf6379cb6d63873/src/main/java/io/pinecone/configs/PineconeConnection.java#L140.
- Rename `integrationId` -> `sourceTag`. Thread `integrationId` from `Pinecone` -> `PineconeConfig`. Apply normalization rules for the `sourceTag` slug:
  - Lower case
  - Condense multiple spaces to one space
  - Replace spaces with "_"
  - Filter to charset [a-z0-9_]
- Refactor `PineconeIndexOperationClientTest` to `PineconeTest`. Add new JSON fixtures under `/test/java/resources/` to support the new tests.

## Type of Change

## Test Plan
You should be able to set the `PINECONE_DEBUG` environment variable to see log output while running integration tests locally. This is the approach I used for validating `source_tag` was added to the user-agent, but it's a bit of setup to get it up and running locally.

I'm going to look at updating some of the integration tests to explicitly pass a test `sourceTag` for testing and we could validate those results in Datadog.


src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -21,19 +21,49 @@ public Pinecone(String apiKey) {
...
+        config = pineconeConfig;
+
         ApiClient apiClient = new ApiClient(okHttpClient);
-        apiClient.setApiKey(apiKey);
+        apiClient.setApiKey(config.getApiKey());
+        apiClient.setUserAgent(config.getUserAgent());
+
+        if (Boolean.parseBoolean(System.getenv("PINECONE_DEBUG"))) {
+            apiClient.setDebugging(true);
+        }
```
austin-denoble, line 58: We support this same environment variable on the other clients for logging out information about the calls being made. I enabled this here in order to validate the `user-agent` was being sent. I'm fine removing it if we want, but I don't see an issue with leaving it.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -52,12 +82,14 @@ public IndexModel describeIndex(String indexName) throws PineconeException {
         return indexModel;
     }
 
-    public void configureIndex(String indexName, ConfigureIndexRequest configureIndexRequest) throws PineconeException {
+    public IndexModel configureIndex(String indexName, ConfigureIndexRequest configureIndexRequest) throws PineconeException {
+        IndexModel indexModel = new IndexModel();
```
austin-denoble, line 86: `configureIndex` actually returns an index, updated to handle that appropriately.

src/test/java/io/pinecone/PineconeTest.java
```diff
@@ -0,0 +1,239 @@
+package io.pinecone;
```
austin-denoble, line 1: This is essentially a refactor of what was in `PineconeIndexOperationClientTest` which was completely commented out. I think these could be expanded and improved further down the line, for now I just restored the basic validation that was in place previously.

It's slightly different now since we rely on generated classes.

src/test/resources/indexListJsonString.json
```diff
@@ -0,0 +1,36 @@
+{
```
austin-denoble, line 1: Replaced the old index fixtures with these updated ones.

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -7,11 +7,16 @@ public class PineconeConfig {
 
     private String apiKey;
     private String host;
-    private String integrationId;
+    private String sourceTag;
```
austin-denoble, line 10: `integrationId` has been renamed to `sourceTag`

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -21,19 +21,49 @@ public Pinecone(String apiKey) {
...
+        this(new PineconeConfig(apiKey), sourceTag, new OkHttpClient());
+    }
+
     public Pinecone(String apiKey, OkHttpClient okHttpClient) {
-        if (apiKey == null || apiKey.isEmpty()) {
-            throw new PineconeValidationException("The API key is required and must not be empty or null");
-        }
-        config = new PineconeConfig(apiKey);
+        this(new PineconeConfig(apiKey), null, okHttpClient);
+    }
```
austin-denoble, line 30: I did a few things here that I think feel a bit better to reason about in terms of encapsulation, but let me know.

1. Added additional constructor overrides to support passing `PineconeConfig` as well as the generic type arguments. `PineconeConfig` is where `sourceTag` is stored, and is used for `PineconeConnection`. People will be able to import and use these classes directly, so it felt like it made sense to make it flexible.
2. Encapsulated the validation logic inside `PineconeConfig.validate()`. We construct the config then validate it before using it to configure the `ApiClient`.
3. Threading `sourceTag` through to `PineconeConfig` where it's set and validated when set.
4. Added a check for `PINECONE_DEBUG` in environment variables and enables logging for the `ApiClient` if so. This was helpful for validating `user-agent` changes.

ssmith-pc: I think with this number of constructor params/combinations the permutations of constructor overloads will be super confusing for users -- it's an anti-pattern known as "telescoping constructors".

For `sourceTag` in particular, it's an optional param that we'll tell partners to explicitly set, so we *could* do something like:
```java
Pinecone pc = new Pinecone("my-api-key");
pc.setSourceTag("my-tag");
```

A safer approach, especially if there are lots of such optional params/configurations would be to use a builder pattern. It would look something like this:

```java
public class Pinecone {

  public static class Builder {
     // required params
     private final String apiKey;

     // optional params
     private String sourceTag;

     public Builder(apiKey) {
       this.apiKey = apiKey;
     }

     public Builder withSourceTag(String sourceTag) {
         this.sourceTag = sourceTag;
         return this;
      }

     public Pinecone build() {
         config = .... // construct PineconeConfig from contents of Builder
         return new Pinecone(config);
     }
  }
}

// and to use it:
Pinecone pc = new Pinecone.Builder("my-api-key").withSourceTag("my-tag").build();
```

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: We shouldn't need to construct it ahead of time -- you can just do:

```java
IndexModel indexModel;
```

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -50,17 +56,37 @@ public interface CustomChannelBuilder {
...
-        String userAgentLanguage = "lang=java; pineconeClientVersion = v0.8.0";
-        if (this.getIntegrationId() == null) {
+        String userAgentLanguage = "lang=java; pineconeClientVersion=v0.8.0";
+        if (this.getSourceTag() == null) {
             return userAgentLanguage;
         } else {
-            return userAgentLanguage + "; usageContext=" + this.getIntegrationId();
+            return userAgentLanguage + "; source_tag=" + this.getSourceTag();
+        }
+    }
```
ssmith-pc, line 71: Seems like the logic could be simplified a bit, something like:

```
set user agent
if source tag is set
    append source tag to user agent
return user agent
```

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -50,17 +56,37 @@ public interface CustomChannelBuilder {
...
+    private String normalizeSourceTag(String input) {
+        if (input == null) {
+            return null;
         }
+
+        /*
+         * Normalize the source tag
+         * 1. Lowercase
+         * 2. Trim left/right whitespace
+         * 3. Condense multiple spaces to one
```
ssmith-pc, line 82: ...and replace space with underscore (so someone doesn't come along and "fix" it later

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -50,17 +56,37 @@ public interface CustomChannelBuilder {
...
+        String userAgentLanguage = "lang=java; pineconeClientVersion=v0.8.0";
+        if (this.getSourceTag() == null) {
             return userAgentLanguage;
         } else {
-            return userAgentLanguage + "; usageContext=" + this.getIntegrationId();
+            return userAgentLanguage + "; source_tag=" + this.getSourceTag();
+        }
+    }
+
+    private String normalizeSourceTag(String input) {
```
ssmith-pc, line 73: We should add some unit tests around this normalization

ssmith-pc (commented):
We should probably have a wider team review about the public interface we expose, to Builder or not to Builder, etc. 

Also, I didn't see any tests checking for sourceTag / user agent, unless I missed them?

src/main/java/io/pinecone/clients/Pinecone.java
aulorbe: TIL ^ !

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -50,17 +56,37 @@ public interface CustomChannelBuilder {
         ManagedChannel buildChannel();
     }
 
-    void validate() {
-        if (apiKey == null)
-            throw new PineconeValidationException("Invalid PineconeConfig: missing apiKey");
+    public void validate() {
+        if (apiKey == null || apiKey.isEmpty())
```
aulorbe, line 60: Oo adding this 2nd check to my example app's code, thnx :)

src/test/java/io/pinecone/PineconeTest.java
```diff
@@ -0,0 +1,238 @@
...
+        OkHttpClient mockClient = mock(OkHttpClient.class);
+        when(mockClient.newCall(any(Request.class))).thenReturn(mockCall);
+        Pinecone client = new Pinecone("testApiKey", mockClient);
+        IndexModel configuredIndex = client.configureIndex("testIndex", configureIndexRequest);
+
+        verify(mockClient, times(1)).newCall(any(Request.class));
+        verify(mockCall, times(1)).execute();
+        assertEquals(expectedConfiguredIndex, configuredIndex);
+    }
+}
```
aulorbe, line 238: Since this file is so long, would it be worth it to break it up into more readable files?

src/test/resources/indexListJsonString.json
aulorbe: Do we usually have binary files that act as fixtures for our tests? From what I remember from Python-world, I would make a `setUp` function that'd write whatever fixtures I'd need for my tests, but then I'd also write a `tearDown` function that'd delete the files after the tests ran, just to avoid clogging up the repo.

aulorbe (approved):
Just lurking mostly, lgtm!

src/main/java/io/pinecone/clients/Pinecone.java
austin-denoble: Thank you, this is good to know. I've heard about the builder pattern, but wasn't entirely clear on what that looks like or when it's employed. I think we should look at moving to this for some areas where we have a lot of variations of optional parameters and complex assortments of constructors. 👍

src/main/java/io/pinecone/configs/PineconeConfig.java
austin-denoble: Yeah absolutely, not sure why I didn't. Will add some in, thanks!

src/main/java/io/pinecone/clients/Pinecone.java
austin-denoble: Thank you! I think we should adjust all of these control plane functions given that rather than having all these empty objects that may possibly be returned.

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: We have flexibility with the naming as well, it doesn't *have* to be Builder. It can be based on the semantics we'd like to provide users with. 

So for example, here are some other options, but you and @rohanshah18 can probably come up with better ones:

```java
Pinecone pc = new Pinecone.Connection("my-api-key").withSourceTag("my-tag").connect();
Pinecone pc = new Pinecone.Client("my-api-key").withSourceTag("my-tag").build();

// we don't have to use "with" either, some APIs skip it
Pinecone pc = new Pinecone.Builder("my-api-key")
      .sourceTag("my-tag")
      .okHttpClient(okClient)
      .build();

// and there's another flavor where we use a static function as a constructor, which would internally use a private builder.
Pinecone pc = Pinecone.newConnection("my-api-key")
      .sourceTag("my-tag")
      .okHttpClient(okClient)
      .build();
```

src/main/java/io/pinecone/configs/PineconeConfig.java
ssmith-pc: Oh java 🤦‍♂️

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: We used to offer something like `withMember1()` in deprecated configs, so I have no issue using `with` keyword. I think its widely used and clear for java users on what the intention is.

src/test/resources/indexListJsonString.json
austin-denoble: I'm not sure what the standard is generally, but I don't necessarily see an issue with keeping the files themselves around, especially given they're in the `/test/resources/` folder which I think is somewhat standard in Java projects. For this PR though I was mainly following what was being done previously for representing some of these request responses.

In my opinion it's a bit easier to grep what's in the JSON payload with things represented in a specific file rather than done programmatically as part of the test but that's just imo. I'd definitely be open to changing up how things are done.

ssmith-pc: I think both approaches are fine, and in this case where it seems to static test configuration then keeping it around as a resource is totally fine. 

Sometimes it can be helpful/easier to generate test data and in those cases we definitely want to clean up afterwards. An example would be, let's say we need to have a JSON list of 10,000 fake users. It's like 5 lines of code to generate that vs. checking in an actual test file with 10,000 users, and managing that over time.

src/main/java/io/pinecone/configs/PineconeConfig.java
austin-denoble: Good catch. :D

src/test/java/io/pinecone/PineconeTest.java
austin-denoble: It is a bit long, I'm planning on adding more in as well so it probably does make sense to break it into chunks. Let me take a look.

src/test/resources/indexListJsonString.json
aulorbe: Cool, sounds good!

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -14,26 +14,57 @@
...
-    public Pinecone(String apiKey) {
-        this(apiKey, new OkHttpClient());
+    private Pinecone(PineconeConfig config) {
+        this.config = config;
     }
 
-    public Pinecone(String apiKey, OkHttpClient okHttpClient) {
-        if (apiKey == null || apiKey.isEmpty()) {
-            throw new PineconeValidationException("The API key is required and must not be empty or null");
+    public static class Client {
```
austin-denoble, line 24: Swapped to the builder pattern here, let me know what you think.

src/main/java/io/pinecone/configs/PineconeConfig.java
austin-denoble: Unit tests added in `PineconeConfigTest.java`.

src/test/java/io/pinecone/PineconeClientTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.AbstractMap;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+public class PineconeClientTest {
```
austin-denoble, line 19: This covers testing aspects of the `Client` builder.

src/test/java/io/pinecone/PineconeConfigTest.java
```diff
@@ -0,0 +1,66 @@
+package io.pinecone;
+
+import io.pinecone.configs.PineconeConfig;
+import io.pinecone.exceptions.PineconeConfigurationException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class PineconeConfigTest {
```
austin-denoble, line 10: New test file focusing on `PineconeConfig` which has the `user-agent` generation logic for now.

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.times;
+
+public class PineconeIndexOperationsTest {
```
austin-denoble, line 23: For now we can use this to test any logic related to control plane operations.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -14,26 +14,57 @@
 
 public class Pinecone {
 
-    private final ManageIndexesApi manageIndexesApi;
+    private static ManageIndexesApi manageIndexesApi;
```
ssmith-pc, line 17: Reason for changing this to static?

austin-denoble: It's instantiated in `Client.Build` and the compiler was complaining.

ssmith-pc: ah, I see. I'll make a code suggestion

ssmith-pc: I think we do not want to make it static or else the same index instance will be shared across all instances of Pinecone, which is probably not what we want

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -14,26 +14,57 @@
...
+            apiClient.setUserAgent(clientConfig.getUserAgent());
+
+            if (Boolean.parseBoolean(System.getenv("PINECONE_DEBUG"))) {
+                apiClient.setDebugging(true);
+            }
+
+            manageIndexesApi = new ManageIndexesApi();
+            manageIndexesApi.setApiClient(apiClient);
+
+            return new Pinecone(clientConfig);
```
ssmith-pc, line 62:
```suggestion
            ManagesIndexAPI manageIndexesApi = new ManageIndexesApi();
            manageIndexesApi.setApiClient(apiClient);

            return new Pinecone(clientConfig, manageIndexesApi);
```

src/test/java/io/pinecone/PineconeClientTest.java
ssmith-pc: Nice!

src/test/java/io/pinecone/PineconeConfigTest.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    public void testSourceTagIsNormalized() {
+        PineconeConfig config = new PineconeConfig("testApiKey");
+        config.setSourceTag("test source tag !! @@ ##");
+        assertEquals(config.getSourceTag(), "test_source_tag___");
+
+        config.setSourceTag("TEST SOURCE Tag     ----");
+        assertEquals(config.getSourceTag(), "test_source_tag_");
```
ssmith-pc, line 61: So I think if someone gives us a source tag of `test source tag !! @@ ##` we should turn that into: `test_source_tag`

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -0,0 +1,212 @@
...
+                .build());
+        OkHttpClient mockClient = mock(OkHttpClient.class);
+        when(mockClient.newCall(any(Request.class))).thenReturn(mockCall);
+
+        Pinecone client = new Pinecone.Client("testAPiKey").withOkHttpClient(mockClient).build();
+        client.deleteIndex("testIndex");
+
+        verify(mockClient, times(1)).newCall(any(Request.class));
+        verify(mockCall, times(1)).execute();
+    }
```
ssmith-pc, line 44: If I'm reading the validation correctly here, it's just checking that *some/any* request was called. Is there a way we can check whether the delete index request was made?

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -0,0 +1,212 @@
...
+                .body(ResponseBody.create(indexJsonStringPod, MediaType.parse("application/json")))
+                .build());
+
+        OkHttpClient mockClient = mock(OkHttpClient.class);
+        when(mockClient.newCall(any(Request.class))).thenReturn(mockCall);
+        Pinecone client = new Pinecone.Client("testAPiKey").withOkHttpClient(mockClient).build();
+        client.createIndex(createIndexRequest);
+
+        verify(mockClient, times(1)).newCall(any(Request.class));
+        verify(mockCall, times(1)).execute();
```
ssmith-pc, line 69: In general I like how clean each of these tests is, even with all the mocking that's going on. Nice work

src/test/java/io/pinecone/PineconeConfigTest.java
ssmith-pc: And I realized I needed to make an update to the Python impl of this. Here is what it looks like now. Note the inclusion of the `' '` space char to the charset regex, and then later the spaces will get replaced with underscore

```python
def _build_source_tag_field(source_tag):
    # normalize source tag
    # 1. Limit charset to [a-z0-9_ ]
    # 2. Lowercase
    # 3. Trim left/right whitespace
    # 4. Condense multiple spaces to one, and replace with underscore
    tag = re.sub(r'[^a-z0-9_ ]', '', source_tag)
    tag = tag.lower().strip()
    tag = "_".join(tag.split())
    return f"{SOURCE_TAG}={tag}"
```

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
rohanshah18: I think at the time of coding this, my logic was there should exactly be one request made and since its client.deleteIndex, we assumed its delete index. But that's a good point.

ssmith-pc: I think we can separate the potential test changes in a follow-up PR

src/test/java/io/pinecone/PineconeConfigTest.java
austin-denoble: Agreed, and thanks for the catch. I've updated the Java implementation to match.

src/main/java/io/pinecone/clients/Pinecone.java
austin-denoble: We definitely don't, thanks for the callout. I've updated to leverage the pattern you recommended of passing `ManageIndexesApi` to the constructor.

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
austin-denoble: Appreciated, but credit goes to @rohanshah18 I'm mostly cribbing from what was here originally: https://github.com/pinecone-io/pinecone-java-client/blob/main/src/test/java/io/pinecone/PineconeIndexOperationClientTest.java

src/test/java/io/pinecone/PineconeTest.java
austin-denoble: Things have been split into:

- `PineconeBuilderTest`
- `PineconeConfigTest`
- `PineconeIndexOperationsTest`

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
austin-denoble: Made some updates to this to validate a bit more granularly. We're now using `ArgumentCaptor` to capture information about the request and validate the HTTP method and URL. There's probably more we can do but this is at least specific about what it's validating for.

ssmith-pc (approved):

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -14,26 +14,58 @@
 
 public class Pinecone {
 
-    private final ManageIndexesApi manageIndexesApi;
+    private ManageIndexesApi manageIndexesApi;
```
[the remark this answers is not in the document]
austin-denoble: There was no specific reason, I accidentally removed it when removing `static` at one point in this PR. Thanks for the call out.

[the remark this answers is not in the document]
rohanshah18: Okay thanks!!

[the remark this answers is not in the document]
rohanshah18: I'm working on the Pinecone class for adding singleton cache, so I'll add it back.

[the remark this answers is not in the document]
austin-denoble: Sounds good, thank you.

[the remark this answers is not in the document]
ssmith-pc: Good catch

PR #87, opened 2024-03-27
Poll for Index ready during cleanup in `ConfigureIndexTest`

austin-denoble:
## Problem
We've started seeing consistent failures with the `ConfigureIndexTest`. 

The issue we're seeing is we're trying to delete the index before it has finished upgrading:
Screenshot 2024-03-27 at 11 45 30 AM

```
io.pinecone.exceptions.PineconeBadRequestException: {"error":{"code":"INVALID_ARGUMENT","message":"Bad request: Cannot delete an index that is upgrading"},"status":400}
	at io.pinecone.exceptions.HttpErrorMapper.mapHttpStatusError(HttpErrorMapper.java:9)
	at io.pinecone.clients.Pinecone.handleApiException(Pinecone.java:167)
	at io.pinecone.clients.Pinecone.deleteIndex(Pinecone.java:111)
	at io.pinecone.integration.controlPlane.pod.ConfigureIndexTest.cleanUp(ConfigureIndexTest.java:34)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:725)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
        ...
```
Initially, I tried implementing polling using `waitUntilIndexIsReady()` before each test and when cleaning up the index to make sure we're not trying to operate on an `Upgrading` index. I noticed that there are times when an index is marked as `ready:true` but `state=Initializing` which seems to cause problems.

## Solution
I've added a `waitUntilIndexUpgraded` function to `ConfigureIndexTest`. This ensures that we're waiting for the index to return to a `Ready` state before sending additional requests or deleting.

Screenshot 2024-03-27 at 12 03 34 PM

## Type of Change
- [X] Infrastructure change (CI configs, testing)

## Test Plan

### DIFF
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
@@ -96,6 +96,7 @@ public class ConfigureIndexTest {
             assertNotNull(podSpec);
             assertEquals(podSpec.getReplicas(), 1);
         });
+        waitUntilIndexUpgraded(indexName);
     }
 
     @Test
@@ -137,5 +138,23 @@ public class ConfigureIndexTest {
             assertNotNull(podSpec);
             assertEquals(podSpec.getPodType(), "p1.x2");
         });
+        waitUntilIndexUpgraded(indexName);
+    }
+
+    // After calling configureIndex the index needs to be upgraded which
+    // can take some time. Poll for a bit until we're ready to continue operating on the index.
+    private void waitUntilIndexUpgraded(String indexName) throws InterruptedException {
+        int timeToWaitMs = 30000;
+        IndexModel index = controlPlaneClient.describeIndex(indexName);
+
+        while (index.getStatus().getState() != IndexModelStatus.StateEnum.READY || timeToWaitMs > 0) {
+            Thread.sleep(2000);
+            timeToWaitMs -= 2000;
+            index = controlPlaneClient.describeIndex(indexName);
+        }
+
+        if (!index.getStatus().getReady()) {
+            fail("Index " + indexName + " did not finish upgrading after " + timeToWaitMs + "ms");
+        }
     }
 }
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
@@ -29,8 +29,7 @@ public class ConfigureIndexTest {
     }
 
     @AfterAll
-    public static void cleanUp() throws InterruptedException {
-        waitUntilIndexUpgraded(indexName);
+    public static void cleanUp() {
         controlPlaneClient.deleteIndex(indexName);
     }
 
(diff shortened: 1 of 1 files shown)


src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -138,4 +144,20 @@ public void sizeIncrease() throws InterruptedException {
...
+
+        while (index.getStatus().getState() != IndexModelStatus.StateEnum.READY && timeToWaitMs > 0) {
+            Thread.sleep(2000);
+            timeToWaitMs -= 2000;
+            System.out.println("waited 2000ms for index to upgrade, time left: " + timeToWaitMs);
+            System.out.println("System model state: " + index.getStatus());
+            index = controlPlaneClient.describeIndex(indexName);
+        }
+        if (!index.getStatus().getReady()) {
+            fail("Index " + indexName + " did not finish upgrading after " + timeToWaitMs + "ms");
```
jhamon, line 160: This is fail condition is fine, but I think the message is going to say 0ms in the message which isn't quite right.

jhamon (approved):

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
austin-denoble: Ah doh, you're right nice catch.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -138,4 +144,21 @@ public void sizeIncrease() throws InterruptedException {
...
+
+    private static void waitUntilIndexStateIsReady(String indexName) throws InterruptedException {
+        int timeToWaitMs = 30000;
+        int timeWaited = 0;
+        IndexModel index = controlPlaneClient.describeIndex(indexName);
+
+        while (index.getStatus().getState() != IndexModelStatus.StateEnum.READY && timeWaited <= timeToWaitMs) {
+            Thread.sleep(2000);
+            timeWaited += 2000;
+            System.out.println("waited 2000ms for index to upgrade, time left: " + timeToWaitMs);
```
aulorbe, line 156: My only addition to this PR would be to turn the print stmts into logging (info) statements so that this dat is useful to those in production who likely aren't looking at the console

aulorbe (approved):

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
austin-denoble: Right, agreed thank you! Updated.

PR #88, opened 2024-03-29
Update gRPC version to 1.60.2 to address vulnerability concerns and fix data plane integration tests

rohanshah18:
## Problem
1. The gRPC netty version 1.58.0 has vulnerability. 
2. I was still seeing some failing CI tests for data plane operations.

## Solution

Updated the following dependencies version from 1.58.0 to 1.60.2.
"io.grpc:grpc-protobuf"
"io.grpc:grpc-stub"
"io.grpc:grpc-netty"

Compatibility matrix for `grpc-netty` and `netty-tcnative-boringssl-static` can be found here: https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty

Also as a part of this PR, I have reduced the delay that is set after upserting vectors in the integration tests to 90 seconds instead of 120 seconds and for updateAndQueryPodTest, I have followed the same pattern as serverless indexes to upsert once and run all of the tests after the upsertion. 

## Type of Change

Dependencies were updated so it doesn't qualify for bug fix but instead security fix.

## Test Plan

Ran integration tests on local a good couple of times and on the CI. Fixing configure index test is in flight and will be a future PR from @austin-denoble.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ sourceSets {
     }
 }
 
-def grpcVersion = '1.58.0'
+def grpcVersion = '1.60.2'
 
 dependencies {
     api "io.grpc:grpc-protobuf:${grpcVersion}"
@@ -46,16 +46,16 @@ dependencies {
     implementation 'org.slf4j:slf4j-api:2.0.5'
     implementation 'com.google.api.grpc:proto-google-common-protos:2.14.3'
     implementation 'com.squareup.okhttp3:okhttp:4.12.0'
+    implementation 'com.squareup.okhttp3:logging-interceptor:4.12.0'
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.14.2'
     implementation 'com.fasterxml.jackson.core:jackson-core:2.14.2'
     implementation 'com.google.code.gson:gson:2.9.1'
-    implementation 'com.squareup.okhttp3:logging-interceptor:4.10.0'
     implementation 'io.gsonfire:gson-fire:1.8.5'
     implementation 'org.openapitools:jackson-databind-nullable:0.2.6'
     implementation 'com.google.protobuf:protobuf-java:3.25.2'
     compileOnly "org.apache.tomcat:annotations-api:6.0.53" // necessary for Java 9+
 
-    testImplementation("com.squareup.okhttp3:mockwebserver:4.10.0")
+    testImplementation("com.squareup.okhttp3:mockwebserver:4.12.0")
     testImplementation "io.grpc:grpc-testing:${grpcVersion}"
     testImplementation "org.hamcrest:hamcrest:2.2"
     testImplementation 'org.mockito:mockito-inline:4.8.0'
--- a/src/integration/java/io/pinecone/helpers/IndexManager.java
+++ b/src/integration/java/io/pinecone/helpers/IndexManager.java
@@ -35,9 +35,10 @@ public class IndexManager {
         }
 
         for (IndexModel indexModel : indexModels) {
-            if (indexModel.getDimension() == dimension
-                && (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null)
-                || (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS) && indexModel.getSpec().getServerless() != null)
+            // Sparse-dense is only supported with DOTPRODUCT
+            if (indexModel.getDimension() == dimension && indexModel.getMetric() == IndexMetric.DOTPRODUCT
+                && ((indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null)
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -46,16 +46,16 @@ dependencies {
     implementation 'org.slf4j:slf4j-api:2.0.5'
     implementation 'com.google.api.grpc:proto-google-common-protos:2.14.3'
     implementation 'com.squareup.okhttp3:okhttp:4.12.0'
+    implementation 'com.squareup.okhttp3:logging-interceptor:4.12.0'
```
rohanshah18, line 49: I should have updated this to 4.12.0 when updating `com.squareup.okhttp3:okhttp` to 4.12.0 in the previous PR

build.gradle
```diff
@@ -46,16 +46,16 @@ dependencies {
...
     implementation 'com.fasterxml.jackson.core:jackson-core:2.14.2'
     implementation 'com.google.code.gson:gson:2.9.1'
-    implementation 'com.squareup.okhttp3:logging-interceptor:4.10.0'
     implementation 'io.gsonfire:gson-fire:1.8.5'
     implementation 'org.openapitools:jackson-databind-nullable:0.2.6'
     implementation 'com.google.protobuf:protobuf-java:3.25.2'
     compileOnly "org.apache.tomcat:annotations-api:6.0.53" // necessary for Java 9+
 
-    testImplementation("com.squareup.okhttp3:mockwebserver:4.10.0")
+    testImplementation("com.squareup.okhttp3:mockwebserver:4.12.0")
```
rohanshah18, line 58: I should have updated this to 4.12.0 when updating `com.squareup.okhttp3:okhttp` to 4.12.0 in the previous PR

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -35,9 +35,10 @@ public static String findIndexWithDimensionAndType(Pinecone pinecone, int dimens
         }
 
         for (IndexModel indexModel : indexModels) {
-            if (indexModel.getDimension() == dimension
-                && (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null)
-                || (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS) && indexModel.getSpec().getServerless() != null)
+            // Sparse-dense is only supported with DOTPRODUCT
```
rohanshah18, line 38: Even though we create new indexes of type DOT PRODUCT in this class when an index doesn't exist, the chances are there can be an index in console that was manually created or via other means but has a metric COSINE or EUCLIDEAN. This can be an issue for sparse-dense vectors tests since they are only supported with DOT PRODUCT metric

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -63,7 +64,7 @@ public static String createNewIndex(Pinecone pinecone, int dimension, String ind
         CreateIndexRequest createIndexRequest = new CreateIndexRequest()
                 .name(indexName)
                 .dimension(dimension)
-                .metric(IndexMetric.DOTPRODUCT)
+                .metric(IndexMetric.DOTPRODUCT) // Sparse-dense is only supported with DOTPRODUCT
```
rohanshah18, line 67: Added comment for future reference on why we create index with DOTPRODUCT metric.

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: Is `findIndexWithDimensionAndType` only used for sparse-dense index tests? I'm not really understanding why we needed to add `indexModel.getMetric() == IndexMetric.DOTPRODUCT` to this.

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeletePodTest.java
```diff
@@ -23,56 +23,59 @@
...
         });
 
         // Delete 1 vector
         List<String> idsToDelete = new ArrayList<>(3);
-        idsToDelete.add(upsertIds.get(0));
-        indexClient.delete(idsToDelete, false, namespace, nullFilterStruct);
-        vectorCount -= idsToDelete.size();
-        int testSingleDeletedVectorCount = vectorCount;
+        String vectorIdToDelete = upsertIds.get(0);
+        idsToDelete.add(vectorIdToDelete);
```
austin-denoble, line 71: Why the change pulling out `vectorIdToDelete`?

austin-denoble (approved):
Thanks for upgrading these dependency versions and reducing the wait time in these tests. 👍

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeletePodTest.java
rohanshah18: We are just assigning upsertIds.get(0) to vectorIdToDelete which is still added to idsToDelete.

src/integration/java/io/pinecone/helpers/IndexManager.java
rohanshah18: Is Cosine or Euclidean metric stricly needed for any other tests?

PR #89, opened 2024-03-29
Add `TestIndexResourcesManager` and `CleanupAllTestResourcesListener`

austin-denoble:
## Problem
Our current integration test setup is a bit difficult to reason about and debug. When adding new integration tests, it can be hard to understand the mechanics of how the current `IndexManager` creates and manages Index resources. We also end up with a lot of lingering index resources as indexes created via `createIndexIfNotExistsDataPlane` aren't generally cleaned up as a part of the tests they're used in.

Ultimately, I think we should move away from `createIndexIfNotExistsDataPlane` and `findIndexWithDimensionAndType` entirely and implement a way for indexes to be explicitly shared across tests by name.

## Solution
- Add new `TestIndexResourcesManager` singleton to handle creating, sharing, and cleaning up index resources across an integration test run.
- Add new `CleanupAllTestResourcesListener` which implements Junit's `TestExecutionListener` allowing us to run the singleton cleanup at the end of the entire test run.
- Update `/integration/controlPlane/*` tests to use the new pattern.

My goal here was to have a centralized `TestIndexResourcesManager` class that is used across all our integration tests. This class should handle global setup and teardown before and after _all_ test classes run. Essentially, I wanted this class to act as a generator and manager for external resources.

I wanted to also have the `TestIndexResourcesManager` only create the required resources as needed. Instead of trying to create and seed multiple indexes before any tests run, I'm trying to do this only at the time a test that's included in the run requests the resource.

The strategy I've used here is to leverage the Junit Platform Launcher `TestExecutionListener` interface: https://junit.org/junit5/docs/5.0.3/api/org/junit/platform/launcher/TestExecutionListener.html

We can create a new class `CleanupAllTestResourcesListener` that implements `TestExceutionListener` and overrides `testPlanExecutionFinished`. We then need to register this listener which I've done by adding the `/META-INF/services/org.junit.platform.launcher.TestExecutionListener` file to the `integration` module. Check the code comments for a bit more info about where I figured out how to do this.

In this PR I've only updated the tests under `/integration/controlPlane` to leverage this new pattern, and it would be very similar for updating the data plane operations. Basically, we would go through and replace usages of `findIndexWithDimensionAndType`, `createNewIndex`, and `createIndexIfNotExistsDataPlane` with calls to `getPodIndex()` from the singleton.

I tried a few different methods of doing "global setup and teardown" in Junit and found it quite difficult. It was a lot easier to trigger code before and after each test class rather than the entirety of the run.

Other approaches tried:
- I tried implementing a Junit extension and using the `@ExtendWith()` annotation across all tests: https://junit.org/junit5/docs/current/user-guide/#extensions. This worked for running code before and after each class but not the suite as a whole.
- I tried having a base class `BaseTest` that was extended by each test class, and had its own `@BeforeAll` and `@AfterAll`. Again, this triggers every test class and not at the test run level.
- Using "test suites", or organizing our tests at a higher level than discrete classes. This ultimately felt very cumbersome, and I couldn't find straightforward ways of running a suite directly with gradle, or handling running tests individually outside the suite context without losing the setup and teardown.

 
## Type of Change

## Test Plan
Run integration tests locally and in CI. Make sure control plane tests pass as expected with this new architecture.


.github/workflows/pr.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       run: gradle clean build
 
     - name: Run integration tests
-      run: gradle integrationTest
+      run: gradle integrationTest --info
```
austin-denoble, line 82: Ignore this, will remove before merging just wanted to be able to troubleshoot issues in CI more easily.

src/main/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
```diff
@@ -0,0 +1 @@
+io.pinecone.AfterAllTestsListener
```
austin-denoble, line 1: Apparently, this is one way of registering a `TestExecutionListener` with Junit. 

Tbh I went through a lot of trial and error, but I found guidance for this in a few places: https://stackoverflow.com/a/63774284

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -35,10 +35,10 @@ public static String findIndexWithDimensionAndType(Pinecone pinecone, int dimens
         }
 
         for (IndexModel indexModel : indexModels) {
-            // Sparse-dense is only supported with DOTPRODUCT
-            if (indexModel.getDimension() == dimension && indexModel.getMetric() == IndexMetric.DOTPRODUCT
-                && ((indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null)
-                || (indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS) && indexModel.getSpec().getServerless() != null))
+            if (indexModel.getDimension() == dimension
```
austin-denoble, line 38: I'm pretty sure this code was still broken. 

The second clause checking `indexType` and `indexModel.getSpec()` was doing `&& (type pod?) || (type serverless?)` rather than `&& ((type pod?) || (type serverless?))`

src/integration/java/io/pinecone/AfterAllTestsListener.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @Override
+    public void testPlanExecutionStarted(TestPlan testPlan) {
+        System.out.println("Integration tests starting. Running setup code...");
+
+    }
+    @Override
+    public void testPlanExecutionFinished(TestPlan testPlan) {
+        System.out.println("Integration tests completed. Running cleanup code...");
+        IndexManagerSingleton.getInstance().cleanupResources();
```
austin-denoble, line 17: This is the key detail involving the `TestExecutionListener` for now. Basically, we want to call the `IndexManagerSingleton` cleanup method once at the very end after all tests classes have run, which is what this accomplishes.

src/integration/java/io/pinecone/helpers/IndexManagerSingleton.java
```diff
@@ -0,0 +1,208 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static io.pinecone.helpers.BuildUpsertRequest.buildRequiredUpsertRequestByDimension;
+import static io.pinecone.helpers.IndexManager.*;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class IndexManagerSingleton {
```
austin-denoble, line 19: This is the most crucial part of this work. The goal here is to have a centralized singleton that manages all the index resources across integration tests explicitly, and is able to tear things down at the end.

build.gradle
```diff
@@ -62,6 +62,7 @@ dependencies {
     testImplementation 'org.slf4j:slf4j-simple:2.0.5'
     testImplementation 'com.fasterxml.jackson.core:jackson-databind:2.14.2'
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.0'
+    testImplementation 'org.junit.platform:junit-platform-launcher:1.8.0'
```
austin-denoble, line 65: This dependency is needed for using the `TestExecutionListener` and registering it.

src/test/java/io/pinecone/AfterAllTestsListener.java
```diff
@@ -0,0 +1,17 @@
+package io.pinecone;
```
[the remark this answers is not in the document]
ssmith-pc: That does feel weird. Do we have to call it the same thing, and implement the same methods? I'm wondering if we could do something like this:

```java
public class IgnoreMeListener implements TestExecutionListener {
   // leave a comment explaining that this has to be there for whatever reason, but see integration tests for more info
}
```

src/integration/java/io/pinecone/AfterAllTestsListener.java
```diff
@@ -0,0 +1,22 @@
...
+
+import io.pinecone.helpers.IndexManagerSingleton;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestPlan;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AfterAllTestsListener implements TestExecutionListener {
+    private static final Logger logger = LoggerFactory.getLogger(AfterAllTestsListener.class);
+
```
ssmith-pc, line 11: Should we call this class the `CleanupAllTestResourcesListener`?

src/integration/java/io/pinecone/AfterAllTestsListener.java
```diff
@@ -0,0 +1,22 @@
...
+    @Override
+    public void testPlanExecutionStarted(TestPlan testPlan) {
+        logger.info("Integration tests starting. Running setup...");
+
```
ssmith-pc, line 15: There is nothing to do on startup? Is that because resources are created on-demand throughout the test run?

src/integration/java/io/pinecone/helpers/IndexManager.java
ssmith-pc: I think the precedence order says that `&&` has higher precedence than `||`, but I find the lack of parens combined with the numerous conditions hard to parse in my head. Whatever the logic needs to be, I'd suggest we keep some kind of explicit parens grouping, or else introduce some intermediate booleans, e.g.:

```java
boolean typePod = indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null;
boolean typeServerless = indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS) && indexModel.getSpec().getServerless() != null;

if (typePod || typeServerless) {
.....
```

src/integration/java/io/pinecone/helpers/IndexManagerSingleton.java
ssmith-pc: I think we should consider renaming this class to make it clear that it's a test utility, not part of the client itself.

Naming alternatives?
* `TestResourceManager`
* `TestIndexResourcesManager`
* `IndexTestingResourcesManager`

src/integration/java/io/pinecone/helpers/IndexManagerSingleton.java
```diff
@@ -0,0 +1,211 @@
...
+            ? 4
+            : Integer.parseInt(System.getenv("DIMENSION"));
+    private final String environment = System.getenv("PINECONE_ENVIRONMENT") == null
+            ? "us-east4-gcp"
+            : System.getenv("PINECONE_ENVIRONMENT");
+    private static final IndexMetric metric = System.getenv("METRIC") == null
+            ? IndexMetric.COSINE
+            : IndexMetric.valueOf(System.getenv("METRIC"));
+    private static final PineconeConfig pineconeConfig = new PineconeConfig(apiKey);
+    private static Pinecone pineconeClient;
```
ssmith-pc, line 33: I don't think `pineconeConfig` needs to be a member var

I don't think `pineconeClient` needs to be static.

src/integration/java/io/pinecone/helpers/IndexManagerSingleton.java
```diff
@@ -0,0 +1,211 @@
...
+            serverlessIndexName = createServerlessIndex();
+        }
+        return serverlessIndexName;
+    }
+
+    public String getCollectionName() throws InterruptedException {
+        if (collectionName == null) {
+            collectionName = createCollection();
+        }
+        return collectionName;
```
ssmith-pc, line 71: I would consider moving all of these existence conditionals to within the `createCollection` method, or `createServerlessIndex`, etc.

For example, if it was instead called `createOrGetServerlessIndex` you could use it like this:

```java
    public String getServerlessIndexName() throws InterruptedException, PineconeException {
        return createOrGetServerlessIndex();
    }

    private String createOrGetServerlessIndex() throws InterruptedException, PineconeException {
       // -----> null checks encapcsulated here
       if (serverlessIndexName == null) {
            return serverlessIndexName;
        }

        String indexName = RandomStringBuilder.build("serverless-index", 8);

        // Create index
        ServerlessSpec serverlessSpec = new ServerlessSpec().cloud(ServerlessSpec.CloudEnum.AWS).region("us-west-2");
        CreateIndexRequestSpec spec = new CreateIndexRequestSpec().serverless(serverlessSpec);
        CreateIndexRequest createIndexRequest = new CreateIndexRequest().name(indexName).dimension(dimension).metric(metric).spec(spec);
        serverlessIndexModel = pineconeClient.createIndex(createIndexRequest);

        // -------> have to save it here now
        this.serverlessIndexName = indexName;
        return indexName;
    }
```

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -13,37 +13,28 @@
...
-    }
-
-    @AfterAll
-    public static void cleanUp() {
-        // Delete the index
-        controlPlaneClient.deleteIndex(indexName);
+        indexName = indexManager.getPodIndexName();
+        indexDimension = indexManager.getDimension();
+        IndexModel podIndex = indexManager.getPodIndexModel();
+        indexPodType = podIndex.getSpec().getPod().getPodType();
```
ssmith-pc, line 29: LOVE how much code is being removed from individual tests

src/main/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
ssmith-pc: Odd, I wouldn't have guessed this. 

Can you move the file to META-INF in test folder? `test/main/resources/META-INF/etc...`

ssmith-pc (approved):
Really nice! Makes individual tests look a lot cleaner and simplifies obtaining test resources, and cleaning when you're done. 

I left a couple of comments for things to consider, but nothing really blocking

src/integration/java/io/pinecone/AfterAllTestsListener.java
austin-denoble: Yeah, since it's a singleton and we're lazily creating index resources as needed, we only need to call for cleanup at the end of our test run at the moment.

src/integration/java/io/pinecone/helpers/IndexManager.java
rohanshah18: Per Silas's suggestion, this is the simplified/easily readable code:
```
boolean typePod = indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_POD) && indexModel.getSpec().getPod() != null;
boolean typeServerless = indexType.equalsIgnoreCase(IndexModelSpec.SERIALIZED_NAME_SERVERLESS) && indexModel.getSpec().getServerless() != null;

        && indexModel.getMetric() == IndexMetric.DOTPRODUCT
        && (typePod || typeServerless))
```

If the metric is not DOTPRODUCT and if an index exists matching the dimension for serverless or pod, than the dataplane tests for upsert and query with sparse values will fail.

src/main/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
austin-denoble: Good call! I had tried adding a `/META-INFO/services/` folder to the `test` and `integration` modules, but I didn't delete things out of `main` which is why it wasn't registering properly.

I've managed to get things working properly by adding `/resources/META-INFO/services/` to the `integration` folder and removing it from elsewhere. 🎉

src/integration/java/io/pinecone/AfterAllTestsListener.java
austin-denoble: Yes, I think we should. Updated.

aulorbe:
> implement a way for indexes to be explicitly shared across tests by name.
Yes!

.github/workflows/pr.yml
aulorbe: Don't forget! :)

src/integration/java/io/pinecone/helpers/IndexManager.java
aulorbe: Maybe we should comment the code in this case just to make it super clear?

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -80,7 +81,7 @@ public static IndexModel waitUntilIndexIsReady(Pinecone pinecone, String indexNa
         int waitedTimeMs = 0;
         int intervalMs = 2000;
 
-        while (!index.getStatus().getReady()) {
+        while (index.getStatus().getState() != IndexModelStatus.StateEnum.READY) {
```
aulorbe, line 84: Not impt, but I was just reading in Effective Java that it's better to have binary `Enum` values instead of booleans, so nice job w/the `StateEnum` guy!

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -0,0 +1,200 @@
...
+
+import static io.pinecone.helpers.BuildUpsertRequest.buildRequiredUpsertRequestByDimension;
+import static io.pinecone.helpers.IndexManager.*;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class TestIndexResourcesManager {
+    private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
+    private static TestIndexResourcesManager instance;
+    private static final String apiKey = System.getenv("PINECONE_API_KEY");
+    private final int dimension = System.getenv("DIMENSION") == null
```
aulorbe, line 22: These env vars are all stored in Github Secrets (or whatever it's called) when these tests are run online, right?

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -0,0 +1,200 @@
...
+            ? 4
+            : Integer.parseInt(System.getenv("DIMENSION"));
+    private final String environment = System.getenv("PINECONE_ENVIRONMENT") == null
+            ? "us-east4-gcp"
```
aulorbe, line 26: Nice, I like that we now have a default. Much easier to run locally.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -0,0 +1,200 @@
...
+            ? IndexMetric.COSINE
+            : IndexMetric.valueOf(System.getenv("METRIC"));
+    private Pinecone pineconeClient;
+    private String podIndexName;
+    private IndexModel podIndexModel;
+    private String serverlessIndexName;
+    private IndexModel serverlessIndexModel;
+    private String collectionName;
+    private CollectionModel collectionModel;
+    private final List<String> podIndexVectorIds = Arrays.asList("v1", "v2", "v3");
```
aulorbe, line 38: Total nit, which I know I'll likely get pushback on because I think `v1` etc is how we represent vector IDs in other parts of the system and documentation, but I find it more intuitive to read if IDs are like `id1`. IMO `v1` seems more like an object, like the actual "vector 1".

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
aulorbe: Yeah this is so sweet

aulorbe (approved):
Just nits. Looks absolutely fantastic to me. I'm so excited about this change. Great job!!

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: Thanks, but it's part of the generated OpenAPI code. 🫣

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
austin-denoble: Yeah, I was trying to provide some flexibility in how we set them. So we can set them in CI and drive different variations of testing. Jen did a really good job of writing the integration tests in this way in the Python client, and I wanted to give us the option here.

We mostly rely on the defaults for now.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
austin-denoble: I think it's a bit arbitrary what the ids are, I can change them up. 👍

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: Awesome!

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -16,71 +16,35 @@
...
-    private static final String collectionName = RandomStringBuilder.build("collection-test", 8);
-    private static final ArrayList<String> indexesToCleanUp = new ArrayList<>();
-    private static final IndexMetric indexMetric = IndexMetric.COSINE;
-    private static final List<String> upsertIds = Arrays.asList("v1", "v2", "v3");
-    private static final String namespace = RandomStringBuilder.build("ns", 8);
-    private static final String apiKey = System.getenv("PINECONE_API_KEY");
-    private static final String environment = System.getenv("PINECONE_ENVIRONMENT");
-    private static final int dimension = 4;
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
+    private static Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
```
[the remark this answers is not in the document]
austin-denoble: We can just mark the `pineconeClient` itself as final and that should accomplish the same thing. I don't think we're doing that in any of the other test files where we setup like this, we can update it as we go.

src/integration/java/io/pinecone/helpers/IndexManager.java
austin-denoble: The goal is to lose this function entirely after converting the data plane tests to use the new `TestIndexResourcesManager`, so I think it's fine to leave this as is for now. The suggested refactoring helps a lot for readability too.

PR #91, opened 2024-04-01
Handle nulls for some creation and configure methods and provide alternative method for creating collections

aulorbe:
## Problem

Nulls:
We are not currently handling nulls gracefully when they are passed to the following methods in the Pinecone class: 
- CreateIndex
- ConfigureIndex
- CreateCollection

This PR captures them and raises a `PineconeValidationException` when they're encountered.

`Request` objects:
- It's currently cumbersome for users to create `Request` objects each time they have to call the above methods. 

This PR overrides the `createCollection` method with more intuitive params for the user. This abstracts away the `createCollectionRequest` object w/native Java types, so that it matches the strategy we have for similar patterns in the [dataplane](https://github.com/pinecone-io/pinecone-java-client/blob/main/src/main/java/io/pinecone/clients/Index.java#L37-L68). Now, a user can simply pass 2 `String`s to in order to create a collection (the `collectionName` and the `sourceIndex`). Within this method, the `createCollectionRequest` object is built. 

If this PR is approved, I will follow up with a PR to do the same for `CreateIndex` and `ConfigureIndex`. These need a bit more thought, so I'll make them into separate PRs.

## Type of Change

## Test Plan

Unit tests and integration tests included.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206975505478879

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -79,7 +79,7 @@ jobs:
       run: gradle clean build
 
     - name: Run integration tests
-      run: gradle integrationTest
+      run: gradle integrationTest --info
       env:
         PINECONE_API_KEY: ${{ secrets.PINECONE_API_KEY }}
         PINECONE_ENVIRONMENT: ${{ secrets.PINECONE_ENVIRONMENT }}
--- a/src/integration/java/io/pinecone/helpers/IndexManager.java
+++ b/src/integration/java/io/pinecone/helpers/IndexManager.java
@@ -126,8 +126,7 @@ public class IndexManager {
     }
 
     public static CollectionModel createCollection(Pinecone pinecone, String collectionName, String indexName, boolean waitUntilReady) throws InterruptedException {
-        CreateCollectionRequest createCollectionRequest = new CreateCollectionRequest().name(collectionName).source(indexName);
-        CollectionModel collection = pinecone.createCollection(createCollectionRequest);
+        CollectionModel collection = pinecone.createCollection(collectionName, indexName);
 
         assertEquals(collection.getStatus(), CollectionModel.StatusEnum.INITIALIZING);
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -15,6 +15,7 @@ import org.openapitools.client.model.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.sound.midi.SysexMessage;
 import java.util.*;
 
 import static io.pinecone.helpers.BuildUpsertRequest.*;
@@ -38,21 +39,33 @@ public class CollectionTest {
 
     @BeforeAll
     public static void setUp() throws InterruptedException {
+        System.out.println("In setUp method");
         pineconeClient = new Pinecone.Builder(apiKey).build();
 
         // Create and upsert to index
         CreateIndexRequestSpecPod podSpec =
                 new CreateIndexRequestSpecPod().pods(1).podType("p1.x1").replicas(1).environment(environment);
         CreateIndexRequestSpec spec = new CreateIndexRequestSpec().pod(podSpec);
+        System.out.println("Starting to create index " + indexName);
         Index indexClient = createNewIndexAndConnectSync(pineconeClient, indexName, dimension,
                 indexMetric, spec);
+        System.out.println("Successfully created index " + indexName);
+
+        System.out.println("Adding " + indexName + " to indexesToCleanUp list");
         indexesToCleanUp.add(indexName);
+        System.out.println("Added " + indexName + " to indexesToCleanUp list");
 
         // Sometimes we see grpc failures when upserting so quickly after creating, so retry if so
+        System.out.println("Starting to upsert vectors into " + indexName + " index");
         assertWithRetry(() -> indexClient.upsert(buildRequiredUpsertRequestByDimension(upsertIds, dimension), namespace), 3);
+        System.out.println("Successfully upserted vectors into " + indexName + " index");
+
 
         // Create collection from index
+        System.out.println("Starting to create collection " + collectionName + " from index " + indexName);
         collection = createCollection(pineconeClient, collectionName, indexName, true);
+        System.out.println("Successfully created collection " + collectionName + " from index " + indexName);
+
         assertEquals(collection.getName(), collectionName);
         assertEquals(collection.getEnvironment(), environment);
         assertEquals(collection.getStatus(), CollectionModel.StatusEnum.READY);
@@ -60,11 +73,16 @@ public class CollectionTest {
 
     @AfterAll
     public static void cleanUp() throws InterruptedException {
+        System.out.println("In cleanUp method");
+
         // wait for things to settle before cleanup...
         Thread.sleep(2500);
 
         // Verify we can delete the collection
+        System.out.println("Starting to delete collection " + collectionName);
         pineconeClient.deleteCollection(collectionName);
+        System.out.println("Successfully deleted collection " + collectionName);
+
         Thread.sleep(2500);
 
         List<CollectionModel> collectionList = pineconeClient.listCollections().getCollections();
@@ -84,7 +102,9 @@ public class CollectionTest {
 
         // Clean up indexes
         for (String index : indexesToCleanUp) {
-            pineconeClient.deleteIndex(index);
+            System.out.println("Starting to delete index " + index + " in indexesToCleanUp list");
+            pineconeClient.deleteIndex(index); // errors out here
+            System.out.println("Successfully deleted index " + index + " in indexesToCleanUp list"); // never gets here
         }
     }
 
@@ -179,10 +199,6 @@ public class CollectionTest {
         assertEquals(newIndex.getName(), newIndexName);
         assertEquals(newIndex.getMetric(), targetMetric);
     }
-
-    // TODO:
-    // add positive test case for create collection from real indexName?
-
 
     @Test
     public void testCreateCollectionFromFromNullOrEmptyStringSourceIndex() {
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -5,7 +5,8 @@ import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeException;
 import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.helpers.RandomStringBuilder;
-import io.pinecone.proto.*;
+import io.pinecone.proto.DescribeIndexStatsResponse;
+import io.pinecone.proto.FetchResponse;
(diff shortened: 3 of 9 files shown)


src/test/resources/collectionCreation.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name":"test-collection",
```
aulorbe, line 2: I think maybe I don't need this bc I pass a collection name (`testCollection`) [here](https://github.com/pinecone-io/pinecone-java-client/pull/91/files#diff-a3ec968386997c2c981ec71a27c2f0cf920cdf9263e08eaa7485a8b6bdbce4c0R270)?

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -238,5 +243,59 @@ public void testConfigureIndex() throws IOException {
...
+    }
+
+    @Test
+    public void testCreateCollection() throws IOException {
+        String filePath = "src/test/resources/collectionCreation.json";
+        String JsonStringCollection = new String(Files.readAllBytes(Paths.get(filePath)));
+        CollectionModel expectedCollection = gson.fromJson(JsonStringCollection, CollectionModel.class);
+
+        CreateCollectionRequest createCollectionRequest = new CreateCollectionRequest();
+        createCollectionRequest.setName("testCollection");
```
austin-denoble, line 270: Rather than setting the name in the test itself you could rely on what it is in the JSON fixture and keep things a bit more coupled. Currently the name `testCollection` and `test-collection` differ, which doesn't matter for this specific test, but may be confusing when reading the test.

Something like this:
```suggestion
        createCollectionRequest.setName(expectedCollection.getName());
```

src/test/resources/collectionCreation.json
austin-denoble: Responded above, I think it's better to keep the name here and then use that in the test so we don't have two different representations of `name` for this fixture.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -64,6 +65,9 @@ public Pinecone build() {
     }
 
     public IndexModel createIndex(CreateIndexRequest createIndexRequest) throws PineconeException {
+        if (createIndexRequest == null) {
+            throw new PineconeException("CreateIndexRequest object cannot be null");
```
austin-denoble, line 69: We can swap to `PineconeValidationException` for these errors I think, as it's a bit more specific. It's also what we use when throwing out of data plane calls that are missing required fields in `IndexInterface`: https://github.com/pinecone-io/pinecone-java-client/blob/640424f61c2e75067bf2ddcb478481144d4e8da8/src/main/java/io/pinecone/commons/IndexInterface.java#L59

austin-denoble (commented):
Overall good first steps here, and thanks for adding unit tests! Let's discuss today around @rohanshah18's feedback on validating the required parameters via Java native types. 👍

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
aulorbe: Weirdly it looks like I don't even need that at all in the test...

src/main/java/io/pinecone/clients/Pinecone.java
aulorbe: It's weird... I have changed it now to `PineconeValidationException` and changed the expected `throw` in the tests, but the tests don't care if I have `PineconeValidationException` as the expected `throw` or `PineconeException`... (but it does care if I mess with the expected message, which is good)

austin-denoble: I think it's because it extends it: https://github.com/pinecone-io/pinecone-java-client/blob/640424f61c2e75067bf2ddcb478481144d4e8da8/src/main/java/io/pinecone/exceptions/PineconeValidationException.java#L3

aulorbe: Yaya deleted my comment, thank yewwww!

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -112,6 +123,30 @@ public void deleteIndex(String indexName) throws PineconeException {
...
+        return collection;
+    }
+
+    public CollectionModel createCollection(String collectionName, String sourceIndex) throws PineconeValidationException {
+        if (collectionName == null || collectionName.isEmpty()) {
+            throw new PineconeValidationException("collectionName cannot be null or empty");
+        }
+        if (sourceIndex == null || sourceIndex.isEmpty()) {
+            throw new PineconeValidationException("sourceIndex cannot be null or empty");
+        }
```
austin-denoble, line 144: Ah I'm seeing why exposing the native type functions along with the `CreateCollectionRequest` can be weird.

If we want to expose different arrangements of arguments for these functions, we should try and keep the validation in one spot. The way these functions are now we'd still technically need to validate for `name` and `source` within the `CreateCollectionRequest` object since they can pass that directly. So ideally all the overloads would call one function where we do the actual validation.

So something like:

```java
public CollectionModel createCollection(CreateCollectionRequest createCollectionRequest) throws PineconeException {
  if (createCollectionRequest == null) {
      throw new PineconeValidationException("CreateCollectionRequest object cannot be null");
  }
  if (createCollectionRequest.getName() == null || createCollectionRequest.getName().isEmpty()) {
      throw new PineconeValidationException("name cannot be null or empty");
  }
  if (createCollectionRequest.getSource() == null || createCollectionRequest.getSource().isEmpty()) {
      throw new PineconeValidationException("source cannot be null or empty");
  }
  CollectionModel collection = null;
  try {
      collection = manageIndexesApi.createCollection(createCollectionRequest);
  } catch (ApiException apiException) {
      handleApiException(apiException);
  }
  return collection;
}

public CollectionModel createCollection(String name, String source) {
  CreateCollectionRequest createCollectionRequest = new CreateCollectionRequest().name(name).source(source);
  
  return createCollection(createCollectionRequest);
}
```

But then we lose the specificity of the error message containing the named arguments like `collectionName` / `sourceIndex`. This is probably why we abstract away the request class in `Index` and `AsyncIndex` and use all the explicit types, validate, and then build the request object at the end. I'm not really sure what's best here, maybe @ssmith-pc has some advice on what makes sense.

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -140,6 +141,10 @@ public void testCreateIndexWithAllFields() throws IOException {
         assertEquals(createdIndex, expectedIndex);
         assertEquals(requestCaptor.getValue().method(), "POST");
         assertEquals(requestCaptor.getValue().url().toString(), "https://api.pinecone.io/indexes");
+
+        // Test for null CreateIndexRequest object
+        PineconeException thrown = assertThrows(PineconeException.class, () -> client.createIndex(null));
```
austin-denoble, line 146: Should it be a `PineconeValidationException` inside `assertThrows`? Also for this and the other `assertThrows` in tests could we use `PineconeValidationException` just to be explicit?

```suggestion
        PineconeValidationException thrown = assertThrows(PineconeValidationException.class, () -> client.createIndex(null));
```

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -238,5 +243,105 @@ public void testConfigureIndex() throws IOException {
...
+        OkHttpClient mockClient = mock(OkHttpClient.class);
+        when(mockClient.newCall(any(Request.class))).thenReturn(mockCall);
+        when(mockCall.execute()).thenReturn(mockResponse);
+
+        Pinecone client = new Pinecone.Builder("testAPiKey").withOkHttpClient(mockClient).build();
+        CollectionModel collection = client.createCollection(createCollectionRequest);
+
+        ArgumentCaptor<Request> requestCaptor = ArgumentCaptor.forClass(Request.class);
+
+        verify(mockClient, times(1)).newCall(requestCaptor.capture());
```
austin-denoble, line 291: Since we're not using the `requestCaptor` to assert or test anything, I think it can just be removed entirely. I was using it in other places in this test class to assert on what the `OkHttpClient` client is being called with after a user makes a request to `pinecone.foo()`.

src/main/java/io/pinecone/clients/Pinecone.java
aulorbe: Ohhh I see. So only validate in the OG `createCollection` method?

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
aulorbe: Ah yes good call! I must've missed updating this one!

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
aulorbe: Oo okay thanks! Yeah TBH I kept it b/c I wasn't 100% sure what it was doing there and didn't want to mess w/the pattern 😓

src/main/java/io/pinecone/clients/Pinecone.java
aulorbe: (Resolved IRL)

austin-denoble (approved):
Looks good, thanks for iterating here. I think it's fine to leave the basic null checks in place for `configureIndex` and `createIndex` since you'll be doing another pass to work through those. Nice work!

There's one integration test that looks like it was lost in the shuffle, let's add that back in (testCreateCollectionFromInvalidIndex). You should be able to transpose the `name` and `source` values into your new function and the test is the same.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -180,16 +182,42 @@ public void testIndexFromDifferentMetricCollection() throws InterruptedException
     }
 
     @Test
-    public void testCreateCollectionFromInvalidIndex() {
```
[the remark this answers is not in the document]
aulorbe: Update: we spoke IRL and are added it back w/some changes!

.github/workflows/pr.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       run: gradle clean build
 
     - name: Run integration tests
-      run: gradle integrationTest
+      run: gradle integrationTest --info
```
rohanshah18, line 82: Do we need this?

src/integration/java/io/pinecone/helpers/IndexManager.java
```diff
@@ -126,8 +126,7 @@ public static Index createNewIndexAndConnectSync(Pinecone pinecone, String index
     }
 
     public static CollectionModel createCollection(Pinecone pinecone, String collectionName, String indexName, boolean waitUntilReady) throws InterruptedException {
-        CreateCollectionRequest createCollectionRequest = new CreateCollectionRequest().name(collectionName).source(indexName);
-        CollectionModel collection = pinecone.createCollection(createCollectionRequest);
+        CollectionModel collection = pinecone.createCollection(collectionName, indexName);
```
rohanshah18, line 129: Nice!!

PR #92, opened 2024-04-02
Add concurrent HashMap for storing indexNames and connection objects

rohanshah18:
## Problem

Everytime createIndexConnection() or createAsyncConnection() is called, we are internally calling describeIndex() to get index host which is required for creating blocking or async stubs. 

## Solution

Create a singleton ConcurrentHashMap of indexName (key) and PineconeConnection (value) such that every time the PineconeConnection object is instantiated, the indexName along with the corresponding PineconeConnection is stored into the HashMap. The entry is removed when the connection object is closed. 
Also for readability reasons, @ssmith-pc suggested to update the `createIndexConnection()` and `createAsyncIndexConnection()` methods to `getIndexConnection()` and `getAsyncConnection()` methods.

## Type of Change

## Test Plan

Added unit tests and ran integration tests.

### DIFF
--- a/src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
+++ b/src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
@@ -2,7 +2,6 @@ package io.pinecone.integration.dataPlane;
 
 import io.pinecone.clients.AsyncIndex;
 import io.pinecone.clients.Index;
-import io.pinecone.clients.Pinecone;
 import io.pinecone.configs.PineconeConnection;
 import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.proto.VectorServiceGrpc;
@@ -26,7 +25,6 @@ public class QueryErrorTest {
     @BeforeAll
     public static void setUp() throws IOException, InterruptedException {
         PineconeConnection connectionMock = mock(PineconeConnection.class);
-        Pinecone pineconeMock = mock(Pinecone.class);
 
         VectorServiceGrpc.VectorServiceBlockingStub stubMock = mock(VectorServiceGrpc.VectorServiceBlockingStub.class);
         VectorServiceGrpc.VectorServiceFutureStub asyncStubMock = mock(VectorServiceGrpc.VectorServiceFutureStub.class);
@@ -34,8 +32,8 @@ public class QueryErrorTest {
         when(connectionMock.getBlockingStub()).thenReturn(stubMock);
         when(connectionMock.getAsyncStub()).thenReturn(asyncStubMock);
 
-        index = new Index(pineconeMock, connectionMock, "some-index-name");
-        asyncIndex = new AsyncIndex(pineconeMock, connectionMock, "some-index-name");
+        index = new Index(connectionMock, "some-index-name");
+        asyncIndex = new AsyncIndex(connectionMock, "some-index-name");
     }
 
     @Test
--- a/src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
+++ b/src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
@@ -2,6 +2,7 @@ package io.pinecone.integration.dataPlane;
 
 import io.pinecone.clients.AsyncIndex;
 import io.pinecone.clients.Index;
+import io.pinecone.clients.Pinecone;
 import io.pinecone.configs.PineconeConnection;
 import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.proto.VectorServiceGrpc;
@@ -25,6 +26,7 @@ public class QueryErrorTest {
     @BeforeAll
     public static void setUp() throws IOException, InterruptedException {
         PineconeConnection connectionMock = mock(PineconeConnection.class);
+        Pinecone pineconeMock = mock(Pinecone.class);
 
         VectorServiceGrpc.VectorServiceBlockingStub stubMock = mock(VectorServiceGrpc.VectorServiceBlockingStub.class);
         VectorServiceGrpc.VectorServiceFutureStub asyncStubMock = mock(VectorServiceGrpc.VectorServiceFutureStub.class);
@@ -32,8 +34,8 @@ public class QueryErrorTest {
         when(connectionMock.getBlockingStub()).thenReturn(stubMock);
         when(connectionMock.getAsyncStub()).thenReturn(asyncStubMock);
 
-        index = new Index(connectionMock);
-        asyncIndex = new AsyncIndex(connectionMock);
+        index = new Index(pineconeMock, connectionMock, "some-index-name");
+        asyncIndex = new AsyncIndex(pineconeMock, connectionMock, "some-index-name");
     }
 
     @Test
--- a/src/main/java/io/pinecone/clients/AsyncIndex.java
+++ b/src/main/java/io/pinecone/clients/AsyncIndex.java
@@ -31,6 +31,10 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
     private static final Logger logger = LoggerFactory.getLogger(AsyncIndex.class);
 
     public AsyncIndex(Pinecone pinecone, PineconeConnection connection, String indexName) {
+        if (pinecone == null) {
+            throw new PineconeValidationException("Pinecone object cannot be null.");
+        }
+
         if (connection == null) {
             throw new PineconeValidationException("Pinecone connection object cannot be null.");
         }
@@ -221,7 +225,7 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
     public void close() {
         try {
             logger.debug("closing channel");
-            pinecone.getConnectionsMap().remove(indexName);
+            pinecone.closeConnection(indexName);
             connection.getChannel().shutdownNow().awaitTermination(5, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
             logger.warn("Channel shutdown interrupted before termination confirmed");
--- a/src/main/java/io/pinecone/clients/AsyncIndex.java
+++ b/src/main/java/io/pinecone/clients/AsyncIndex.java
@@ -25,13 +25,18 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
 
     private final PineconeConnection connection;
     private final VectorServiceGrpc.VectorServiceFutureStub asyncStub;
+    private final Pinecone pinecone;
+    private final String indexName;
 
     private static final Logger logger = LoggerFactory.getLogger(AsyncIndex.class);
 
-    public AsyncIndex(PineconeConnection connection) {
+    public AsyncIndex(Pinecone pinecone, PineconeConnection connection, String indexName) {
         if (connection == null) {
             throw new PineconeValidationException("Pinecone connection object cannot be null.");
         }
+
+        this.pinecone = pinecone;
+        this.indexName = indexName;
         this.connection = connection;
         this.asyncStub = connection.getAsyncStub();
     }
@@ -216,6 +221,7 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
     public void close() {
         try {
             logger.debug("closing channel");
+            pinecone.getConnectionsMap().remove(indexName);
             connection.getChannel().shutdownNow().awaitTermination(5, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
             logger.warn("Channel shutdown interrupted before termination confirmed");
--- a/src/main/java/io/pinecone/clients/AsyncIndex.java
+++ b/src/main/java/io/pinecone/clients/AsyncIndex.java
@@ -10,11 +10,8 @@ import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.proto.*;
 import io.pinecone.unsigned_indices_model.QueryResponseWithUnsignedIndices;
 import io.pinecone.unsigned_indices_model.VectorWithUnsignedIndices;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.List;
-import java.util.concurrent.TimeUnit;
 
 public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertResponse>,
         ListenableFuture<QueryResponseWithUnsignedIndices>,
@@ -25,21 +22,13 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
 
     private final PineconeConnection connection;
     private final VectorServiceGrpc.VectorServiceFutureStub asyncStub;
-    private final Pinecone pinecone;
     private final String indexName;
 
-    private static final Logger logger = LoggerFactory.getLogger(AsyncIndex.class);
-
-    public AsyncIndex(Pinecone pinecone, PineconeConnection connection, String indexName) {
-        if (pinecone == null) {
-            throw new PineconeValidationException("Pinecone object cannot be null.");
-        }
-
+    public AsyncIndex(PineconeConnection connection, String indexName) {
         if (connection == null) {
             throw new PineconeValidationException("Pinecone connection object cannot be null.");
         }
 
-        this.pinecone = pinecone;
         this.indexName = indexName;
         this.connection = connection;
         this.asyncStub = connection.getAsyncStub();
@@ -223,12 +212,7 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
      */
     @Override
     public void close() {
-        try {
-            logger.debug("closing channel");
-            pinecone.closeConnection(indexName);
-            connection.getChannel().shutdownNow().awaitTermination(5, TimeUnit.SECONDS);
-        } catch (InterruptedException e) {
-            logger.warn("Channel shutdown interrupted before termination confirmed");
-        }
+        Pinecone.closeConnection(indexName);
+        connection.close();
     }
 }
--- a/src/main/java/io/pinecone/clients/Index.java
+++ b/src/main/java/io/pinecone/clients/Index.java
@@ -21,16 +21,20 @@ public class Index implements IndexInterface<UpsertResponse,
         DescribeIndexStatsResponse> {
 
     private final PineconeConnection connection;
+    private final Pinecone pinecone;
+    private final String indexName;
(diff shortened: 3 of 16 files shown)


src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -216,6 +221,7 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
     public void close() {
         try {
             logger.debug("closing channel");
+            pinecone.getConnectionsMap().remove(indexName);
```
rohanshah18, line 224: remove the connection from the connections map here

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -205,6 +209,7 @@ public DescribeIndexStatsResponse describeIndexStats(Struct filter) {
     public void close() {
         try {
             logger.debug("closing channel");
+            pinecone.getConnectionsMap().remove(indexName);
```
rohanshah18, line 212: remove the connection from the connections map here

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -11,12 +11,15 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
 public class Pinecone {
 
     private final ManageIndexesApi manageIndexesApi;
     private final PineconeConfig config;
+    private static final ConcurrentHashMap<String, PineconeConnection> connectionsMap = new ConcurrentHashMap<>();
 
-    private Pinecone(PineconeConfig config, ManageIndexesApi manageIndexesApi) {
+    Pinecone(PineconeConfig config, ManageIndexesApi manageIndexesApi) {
```
rohanshah18, line 22: removed the private modifier for unit testing

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -11,12 +11,15 @@
 import org.openapitools.client.api.ManageIndexesApi;
 import org.openapitools.client.model.*;
 
+import java.util.concurrent.ConcurrentHashMap;
+
 public class Pinecone {
 
     private final ManageIndexesApi manageIndexesApi;
     private final PineconeConfig config;
+    private static final ConcurrentHashMap<String, PineconeConnection> connectionsMap = new ConcurrentHashMap<>();
```
rohanshah18, line 20: create ConcurrentHashMap so its thread safe

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -149,14 +152,27 @@ public void deleteCollection(String collectionName) throws PineconeException {
         }
     }
 
-    public Index createIndexConnection(String indexName) {
-        PineconeConnection connection = new PineconeConnection(config, indexName);
-        return new Index(connection);
+    public Index getIndexConnection(String indexName) {
```
rohanshah18, line 155: renamed the methods for readability reasons

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -149,14 +152,27 @@ public void deleteCollection(String collectionName) throws PineconeException {
...
+        return new Index(this, connection, indexName);
+    }
+
+    public AsyncIndex getAsyncIndexConnection(String indexName) {
+        PineconeConnection connection = getConnection(indexName);
+        return new AsyncIndex(this, connection, indexName);
+    }
+
+    private PineconeConnection getConnection(String indexName) {
+        return connectionsMap.computeIfAbsent(indexName, key -> new PineconeConnection(config));
```
rohanshah18, line 167: @ssmith-pc made a great suggestion to use computeIfAbsent() which will compute/create a new connection if indexName is not found in the map. Initially, it was an if-else loop so we would have to either acquire and release the locks to make the process of checking if the key doesn't exists and creating the connection object atomic.

ssmith-pc:
> Also for readability reasons, @ssmith-pc suggested to update the createIndexConnection() and createAsyncIndexConnection() methods to getIndexConnection() and getAsyncConnection() methods

FWIW, I was also the one that suggested you rename them from `getIndexConnection` to `createIndexConnection`. Sorry!

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: Where is the `config` object coming from?

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -149,14 +152,27 @@ public void deleteCollection(String collectionName) throws PineconeException {
...
+    public Index getIndexConnection(String indexName) {
+        config.setHost(getIndexHost(indexName));
+        PineconeConnection connection = getConnection(indexName);
+        return new Index(this, connection, indexName);
+    }
+
+    public AsyncIndex getAsyncIndexConnection(String indexName) {
+        PineconeConnection connection = getConnection(indexName);
+        return new AsyncIndex(this, connection, indexName);
+    }
```
ssmith-pc, line 164: Do we not need to do the host lookup in this method?

rohanshah18: Good catch!!

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -149,14 +152,27 @@ public void deleteCollection(String collectionName) throws PineconeException {
...
+
+    private PineconeConnection getConnection(String indexName) {
+        return connectionsMap.computeIfAbsent(indexName, key -> new PineconeConnection(config));
+    }
+
+    ConcurrentHashMap<String, PineconeConnection> getConnectionsMap() {
+        return connectionsMap;
     }
```
austin-denoble, line 172: Is this primarily for testing, or do we expect users to interact with this?

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
```diff
@@ -0,0 +1,118 @@
...
+        when(pinecone1.getIndexConnection("index1")).thenReturn(index1);
+        when(pinecone2.getIndexConnection("index2")).thenReturn(index2);
+
+        Index resultIndex1 = pinecone1.getIndexConnection("index1");
+        Index resultIndex2 = pinecone2.getIndexConnection("index2");
+
+        verify(pinecone1).getIndexConnection("index1");
+        verify(pinecone2).getIndexConnection("index2");
+        assertNotSame(resultIndex1, resultIndex2);
+    }
```
ssmith-pc, line 36: I'm not really clear what is being tested in this method, since everything is being mocked, even the method that is being tested? 

Seems like we need to mock `getIndexHost()` and potentially `getConnection`, but then otherwise we should let the actual `getIndexConnection` method run.

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.openapitools.client.ApiException;
+import org.openapitools.client.api.ManageIndexesApi;
+import org.openapitools.client.model.IndexModel;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ConnectionsMapTest {
+
```
ssmith-pc, line 15: We should also try to add some tests around the connection being re-used. In other words, if someone calls `getIndexConnection("my-index")` multiple times, they should get the same instance back each time.

ssmith-pc (commented):
Looking good, added a few comments. 

At some point we'll need to add clear guidance on the proper (safe) usage especially in a multi-threaded context.

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    public void testMultiplePineconeInstancesForIndexes() {
+        Pinecone pinecone1 = mock(Pinecone.class);
+        Pinecone pinecone2 = mock(Pinecone.class);
+
+        assertNotSame(pinecone1, pinecone2);
+
+        // Tests for blocking indexes
+        Index index1 = mock(Index.class);
+        Index index2 = mock(Index.class);
```
austin-denoble, line 25: Maybe this is me being unfamiliar with mockito, but I'm confused what these tests are doing if we're just mocking `Pinecone` which contains the `ConcurrentHashMap`. Aren't we just stubbing the behavior now instead of testing the internals of the class?

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
```diff
@@ -25,15 +26,16 @@ public class QueryErrorTest {
     @BeforeAll
     public static void setUp() throws IOException, InterruptedException {
         PineconeConnection connectionMock = mock(PineconeConnection.class);
```
aulorbe, line 28: This might be a dumb question, but shouldn't the `pineconeMock` be able to be created from the `connectionMock` obj?

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
```diff
@@ -0,0 +1,134 @@
...
+        when(pinecone2.describeIndex("index2")).thenReturn(indexModel);
+
+        Index index1Pinecone1 = pinecone1.getIndexConnection("index1");
+        Index index1Pinecone2 = pinecone1.getIndexConnection("index2");
+        Index index2Pinecone1 = pinecone2.getIndexConnection("index1");
+        Index index2Pinecone2 = pinecone2.getIndexConnection("index2");
+        assertEquals(pinecone1.getConnectionsMap().size(), 2);
+        assertEquals(pinecone2.getConnectionsMap().size(), 2);
+
+        index1Pinecone1.close();
```
aulorbe, line 40: Should you just put this all in a try with resources loop?

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @Test
+    public void testMultiplePineconeInstancesForIndexes() {
+        String host = "some-host";
+        ManageIndexesApi manageIndexesApi = Mockito.mock(ManageIndexesApi.class);
+        IndexModel indexModel = new IndexModel();
+        indexModel.setHost(host);
+
+        PineconeConfig config = new PineconeConfig("testApiKey");
+        config.setHost(host);
+        Pinecone pinecone1 = new Pinecone(config, manageIndexesApi);
```
aulorbe, line 25: So `pinecone<SomeNumber>` here is the "instance" referred to in this test's name?

aulorbe: The naming of things is here for me to grok here. Like, the word `pinecone` doesn't really connote any action to me, it's just sorta a name for a service, you know?

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+        asyncIndex1Pinecone1.close();
+        asyncIndex2Pinecone1.close();
+        asyncIndex1Pinecone2.close();
+        asyncIndex2Pinecone2.close();
+    }
+
+    @Test
+    public void testConnectionRemovedOnIndexClose() throws ApiException {
+        String host = "some-host";
```
aulorbe, line 78: Nit, but I'd stick to camelcase for the var values. The hyphens seem Python-y in a weird way

aulorbe (approved):
Some Qs just  for my understanding and nits

src/main/java/io/pinecone/clients/AsyncIndex.java
rohanshah18: Good point! Updated this to close connection in Pinecone class

src/main/java/io/pinecone/clients/Index.java
rohanshah18: Good point. 
1. Since pinecone object is instantiated in the Pinecone class and the constructors new Index() + new AsyncIndex() are called within Pinecone class, they are not going to be null. But I have added the null check for pinecone object in the constructor. @ssmith-pc thoughts?
2. If map.get(indexName) is not found, map.remove won't delete the connection. So the checks are not required on the indexName.

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
rohanshah18: Updated the test and in future, we will like to tests for the contents as well

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
rohanshah18: Updated the tests for number of connections created and in future, we will like to test for the contents as well

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -21,16 +21,20 @@ public class Index implements IndexInterface<UpsertResponse,
...
+    private final String indexName;
     private final VectorServiceGrpc.VectorServiceBlockingStub blockingStub;
 
     private static final Logger logger = LoggerFactory.getLogger(Index.class);
 
-    public Index(PineconeConnection connection) {
+    public Index(Pinecone pinecone, PineconeConnection connection, String indexName) {
         if (connection == null) {
             throw new PineconeValidationException("Pinecone connection object cannot be null.");
         }
```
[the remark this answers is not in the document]
rohanshah18: Good point! I was thinking of renaming it to PineconeException but fine with NullPointerException as well

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes, for testing only

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
rohanshah18: Updated the tests

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
rohanshah18: Yes that's another option, it'll close all of the connections

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
rohanshah18: Ideally we try to mimic the behavior of how the variable's value will look like but since these are values for unit test variables, it won't matter much

src/integration/java/io/pinecone/integration/dataPlane/QueryErrorTest.java
rohanshah18: the constructor for Pinecone accepts PineconeConfig config and ManageIndexesApi manageIndexesApi, so we can't instantiate it using PineconeConnection class

src/test/java/io/pinecone/clients/ConnectionsMapTest.java
rohanshah18: Yes, its an instance of the class pinecone. We want to have multiple instances so the numbers represent what instance of the Pinecone class is the object. Next, we want to create Pinecone Connections for 2 different indexes per pinecone instance. Finally, verify that the HashMap is shared between the instances and if index 1 connection was created by pinecone1, than pinecone2 should be utilizing the same connection for index1 and not create a new connection

src/main/java/io/pinecone/clients/Index.java
rohanshah18: @ssmith-pc another option I was thinking for closeConnection(indexName) is making it static, so I can call the method from close() of Index and AsyncIndex classes with class Pinecone instead of checking if the pinecone object is available or not.

ssmith-pc: > If map.get(indexName) is not found, map.remove won't delete the connection. So the checks are not required on the indexName.
Check the Conc Hash map docs to confirm, but I think `CHM.remove(null)` will throw an NPE. There are a few other cases in the API where the client is expected to pass in the name of an index and I don't see it being validated anywhere that's a non-empty string. Would be a good idea to do some basic input validation before sending to server.

> another option I was thinking for closeConnection(indexName) is making it static
Is it true that users are never expected to call this, it's just an internal method to close the connection? I think that would be ok to make it static since it only interacts with the static CHM.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -149,14 +152,32 @@ public void deleteCollection(String collectionName) throws PineconeException {
...
+        config.setHost(getIndexHost(indexName));
+        PineconeConnection connection = getConnection(indexName);
+        return new Index(this, connection, indexName);
+    }
+
+    public AsyncIndex getAsyncIndexConnection(String indexName) {
+        config.setHost(getIndexHost(indexName));
+        PineconeConnection connection = getConnection(indexName);
+        return new AsyncIndex(this, connection, indexName);
+    }
```
ssmith-pc, line 165: For example, we should validate `indexName` is non-empty here.

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -216,6 +225,7 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
     public void close() {
         try {
             logger.debug("closing channel");
+            pinecone.closeConnection(indexName);
             connection.getChannel().shutdownNow().awaitTermination(5, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
```
ssmith-pc, line 230: Rather than get the channel from the connection and shutting it down, can't we just call `connection.close()`?

src/main/java/io/pinecone/clients/Index.java
rohanshah18: 1. That's correct, CHM.remove(null) throws NPE since it calls containsKey() first.
CHM.remove(key) where key doesn't exists in the map results in no action. 

2. Yes, users are not expected to interact with closeConnection(indexName).

src/main/java/io/pinecone/clients/AsyncIndex.java
rohanshah18: Yes, totally!!

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Right, there's no need to attempt creating a connection

PR #93, opened 2024-04-04
Abstract away Request objs in the ConfigureIndex method

aulorbe:
## Problem

Currently, when users call the `ConfigureIndex` method (in `Pinecone.java`), they have to pass a `ConfigureIndexRequest` object. This object in turn, needs to be created with a `ConfigureIndexRequestSpec` object, which itself needs to be created with a `ConfigureIndexRequestSpec`, WHICH THEN needs to be created with a `ConfigureIndexRequestSpecPod` object.

This is hella cumbersome.

## Solution

This PR abstracts away these Request objects, so the user doesn't have to deal with them at all. Instead, the user can simply call the `ConfigureIndex` method 1 of 2 ways:
- `ConfigureIndex("indexName", someNumberOfReplicas)`
- `ConfigureIndex("indexName", "somePodType")`

Relevant Asana ticket: https://app.asana.com/0/1203260648987893/1206994598304485/f

## Type of Change

## Test Plan

Unit tests and updated integration tests included in PR.

### DIFF
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
@@ -1,12 +1,16 @@
 package io.pinecone.integration.controlPlane.pod;
 
 import io.pinecone.clients.Pinecone;
-import io.pinecone.exceptions.PineconeForbiddenException;
 import io.pinecone.exceptions.PineconeBadRequestException;
+import io.pinecone.exceptions.PineconeForbiddenException;
 import io.pinecone.exceptions.PineconeNotFoundException;
 import io.pinecone.helpers.TestIndexResourcesManager;
-import org.junit.jupiter.api.*;
-import org.openapitools.client.model.*;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.openapitools.client.model.IndexModel;
+import org.openapitools.client.model.IndexModelStatus;
+import org.openapitools.client.model.PodSpec;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -22,48 +22,6 @@ private Pinecone(PineconeConfig config, ManageIndexesApi manageIndexesApi) {
         this.manageIndexesApi = manageIndexesApi;
     }
 
-    public static class Builder {
```
aulorbe, line 25: Uh... I don't know why these removals of `Builder` are in here...I'll look into it

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -16,27 +20,39 @@
...
+    private static final Pinecone controlPlaneClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
     private static String indexName;
 
     @BeforeAll
-    public static void setUp() throws InterruptedException {;
+    public static void setUp() throws InterruptedException {
         indexName = indexManager.getPodIndexName();
     }
 
+    private static void waitUntilIndexStateIsReady(String indexName) throws InterruptedException {
```
aulorbe, line 31: This is just my linter -- brought this method up here

aulorbe:
@austin-denoble re: this question, I just used the one in IntelliJ. The one you get to by doing `Code` > `Reformat file`

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -87,14 +45,41 @@ public IndexModel describeIndex(String indexName) throws PineconeException {
...
-        }
+    public IndexModel configureIndex(String indexName, String podType, Integer replicas) throws PineconeValidationException {
         if (indexName == null || indexName.isEmpty()) {
-            throw new PineconeValidationException("Index name cannot be null or empty");
+            throw new PineconeValidationException("indexName cannot be null or empty");
+        }
+
+        // Set default values if parameters are not provided
+        String defaultPodType = "p1.x1";
+        int defaultReplicas = 1;
```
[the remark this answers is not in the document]
aulorbe: Hmmm... I might be misunderstanding everything, but I don't I think... how else do we pass default values when the user only provides either `podType` or `replicas`?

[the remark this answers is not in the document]
austin-denoble: The user is either updating replicas, podType, or both. We wouldn't want to default any values for a configure / PATCH request because we may be overriding the existing value. Does that make sense?

I don't think these variables are actually being used anywhere in the code though, right?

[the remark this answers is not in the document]
aulorbe: Ohhhh

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -46,11 +62,8 @@ public void configureIndexWithInvalidIndexName() {
 
     @Test
     public void configureIndexExceedingQuota() {
-        ConfigureIndexRequestSpecPod pod = new ConfigureIndexRequestSpecPod().replicas(400);
-        ConfigureIndexRequestSpec spec = new ConfigureIndexRequestSpec().pod(pod);
-        ConfigureIndexRequest configureIndexRequest = new ConfigureIndexRequest().spec(spec);
         try {
-            controlPlaneClient.configureIndex(indexName, configureIndexRequest);
+            controlPlaneClient.configureIndex(indexName, 400);
```
austin-denoble, line 66: I love how much cleaner this is with the new argument patterns!

austin-denoble (approved):
Looks good! Thanks for iterating. 🚢

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
aulorbe: 🎉 !

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -87,14 +45,31 @@ public IndexModel describeIndex(String indexName) throws PineconeException {
...
-        if (configureIndexRequest == null) {
-            throw new PineconeValidationException("ConfigureIndexRequest object cannot be null");
-        }
+    public IndexModel configureIndex(String indexName, String podType, Integer replicas) throws PineconeValidationException {
         if (indexName == null || indexName.isEmpty()) {
-            throw new PineconeValidationException("Index name cannot be null or empty");
+            throw new PineconeValidationException("indexName cannot be null or empty");
+        }
+
+        if (podType == null && replicas == null) {
```
rohanshah18, line 53: Any reason why we dont want to check for empty string values for podtype and replicas?

PR #95, opened 2024-04-05
[Fix] Accept additional properties in API JSON responses

jhamon:
## Problem

When new properties are added to API responses, this causes the Java client to error. This prevents us from making additive API changes that would otherwise be non-breaking to the end user.

## Solution

- Add tests that exercise the JSON parsing behavior in `src/test/java/org/openapitools/client/JsonParsingTest.java` and ensure the presence of additional properties will not cause the SDK to error. These tests use JSON fixtures added in `src/test/resources`.
- Regenerate the control plane rest client using the OpenAPI [java client generator](https://openapi-generator.tech/docs/generators/java/). This time, configuring the generator with `disallowAdditionalPropertiesIfNotPresent=false` should activate the correct JSON parsing behavior in the generated code.

## How the code was generated

The code generation step was performed with this command:

```bash
mkdir codegen
mkdir codegen/gen
cd codegen

docker run --rm -v $(pwd):/workspace openapitools/openapi-generator-cli:v7.0.1 generate \
    --input-spec $SPEC_FILE \
    --additional-properties=dateLibrary='java8',disallowAdditionalPropertiesIfNotPresent=false \
    --generator-name java \
    --output /workspace/gen/java

## We want to replace all generated files in src, so remove the old ones
rm -rf ../src/main/java/org/openapitools/client

# We don't want an entire generated project structure, just the java src files.
cp -r gen/java/src/main/java/org/openapitools/client ../src/main/java/org/openapitools/client
```

It took some trial and error to discover an openapi generator version and spec version that produced similar results to the ones Rohan created previously using steps and commands that were not captured anywhere. These steps will be scripted and automated in a future PR to remove this element of guesswork.

## Type of Change

## Test Plan

Since the new tests exercise all the generated code up to the boundary where a response is returned from the okhttp library used for making requests, that gives a high confidence this will work if unexpected fields appear in future API responses.

### DIFF
--- a/src/test/java/org/openapitools/client/JsonParsingTest.java
+++ b/src/test/java/org/openapitools/client/JsonParsingTest.java
@@ -1,0 +1,116 @@
+package org.openapitools.client;
+
+import okhttp3.*;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.openapitools.client.api.ManageIndexesApi;
+import org.openapitools.client.model.*;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
+
+public class JsonParsingTest {
+    private OkHttpClient okHttpClient;
+    private ManageIndexesApi api;
+
+    @BeforeEach
+    public void setup() {
+        okHttpClient = mock(OkHttpClient.class);
+        ApiClient apiClient = new ApiClient(okHttpClient);
+        api = new ManageIndexesApi(apiClient);
+    }
+
+    private void setupMockResponse(String jsonResponse) throws Exception {
+        Call call = mock(Call.class);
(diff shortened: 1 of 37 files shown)


src/test/java/org/openapitools/client/JsonParsingTest.java
```diff
@@ -0,0 +1,116 @@
+package org.openapitools.client;
```
austin-denoble, line 1: I know this issue basically forces it, but good idea for adding a testing section focused on some of the generated bits. 👍

src/test/resources/createIndexResponse.withUnknownProperties.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "serverless-index",
+  "metric": "cosine",
+  "dimension": 1536,
+  "ux_feedback": "much wow",
+  "status": {
+    "ready": true,
+    "state": "Ready",
+    "reticulating_splines": true
```
austin-denoble, line 9: lol 😭

austin-denoble (approved):
Thanks again for getting this resolved and adding tests! The logic around the flag you needed to pass is really gnarly.

PR #94, opened 2024-04-05
Add new CreateServerlessIndex method

aulorbe:
## Problem

It's currently cumbersome to force users to create `CreateIndexRequest` object, a `ServerlessSpec` object, and the required `enum`s (`IndexMetric` and `CloudEnum`. 

## Solution

This PR is a *start* on the road towards removing the `createIndex` method. 

In this PR, you will find a new method called `CreateServerlessIndex`. This is what users will use to create a serverless index. All they have to do is pass strings/ints for `indexName`, `metric`, `dimension`, `cloud`, and `region`. All validation of the inputs, transforms into `enum` objects, and construction of the `Request` objects takes place within the function itself.

Note: 
- From reading and asking ChatGPT, it seems that it's more idiomatic in Java to force users to construct the `enum`s (necessary for `metric` and `cloud`), rather than pass in strings, since constructing `enum`s directly is more type-safe and self-documenting. _However_, I think ease-of-use trumps those concerns, since we are building these SDKs primarily to increase developer experience.

LMK if you agree in the comments! 

Asana ticket: https://app.asana.com/0/1203260648987893/1206510790411726/f

## Follow-ups
- Next up will be doing the same thing, but for a new `CreatePodsIndex` method ([ticket](https://app.asana.com/0/1203260648987893/1206994598304491/f))
- After that, I will remove `CreateIndex` from the client altogether, updating unit tests and integration tests as necessary.
    - Within this PR ^, I will also make the necessary documentation updates to Rohan's README, located [here](https://github.com/pinecone-io/pinecone-java-client/blob/c3762bb3d1e34d454047e4ffec875b3a4b159e88/README.md).

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update
- [ ] Infrastructure change (CI configs, etc)
- [ ] Non-code change (docs, etc)
- [ ] None of the above: (explain here)

## Test Plan

Unit test & integration tests included in PR.

PR #98, opened 2024-04-08
Add new createPodsIndex method

aulorbe:
## Problem

It's currently cumbersome to make the user create a bunch of objects from our generated code, e.g. `CreateIndexRequest`, etc. just to create an index.

## Solution

Piggybacking off of [my last PR](https://github.com/pinecone-io/pinecone-java-client/pull/94), this PR creates a new method called `CreatePodsIndex`. Together with `CreateServerlessIndex`, we can now deprecate `createIndex` altogether (in a forthcoming PR). 

The overloaded method signatures for the `createPodsIndex` methods are as follows. (Please lmk if you having feelings about these options. It was a bit of a crapshoot figuring out which I thought would be most useful to users.)

- The minimal required parameters (note: `podType` is required by the backend, not our OpenAPI spec, FWIW):
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    
- Minimal + metric: 
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - String metric

- Minimal + metric + metadata:
    -  String indexName
    - Integer dimension
    - String environment
    - String podType
    - String metric
    - PodSpecMetadataConfig metadataConfig

- Minimal + metric + source collection
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - String metric
    - String sourceCollection

- Minimal + pods
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - Integer pods

- Minimal + pods + metadata:
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - Integer pods
    - PodSpecMetadataConfig metadataConfig

- Minimal + replicas and shards
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - Integer replicas
    - Integer shards

- Minimal + replicas, shards + metadata:
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - Integer replicas
    - Integer shards
    - PodSpecMetadataConfig metadataConfig

- Max: 
    - String indexName
    - Integer dimension
    - String environment
    - String podType
    - String metric
    - Integer replicas
    - Integer shards
    - Integer pods
    - PodSpecMetadataConfig metadataConfig
    - String sourceCollection

## Callouts: 
- When a user provides `pods`, they should not provide `shards` or `replicas`, and vice versa (unless in the "max" method)
- If `replicas * shards` do not equal `pods`, an error is thrown
- Like with `createServerlessIndex` and `cloud`, the user passes `metric` as a string, and it's internally validated against the `IndexMetric.Enum` types
- Besides checking for nulls and empties, there are no checks for the actual values in `environment`, `podType`, or `sourceCollection`.

## Type of Change

- [x] This change requires a documentation update

## Test Plan

Unit and integration tests included.

Note: in the next PR I will deprecate `createIndex` altogether. Within this PR, I will update the `TestIndexResourcesManager` singleton to work with the new methods. I will also update the README, as per this [ticket](https://app.asana.com/0/1203260648987893/1206994598304493/f).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -209,10 +209,10 @@ nexusPublishing {
     }
 }
 
-//signing {
-//    def signingKeyId = findProperty("signingKeyId")
-//    def signingKey = findProperty("signingKey")
-//    def signingPassword = findProperty("signingPassword")
-//    useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
-//    sign publishing.publications.pineconeClientMaven
-//}
+signing {
+    def signingKeyId = findProperty("signingKeyId")
+    def signingKey = findProperty("signingKey")
+    def signingPassword = findProperty("signingPassword")
+    useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
+    sign publishing.publications.pineconeClientMaven
+}
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,6 @@ dependencies {
     implementation 'io.gsonfire:gson-fire:1.8.5'
     implementation 'org.openapitools:jackson-databind-nullable:0.2.6'
     implementation 'com.google.protobuf:protobuf-java:3.25.2'
-    implementation 'org.junit.jupiter:junit-jupiter-api:5.8.0'
     compileOnly "org.apache.tomcat:annotations-api:6.0.53" // necessary for Java 9+
 
     testImplementation("com.squareup.okhttp3:mockwebserver:4.12.0")
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     implementation 'io.gsonfire:gson-fire:1.8.5'
     implementation 'org.openapitools:jackson-databind-nullable:0.2.6'
     implementation 'com.google.protobuf:protobuf-java:3.25.2'
+    implementation 'org.junit.jupiter:junit-jupiter-api:5.8.0'
     compileOnly "org.apache.tomcat:annotations-api:6.0.53" // necessary for Java 9+
 
     testImplementation("com.squareup.okhttp3:mockwebserver:4.12.0")
--- a/build.gradle
+++ b/build.gradle
@@ -209,10 +209,10 @@ nexusPublishing {
     }
 }
 
-signing {
-    def signingKeyId = findProperty("signingKeyId")
-    def signingKey = findProperty("signingKey")
-    def signingPassword = findProperty("signingPassword")
-    useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
-    sign publishing.publications.pineconeClientMaven
-}
+//signing {
+//    def signingKeyId = findProperty("signingKeyId")
+//    def signingKey = findProperty("signingKey")
+//    def signingPassword = findProperty("signingPassword")
+//    useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
+//    sign publishing.publications.pineconeClientMaven
+//}
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
@@ -63,7 +63,7 @@ public class ConfigureIndexTest {
     @Test
     public void configureIndexExceedingQuota() {
         try {
-            controlPlaneClient.configureIndex(indexName, 400);
+            controlPlaneClient.configureIndex(indexName, 600);
 
             fail("Expected to throw PineconeForbiddenException");
         } catch (PineconeForbiddenException expected) {
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
@@ -64,10 +64,9 @@ public class ConfigureIndexTest {
     public void configureIndexExceedingQuota() {
         try {
             controlPlaneClient.configureIndex(indexName, 600);
-
             fail("Expected to throw PineconeForbiddenException");
         } catch (PineconeForbiddenException expected) {
-            assertTrue(expected.getLocalizedMessage().contains("quota"));
+            assertTrue(expected.getLocalizedMessage().contains("reached the max pods allowed"));
         }
     }
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
@@ -103,7 +103,7 @@ public class CreateDescribeListAndDeleteIndexTest {
         Integer dimension = 3;
         String environment = "us-east-1-aws";
         String podType = "p1.x1";
-        assertThrows(PineconeValidationException.class, () -> {
+        assertThrows(PineconeBadRequestException.class, () -> {
             controlPlaneClient.createPodsIndex(podIndexName, dimension, environment, podType);
         });
     }
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
@@ -21,13 +21,13 @@ public class CreateDescribeListAndDeleteIndexTest {
     private static String indexPodType;
     private static final String environment =    System.getenv("PINECONE_ENVIRONMENT");
 
-//    @BeforeAll
-//    public static void setUp() throws InterruptedException {
-//        indexName = indexManager.getPodIndexName();
-//        indexDimension = indexManager.getDimension();
-//        IndexModel podIndex = indexManager.getPodIndexModel();
-//        indexPodType = podIndex.getSpec().getPod().getPodType();
-//    }
+    @BeforeAll
+    public static void setUp() throws InterruptedException {
+        indexName = indexManager.getPodIndexName();
+        indexDimension = indexManager.getDimension();
+        IndexModel podIndex = indexManager.getPodIndexModel();
+        indexPodType = podIndex.getSpec().getPod().getPodType();
+    }
 
     @Test
     public void describeAndListIndex() {
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
@@ -3,13 +3,13 @@ package io.pinecone.integration.controlPlane.pod;
 import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeBadRequestException;
 import io.pinecone.exceptions.PineconeUnmappedHttpException;
+import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.helpers.RandomStringBuilder;
 import io.pinecone.helpers.TestIndexResourcesManager;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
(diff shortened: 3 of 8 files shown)


src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -13,21 +13,25 @@
...
     private final ManageIndexesApi manageIndexesApi;
     private final PineconeConfig config;
-    private static final ConcurrentHashMap<String, PineconeConnection> connectionsMap = new ConcurrentHashMap<>();
 
     Pinecone(PineconeConfig config, ManageIndexesApi manageIndexesApi) {
         this.config = config;
         this.manageIndexesApi = manageIndexesApi;
     }
 
+    static void closeConnection(String indexName) {
```
aulorbe, line 31: Just lints; ignore

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -109,7 +307,8 @@ public IndexModel describeIndex(String indexName) throws PineconeException {
         return indexModel;
     }
 
-    public IndexModel configureIndex(String indexName, String podType, Integer replicas) throws PineconeValidationException {
+    public IndexModel configureIndex(String indexName, String podType, Integer replicas) throws
```
aulorbe, line 310: lint

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -171,7 +370,8 @@ public void deleteIndex(String indexName) throws PineconeException {
         }
     }
 
-    public CollectionModel createCollection(String collectionName, String sourceIndex) throws PineconeValidationException {
+    public CollectionModel createCollection(String collectionName, String sourceIndex) throws
```
aulorbe, line 373: lint

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -221,7 +421,7 @@ public void deleteCollection(String collectionName) throws PineconeException {
     }
 
     public Index getIndexConnection(String indexName) {
-        if(indexName == null || indexName.isEmpty()) {
+        if (indexName == null || indexName.isEmpty()) {
```
aulorbe, line 424: lint

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -231,7 +431,7 @@ public Index getIndexConnection(String indexName) {
     }
 
     public AsyncIndex getAsyncIndexConnection(String indexName) {
-        if(indexName == null || indexName.isEmpty()) {
+        if (indexName == null || indexName.isEmpty()) {
```
aulorbe, line 434: lint

src/test/java/io/pinecone/PineconeIndexOperationsTest.java
```diff
@@ -87,7 +91,7 @@ public void testCreateServerlessIndex() throws IOException {
         PineconeValidationException thrownNullMetric = assertThrows(PineconeValidationException.class,
                 () -> client.createServerlessIndex("testServerlessIndex", null, 3, "aws", "us-west-2"));
         assertEquals("Metric cannot be null or empty. Must be one of " + Arrays.toString(IndexMetric.values()),
-            thrownNullMetric.getMessage());
+                thrownNullMetric.getMessage());
```
aulorbe, line 94: lint

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -69,6 +69,44 @@ public void createIndexWithPodsAndPodType() {
...
+
+    @Test
+    public void CreatePodsIndexWithInvalidIndexName() {
+        String podIndexName = "Invalid-name";
+        Integer dimension = 3;
+        String environment = "us-east-1-aws";
+        String podType = "p1.x1";
+
+        // Empty indexName
+        assertThrows(PineconeValidationException.class, () -> {
```
aulorbe, line 105: Only put this here, since I didn't want to duplicate all of the unit test logic in the integration tests. LMK thoughts on whether I should test more logic here (e.g. negative dimension passed).

aulorbe:
@ssmith-pc Just tagging you in case you want to look. No pressure!

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: We order functions by usage, so the important ones are at the top and close() is not something we expect end users to consume.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -99,6 +104,301 @@ public IndexModel createServerlessIndex(String indexName, String metric, int dim
         return indexModel;
     }
 
+    // Minimal
+    public IndexModel createPodsIndex(String indexName, Integer dimension, String environment, String podType) {
+        if (indexName == null || indexName.isEmpty()) {
```
rohanshah18, line 109: There's some repetition in each create methods which can be extracted out by creating a validateAndBuildCreatePodsIndexRequest(). I'm sure there's a better name so I'll leave it up to you but a similar approach for validating and building query request in data plane can be found here: 
https://github.com/pinecone-io/pinecone-java-client/blob/main/src/main/java/io/pinecone/commons/IndexInterface.java#L128


Note that the CreateIndexRequestSpecPod podSpec = new CreateIndexRequestSpecPod() can be done before validations and after each parameter is validated, we can keep setting the value.

Similarly, once podSpec object is all created and values are set, we can approach:
CreateIndexRequest createIndexRequest = new CreateIndexRequest();
and start setting the values. 

If the required parameter is missing, we would like to throw an exception.
If the optional parameter is missing, we dont want to set the value. 


Finally, your overloaded methods would look like this:
public IndexModel createPodsIndex(String indexName, Integer dimension, String environment, String podType) {
    return validateAndBuildCreatePodsIndex(indexName, dimension, environment, null, podType, null, null, null, null, null);
}

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -87,6 +91,7 @@ public IndexModel createServerlessIndex(String indexName, String metric, int dim
         CreateIndexRequestSpec createServerlessIndexRequestSpec = new CreateIndexRequestSpec().serverless(serverlessSpec);
 
         IndexModel indexModel = null;
+
```
rohanshah18, line 94: Should have left this comment for previous PR, sorry but For line#72:
1. if we are checking for null of line#69, then line#72 is redundant check, we don't need it.

Similarly, if line#54 is checking for null then line#57's null check is not needed

build.gradle
```diff
@@ -53,6 +53,7 @@ dependencies {
     implementation 'io.gsonfire:gson-fire:1.8.5'
     implementation 'org.openapitools:jackson-databind-nullable:0.2.6'
     implementation 'com.google.protobuf:protobuf-java:3.25.2'
+    implementation 'org.junit.jupiter:junit-jupiter-api:5.8.0'
```
rohanshah18, line 56: if assertEquals was the only use case, we can remove this

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Another option is just validating the required parameters by creating a new method:
private void validateIndexCreation(String indexName, Integer dimension, String environment, String podType) {
    if (indexName == null || indexName.isEmpty()) {
        throw new PineconeValidationException("indexName cannot be null or empty");
    }
    if (dimension == null) {
        throw new PineconeValidationException("Dimension cannot be null");
    }
    if (dimension < 1) {
        throw new PineconeValidationException("Dimension must be greater than 0. See limits for more info: https://docs.pinecone.io/reference/limits");
    }
    if (environment == null || environment.isEmpty()) {
        throw new PineconeValidationException("Environment cannot be null or empty");
    }
    if (podType == null || podType.isEmpty()) {
        throw new PineconeValidationException("Pod type cannot be null or empty");
    }
}

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -99,6 +104,301 @@ public IndexModel createServerlessIndex(String indexName, String metric, int dim
         return indexModel;
     }
 
+    // Minimal
+    public IndexModel createPodsIndex(String indexName, Integer dimension, String environment, String podType) {
```
rohanshah18, line 108: We have redundant checks in the method. This method is checking for null and emptiness, and at line#129, we are calling createPodsIndex(indexName, dimension, environment, null, podType, null, null, null, null, null) which further checks for the null and emptiness.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -99,6 +104,301 @@ public IndexModel createServerlessIndex(String indexName, String metric, int dim
...
+
+        if (environment == null || environment.isEmpty()) {
+            throw new PineconeValidationException("Environment cannot be null or empty");
+        }
+
+        if (podType == null || podType.isEmpty()) {
+            throw new PineconeValidationException("Pod type cannot be null or empty");
+        }
+
+        return createPodsIndex(indexName, dimension, environment, null, podType, null, null, pods, null, null);
```
rohanshah18, line 247: podType is the fourth parameter, so we need to set it at the right place

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -99,6 +104,301 @@ public IndexModel createServerlessIndex(String indexName, String metric, int dim
...
+
+        if (environment == null || environment.isEmpty()) {
+            throw new PineconeValidationException("Environment cannot be null or empty");
+        }
+
+        if (podType == null || podType.isEmpty()) {
+            throw new PineconeValidationException("Pod type cannot be null or empty");
+        }
+
+        return createPodsIndex(indexName, dimension, environment, null, podType, replicas, shards, null, null, null);
```
rohanshah18, line 302: Same as before, podtype is the fourth parameter

rohanshah18 (approved):
Thanks for making these changes. I have left a few comments especially on redundant checks and how we can remove the code redundancy.

build.gradle
aulorbe: But then how do we use `assertEquals` in the `Pinecone` class?

aulorbe:
> Thanks for making these changes. I have left a few comments especially on redundant checks and how we can remove the code redundancy.

Thank you! Will incorporate new validation method now!

src/main/java/io/pinecone/clients/Pinecone.java
aulorbe: Gotcha, I've put it back down at the bottom

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -99,6 +104,301 @@ public IndexModel createServerlessIndex(String indexName, String metric, int dim
...
+
+        if (dimension < 1) {
+            throw new PineconeValidationException("Dimension must be greater than 0. See limits for more info: https://docs.pinecone.io/reference/limits");
+        }
+
+        if (environment == null || environment.isEmpty()) {
+            throw new PineconeValidationException("Environment cannot be null or empty");
+        }
+
+        if (replicas != null && shards != null && pods != null) {
```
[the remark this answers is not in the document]
aulorbe: Done!

build.gradle
aulorbe: Done!

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -63,11 +63,10 @@ public void configureIndexWithInvalidIndexName() {
     @Test
     public void configureIndexExceedingQuota() {
         try {
-            controlPlaneClient.configureIndex(indexName, 400);
-
+            controlPlaneClient.configureIndex(indexName, 600);
```
austin-denoble, line 66: I think this was due to @haruska bumping our project quota, at least it seems like that's the case.

austin-denoble (approved):
Thanks for iterating, covering all this testing, and dealing with testing CI.

PR #99, opened 2024-04-09
Refactor data plane tests to use `TestResourcesManager`, clean up lengthy `Thead.sleep()` calls, general clean up

austin-denoble:
## Problem
The Java integration tests have traditionally been very flakey, and have taken a long time to run. For example, if freshness is lagged or delayed we can end up waiting nearly 30 minutes for a single integration test run to finish due to `assertWithRetry` being used with a backoff of 4.

Previously, I added a new `TestIndexResourcesManager` class to handle managing shared index resources across all tests over the `IndexManager`: https://github.com/pinecone-io/pinecone-java-client/pull/89. I initially only implemented the manager in control plane tests. It also needs to be added to data plane tests.

## Solution
- Rename `TestIndexResourcesManager` to `TestResourcesManager`.
- Rename `IndexManager` to `TestUtilities` and remove unneeded functions for managing indexes.
- Clean up some of the duplicative function in `TestResourcesManager`. Opt for the naming paradigm of `getOrCreateFoo` for the various functions that allow getting or creating indexes or collections.
- Refactor all of the integration tests in `/integration/dataPlane/` to use the `TestResourcesManager` rather than the helper functions in `TestUtilities`.
- Remove all of the `Thread.sleep(90000)` etc calls that were sprinkled all over the data plane tests. I don't think we need to actually wait this long, it was incredibly inefficient and if freshness is degraded they would still usually fail.
- For places where we're calling `assertWithRetry()` for retrying commands in tests when things fail, I swapped from using a backoff of 4 to 3. We should look at refactoring this function to be a bit more flexible, but for now we should avoid backoffs of > 3 as they cause an enormous amount of wait time: 6000ms, 24000ms, 96000ms, 384000ms. We should never wait this long within a test it would be preferable to let things fail earlier and just re-run.
- I did a lot of various refactoring across the data plane tests. Ultimately I don't think I've changed anything we're actually testing, but I've tried to make things less brittle and more readable.

## Type of Change
- [X] Infrastructure change (CI configs, integration tests etc)

## Test Plan
Run the integration tests suite locally or validate we see green in CI.

`gradle integrationTest`

### DIFF
--- a/src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
+++ b/src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
@@ -39,15 +39,15 @@ public class BuildUpsertRequest {
     }
 
     public static ArrayList<Long> generateSparseIndicesByDimension(int dimension) {
-        ArrayList<Long> values = new ArrayList<>();
         Random random = new Random();
+        ArrayList<Long> indices = new ArrayList<>();
         long maxUnsignedInt = (1L << 32) - 1;
 
         for (int i = 0; i < dimension; i++) {
-            values.add(random.nextLong() & maxUnsignedInt);
+            indices.add(random.nextLong() & maxUnsignedInt);
         }
 
-        return values;
+        return indices;
     }
 
     public static Struct generateMetadataStruct() {
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -27,8 +27,8 @@ public class TestIndexResourcesManager {
             ? "us-east4-gcp"
             : System.getenv("PINECONE_ENVIRONMENT");
     private static final String metric = System.getenv("METRIC") == null
-            ? IndexMetric.COSINE.toString()
-            : System.getenv("METRIC");
+            ? IndexMetric.DOTPRODUCT.toString()
+            : IndexMetric.valueOf(System.getenv("METRIC")).toString();
     private static final String cloud = System.getenv("CLOUD") == null
             ? ServerlessSpec.CloudEnum.AWS.toString()
             : System.getenv("CLOUD");
@@ -145,7 +145,7 @@ public class TestIndexResourcesManager {
 
         // Explicitly wait after ready to avoid the "no healthy upstream" issue
         Thread.sleep(30000);
-
+        
         // Seed default vector IDs into default namespace
         seedIndex(vectorIdsForDefaultNamespace, indexName, defaultNamespace);
 
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -84,6 +84,8 @@ public class TestIndexResourcesManager {
         return dimension;
     }
 
+    public String getMetric() { return metric; };
+
     public String getEnvironment() {
         return environment;
     }
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -142,10 +142,10 @@ public class TestIndexResourcesManager {
 
         String indexName = RandomStringBuilder.build("pod-index", 8);
 
-        podIndexModel = pineconeClient.createPodsIndex(indexName, dimension, environment, "p1.x1");
+        podIndexModel = pineconeClient.createPodsIndex(indexName, dimension, environment, "p1.x1", metric);
         waitUntilIndexIsReady(pineconeClient, indexName);
-        
-        // Explicitly wait after ready to avoid the "no healthy upstream" issue
+
+        // Additional sleep after index marked as ready to avoid "no healthy upstream" error
         Thread.sleep(30000);
         
         // Seed default vector IDs into default namespace
@@ -161,7 +161,7 @@ public class TestIndexResourcesManager {
         }
 
         String indexName = RandomStringBuilder.build("serverless-index", 8);
-
+        
         serverlessIndexModel = pineconeClient.createServerlessIndex(indexName, metric, dimension, cloud,
                 region);
         waitUntilIndexIsReady(pineconeClient, indexName);
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -4,6 +4,7 @@ import io.pinecone.clients.AsyncIndex;
 import io.pinecone.clients.Index;
 import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeException;
+import io.pinecone.proto.DescribeIndexStatsResponse;
 import org.openapitools.client.model.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -4,7 +4,6 @@ import io.pinecone.clients.AsyncIndex;
 import io.pinecone.clients.Index;
 import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeException;
-import io.pinecone.proto.DescribeIndexStatsResponse;
 import org.openapitools.client.model.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -13,11 +12,11 @@ import java.util.Arrays;
 import java.util.List;
 
 import static io.pinecone.helpers.BuildUpsertRequest.buildRequiredUpsertRequestByDimension;
-import static io.pinecone.helpers.IndexManager.*;
+import static io.pinecone.helpers.TestUtilities.*;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class TestIndexResourcesManager {
-    private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
+    private static final Logger logger = LoggerFactory.getLogger(TestUtilities.class);
     private static TestIndexResourcesManager instance;
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
     private final int dimension = System.getenv("DIMENSION") == null
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -142,7 +142,7 @@ public class TestIndexResourcesManager {
 
         podIndexModel = pineconeClient.createPodsIndex(indexName, dimension, environment, "p1.x1");
         waitUntilIndexIsReady(pineconeClient, indexName);
-
+        
         // Explicitly wait after ready to avoid the "no healthy upstream" issue
         Thread.sleep(30000);
 
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -145,7 +145,7 @@ public class TestIndexResourcesManager {
         
         // Explicitly wait after ready to avoid the "no healthy upstream" issue
         Thread.sleep(30000);
-
+        
         // Seed default vector IDs into default namespace
         seedIndex(vectorIdsForDefaultNamespace, indexName, defaultNamespace);
 
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -145,7 +145,7 @@ public class TestIndexResourcesManager {
 
         // Explicitly wait after ready to avoid the "no healthy upstream" issue
         Thread.sleep(30000);
-        
+
         // Seed default vector IDs into default namespace
         seedIndex(vectorIdsForDefaultNamespace, indexName, defaultNamespace);
 
--- a/src/integration/java/io/pinecone/helpers/TestUtilities.java
+++ b/src/integration/java/io/pinecone/helpers/TestUtilities.java
@@ -1,24 +1,28 @@
 package io.pinecone.helpers;
 
+import io.pinecone.clients.Index;
 import io.pinecone.clients.Pinecone;
-import io.pinecone.exceptions.PineconeException;
+import io.pinecone.proto.DescribeIndexStatsResponse;
+import io.pinecone.proto.NamespaceSummary;
 import org.openapitools.client.model.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.*;
+
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.fail;
 
-public class IndexManager {
-    private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
+public class TestUtilities {
+    private static final Logger logger = LoggerFactory.getLogger(TestUtilities.class);
 
-    public static IndexModel waitUntilIndexIsReady(Pinecone pinecone, String indexName, Integer totalMsToWait) throws InterruptedException {
-        IndexModel index = pinecone.describeIndex(indexName);
+    public static IndexModel waitUntilIndexIsReady(Pinecone pineconeClient, String indexName, Integer totalMsToWait) throws InterruptedException {
+        IndexModel index = pineconeClient.describeIndex(indexName);
         int waitedTimeMs = 0;
         int intervalMs = 2000;
 
         while (index.getStatus().getState() != IndexModelStatus.StateEnum.READY) {
-            index = pinecone.describeIndex(indexName);
+            index = pineconeClient.describeIndex(indexName);
             if (waitedTimeMs >= totalMsToWait) {
                 logger.info("WARNING: Index " + indexName + " not ready after " + waitedTimeMs + "ms");
                 break;
@@ -35,12 +39,12 @@ public class IndexManager {
         return index;
     }
 
-    public static IndexModel waitUntilIndexIsReady(Pinecone pinecone, String indexName) throws InterruptedException {
-        return waitUntilIndexIsReady(pinecone, indexName, 200000);
+    public static IndexModel waitUntilIndexIsReady(Pinecone pineconeClient, String indexName) throws InterruptedException {
+        return waitUntilIndexIsReady(pineconeClient, indexName, 200000);
     }
 
-    public static CollectionModel createCollection(Pinecone pinecone, String collectionName, String indexName, boolean waitUntilReady) throws InterruptedException {
-        CollectionModel collection = pinecone.createCollection(collectionName, indexName);
+    public static CollectionModel createCollection(Pinecone pineconeClient, String collectionName, String indexName, boolean waitUntilReady) throws InterruptedException {
+        CollectionModel collection = pineconeClient.createCollection(collectionName, indexName);
 
         assertEquals(collection.getStatus(), CollectionModel.StatusEnum.INITIALIZING);
 
@@ -53,7 +57,7 @@ public class IndexManager {
                         "milliseconds...");
                 Thread.sleep(5000);
                 timeWaited += 5000;
-                collection = pinecone.describeCollection(collectionName);
+                collection = pineconeClient.describeCollection(collectionName);
                 collectionReady = collection.getStatus();
             }
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -28,7 +28,7 @@ public class CollectionTest {
     private static final Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
     private static final Logger logger = LoggerFactory.getLogger(CollectionTest.class);
     private static final ArrayList<String> indexesToCleanUp = new ArrayList<>();
-    private static final String indexMetric = IndexMetric.COSINE.toString();
+    private static final String indexMetric = indexManager.getMetric();
     private static final List<String> upsertIds = indexManager.getVectorIdsFromDefaultNamespace();
     private static final String environment = indexManager.getEnvironment();
     private static final int dimension = indexManager.getDimension();
@@ -113,10 +113,17 @@ public class CollectionTest {
 
     @Test
     public void testCreateIndexFromCollectionWithDiffMetric() throws InterruptedException {
-        // Note collection == 1 index
-        String metricForNewIndex = IndexMetric.DOTPRODUCT.toString();
+        // Use a different metric than the source index
+        String[] metrics = {IndexMetric.COSINE.toString(), IndexMetric.EUCLIDEAN.toString(), IndexMetric.DOTPRODUCT.toString()};
+        String targetMetric = IndexMetric.COSINE.toString();
+        for (String metric : metrics) {
+            if (!metric.equals(indexMetric)) {
+                targetMetric = metric;
+            }
+        }
+        
         String newIndexName = RandomStringBuilder.build("from-coll-with-diff-metric", 5);
-        pineconeClient.createPodsIndex(newIndexName, dimension, environment, "p1.x1", metricForNewIndex, collectionName);
+        pineconeClient.createPodsIndex(newIndexName, dimension, environment, "p1.x1", targetMetric, collectionName);
         indexesToCleanUp.add(newIndexName);
     }
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -5,7 +5,6 @@ import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeException;
 import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.helpers.RandomStringBuilder;
-import io.pinecone.proto.DescribeIndexStatsResponse;
 import io.pinecone.proto.FetchResponse;
 import io.pinecone.helpers.TestIndexResourcesManager;
 
@@ -20,7 +19,7 @@ import org.slf4j.LoggerFactory;
 import java.util.*;
 
 import static io.pinecone.helpers.AssertRetry.assertWithRetry;
-import static io.pinecone.helpers.IndexManager.*;
+import static io.pinecone.helpers.TestUtilities.*;
 import static org.junit.jupiter.api.Assertions.*;
 
 public class CollectionTest {
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -121,7 +121,7 @@ public class CollectionTest {
                 targetMetric = metric;
             }
         }
-        
+
         String newIndexName = RandomStringBuilder.build("from-coll-with-diff-metric", 5);
         pineconeClient.createPodsIndex(newIndexName, dimension, environment, "p1.x1", targetMetric, collectionName);
         indexesToCleanUp.add(newIndexName);
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -76,7 +76,6 @@ public class CollectionTest {
 
         assertEquals(collection.getStatus(), CollectionModel.StatusEnum.READY);
         assertEquals(collection.getDimension(), dimension);
-        assertEquals(collection.getVectorCount(), 4);
         assertNotEquals(collection.getVectorCount(), null);
         assertTrue(collection.getSize() > 0);
 
@@ -122,7 +121,7 @@ public class CollectionTest {
         // Note collection == 1 index
         String metricForNewIndex = IndexMetric.DOTPRODUCT.toString();
         String newIndexName = RandomStringBuilder.build("from-coll-with-diff-metric", 5);
-        createNewIndex(pineconeClient, newIndexName, dimension, metricForNewIndex, collectionName, true);
+        pineconeClient.createPodsIndex(newIndexName, dimension, environment, "p1.x1", metricForNewIndex, collectionName);
         indexesToCleanUp.add(newIndexName);
     }
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -97,7 +97,7 @@ public class CollectionTest {
         Thread.sleep(30000);
 
         // If the index is ready, validate contents
-        if (indexDescription.getStatus().getReady()) {
+        if (indexDescription.getStatus().getState() == IndexModelStatus.StateEnum.READY) {
             // Set up new index data plane connection
             Index indexClient = pineconeClient.getIndexConnection(newIndexName);
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -38,9 +38,9 @@ public class CollectionTest {
 
     @BeforeAll
     public static void setUp() throws InterruptedException {
-        indexName = indexManager.getPodIndexName();
-        collectionName = indexManager.getCollectionName();
-        collection = indexManager.getCollectionModel();
+        indexName = indexManager.getOrCreatePodIndex();
+        collectionName = indexManager.getOrCreateCollection();
+        collection = indexManager.getOrCreateCollectionModel();
         namespace = indexManager.getDefaultNamespace();
     }
 
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -47,9 +47,6 @@ public class CollectionTest {
 
     @AfterAll
     public static void cleanUp() throws InterruptedException {
-        // wait for things to settle before cleanup...
-        Thread.sleep(2500);
-
         // Clean up indexes
         for (String index : indexesToCleanUp) {
             pineconeClient.deleteIndex(index);
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -102,11 +102,6 @@ public class CollectionTest {
             Index indexClient = pineconeClient.getIndexConnection(newIndexName);
 
             assertWithRetry(() -> {
-                DescribeIndexStatsResponse describeResponse = indexClient.describeIndexStats();
-
-                // Verify stats reflect the vectors in the collection
-                assertEquals(describeResponse.getTotalVectorCount(), 4);
-
                 // Verify the vectors from the collection -> new index can be fetched
                 FetchResponse fetchedVectors = indexClient.fetch(upsertIds, namespace);
                 for (String key : upsertIds) {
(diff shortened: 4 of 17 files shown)


src/integration/java/io/pinecone/helpers/BuildUpsertRequest.java
```diff
@@ -39,15 +39,15 @@ public static ArrayList<Float> generateVectorValuesByDimension(int dimension) {
     }
 
     public static ArrayList<Long> generateSparseIndicesByDimension(int dimension) {
-        ArrayList<Long> values = new ArrayList<>();
         Random random = new Random();
+        ArrayList<Long> indices = new ArrayList<>();
```
austin-denoble, line 43: Renamed the `values` field in here, was a little confusing given this function generates indices.

src/integration/java/io/pinecone/helpers/TestUtilities.java
```diff
@@ -0,0 +1,71 @@
...
+import org.openapitools.client.model.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class TestUtilities {
```
austin-denoble, line 16: This is a rename of `IndexManager` -> `TestUtilities`. @aulorbe had mentioned it was a bit confusing getting started with some of these helpers, and since the index management has been moved to `TestIndexResourcesManager`, we can rename this class to be a bit more informative.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -20,15 +19,15 @@
...
+import static io.pinecone.helpers.TestUtilities.*;
 import static org.junit.jupiter.api.Assertions.*;
 
 public class CollectionTest {
     private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
     private static final Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
     private static final Logger logger = LoggerFactory.getLogger(CollectionTest.class);
     private static final ArrayList<String> indexesToCleanUp = new ArrayList<>();
-    private static final String indexMetric = IndexMetric.COSINE.toString();
+    private static final String indexMetric = indexManager.getMetric();
```
austin-denoble, line 30: Swapped to using the internal metric value from `TestIndexResourcesManager`.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -77,7 +73,6 @@ public void testIndexFromCollectionHappyPath() throws InterruptedException {
 
         assertEquals(collection.getStatus(), CollectionModel.StatusEnum.READY);
         assertEquals(collection.getDimension(), dimension);
-        assertEquals(collection.getVectorCount(), 3);
```
austin-denoble, line 80: Since we're now using a shared index across tests and there's no guarantee around how many vectors may be in the index by the time this test runs, I removed this check for now. 

We validate the actual vectors in the collection using `upsertIds` below which feels like it should be sufficient for validating the contents.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -120,10 +110,17 @@ public void testIndexFromCollectionHappyPath() throws InterruptedException {
...
+        String[] metrics = {IndexMetric.COSINE.toString(), IndexMetric.EUCLIDEAN.toString(), IndexMetric.DOTPRODUCT.toString()};
+        String targetMetric = IndexMetric.COSINE.toString();
+        for (String metric : metrics) {
+            if (!metric.equals(indexMetric)) {
+                targetMetric = metric;
+            }
+        }
+
         String newIndexName = RandomStringBuilder.build("from-coll-with-diff-metric", 5);
-        createNewIndex(pineconeClient, newIndexName, dimension, metricForNewIndex, collectionName, true);
```
austin-denoble, line 126: We weren't passing `podType` previously.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -120,10 +110,17 @@ public void testIndexFromCollectionHappyPath() throws InterruptedException {
 
     @Test
     public void testCreateIndexFromCollectionWithDiffMetric() throws InterruptedException {
-        // Note collection == 1 index
-        String metricForNewIndex = IndexMetric.DOTPRODUCT.toString();
+        // Use a different metric than the source index
+        String[] metrics = {IndexMetric.COSINE.toString(), IndexMetric.EUCLIDEAN.toString(), IndexMetric.DOTPRODUCT.toString()};
```
austin-denoble, line 114: Added this back in since we're using whatever metric the index was created with in `TestIndexResourcesManager`.

src/integration/java/io/pinecone/integration/controlPlane/pod/ConfigureIndexTest.java
```diff
@@ -127,7 +127,5 @@ public void sizeIncrease() throws InterruptedException {
             assertNotNull(podSpec);
             assertEquals(podSpec.getPodType(), "p1.x2");
         });
-
-        waitUntilIndexStateIsReady(indexName);
```
austin-denoble, line 131: We do this in an `@AfterEach` call so it's not really necessary to call it here.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -6,74 +6,71 @@
...
-import static io.pinecone.helpers.IndexManager.*;
 import static org.junit.jupiter.api.Assertions.*;
 
-public class UpdateFetchAndQueryPodTest {
+public class  UpdateFetchAndQueryPodTest {
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
     private static Index index;
     private static AsyncIndex asyncIndex;
-    private static String namespace;
+    private static final String namespace = RandomStringBuilder.build("ns", 8);
```
austin-denoble, line 33: I made this change in most of our data plane test files. We used to create a new namespace sometimes for each test. Rather than going that route it felt easier to reason about and debug things when using the same shared namespace across each test file.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -6,74 +6,71 @@
...
     public static void setUp() throws IOException, InterruptedException {
-        String apiKey = System.getenv("PINECONE_API_KEY");
-        String indexType = IndexModelSpec.SERIALIZED_NAME_POD;
-        Pinecone pinecone = new Pinecone.Builder(apiKey).build();
+        Pinecone pinecone = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
 
-        String indexName = findIndexWithDimensionAndType(pinecone, dimension, indexType);
-        if (indexName.isEmpty()) indexName = createNewIndex(pinecone, dimension, indexType, true);
+        String indexName = indexManager.getPodIndexName();
+        dimension = indexManager.getDimension();
```
austin-denoble, line 44: In all of these test files, the swapping from `IndexManager` to `TestIndexResourcesManager` should look something like this.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -6,74 +6,71 @@
...
         upsertIds = getIdsList(numOfVectors);
-        List<List<Long>> sparseIndicesList = getSparseIndicesList(numOfSparseVectors, dimension);
-        sparseValuesList = getValuesListLists(numOfSparseVectors, dimension);
-
-        // upsert vectors with required + optional parameters
-        int sparseVectorCount = 0;
-        for (int i = sparseVectorCount; i < numOfSparseVectors; i++) {
-            index.upsert(upsertIds.get(i),
+        sparseUpsertIds = getIdsList(numOfSparseVectors);
+        ArrayList<Long> sparseIndices = generateSparseIndicesByDimension(dimension);
```
austin-denoble, line 53: Previously we were for-looping and calling `upsert` each time. Rather than doing that, I thought it'd be better to construct a `List<VectorWithUnsignedIndices` and build that, then call `upsert` once. I've done that for a bunch of the data plane tests.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -121,49 +115,50 @@ public void updateAllParamsFetchAndQuerySyncTest() throws InterruptedException {
...
-            ScoredVectorWithUnsignedIndices scoredVectorV1 = queryResponse.getMatches(0);
             // Verify the correct vector id was updated
+            assertNotNull(scoredVectorV1);
             assertEquals(scoredVectorV1.getId(), idToUpdate);
 
             // Verify the updated values
-            List<Float> valuesList = new ArrayList<>(scoredVectorV1.getValuesList());
-            Collections.sort(valuesList);
-            Collections.sort(valuesToUpdate);
+            List<Float> valuesList = scoredVectorV1.getValuesList();
```
austin-denoble, line 132: I don't know why these were being sorted, and a lot of this was just really confusing to me so I tried to simplify things.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -181,34 +176,30 @@ public void updateNullSparseIndicesNotNullSparseValuesSyncTest() {
     }
 
     @Test
-    public void queryWithFilersSyncTest() {
+    public void queryWithFiltersSyncTest() throws InterruptedException {
         String fieldToQuery = metadataFields[0];
         String valueToQuery = createAndGetMetadataMap().get(fieldToQuery).get(0);
 
-        try {
```
austin-denoble, line 188: Removed unneeded try / catch from this.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -248,79 +239,87 @@ public void updateAllParamsFetchAndQueryFutureTest() throws InterruptedException
...
 
-            ScoredVectorWithUnsignedIndices scoredVectorV1 = queryResponse.getMatches(0);
+            List<ScoredVectorWithUnsignedIndices> queryResults = queryResponse.getMatchesList();
+            ScoredVectorWithUnsignedIndices scoredVectorV1 = null;
+
+            for (ScoredVectorWithUnsignedIndices vector : queryResults) {
+                if (idToUpdate.equals(vector.getId())) {
+                    scoredVectorV1 = vector;
+                }
+            }
```
austin-denoble, line 249: The way this was being done originally confused me, and also seemed prone to possible flakiness, so I updated these in several test files.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -248,79 +239,87 @@ public void updateAllParamsFetchAndQueryFutureTest() throws InterruptedException
...
             fail("Expected to throw statusRuntimeException");
         } catch (ExecutionException executionException) {
-            assert (executionException.toString().contains("Vector dimension 1 does not match the dimension of the index 3"));
+            assert (executionException.toString().contains("Vector dimension 1 does not match the dimension of the index " + dimension));
         }
     }
 
     @Test
-    public void queryWithFilersFutureTest() throws ExecutionException, InterruptedException {
+    public void queryWithFiltersFutureTest() throws ExecutionException, InterruptedException {
```
austin-denoble, line 285: Spelling fix

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
```diff
@@ -183,34 +184,31 @@ public void updateNullSparseIndicesNotNullSparseValuesSyncTest() {
     }
 
     @Test
-    public void queryWithFilersSyncTest() {
+    public void queryWithFiltersSyncTest() throws InterruptedException {
         String fieldToQuery = metadataFields[0];
         String valueToQuery = createAndGetMetadataMap().get(fieldToQuery).get(0);
 
-        try {
```
austin-denoble, line 190: Remove the unneeded try / catch here.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
```diff
@@ -124,48 +119,54 @@ public void updateAllParamsFetchAndQuerySyncTest() throws InterruptedException {
...
-            Collections.sort(expectedSparseValues);
-
-            assertEquals(expectedSparseValues, actualSparseValues);
-        });
+            List<Float> expectedSparseValues = sparseValuesList;
+            List<Float> actualSparseValues = scoredVectorV1.getSparseValuesWithUnsignedIndices().getValuesList();
+            // TODO - See if this is an issue, sparseValues seem unsorted after upsert -> update
+            for (Float value : expectedSparseValues) {
+                assertTrue(actualSparseValues.contains(value));
+            }
```
austin-denoble, line 148: This was the only test I saw this issue in (I'm guessing this is why we had those `Collections.sort()` calls all over before?).

Basically it seemed like the sparseValues could be sorted differently after upserting a vector. We would need to troubleshoot why, for now this validates the values.

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeletePodTest.java
```diff
@@ -6,124 +6,106 @@
...
         assertWithRetry(() -> {
             // call describeIndexStats to get updated vector count
             DescribeIndexStatsResponse describeIndexStatsResponse = index.describeIndexStats();
-
+            System.out.println("ASSERT WITH RETRY 1");
             // verify the updated vector count
-            assertEquals(describeIndexStatsResponse.getNamespacesMap().get(namespace).getVectorCount(), actualVectorCount);
-        });
-
-        // Delete 1 vector
```
austin-denoble, line 68: I'm not sure why we were deleting one vector, then waiting, then deleting the others. It was confusing to read and work around. I batched things into one `index.delete()` request down below. Faster and easier to reason about, and it still tests the same code paths.

src/integration/java/io/pinecone/helpers/TestUtilities.java
aulorbe: ty!

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
aulorbe: Worth consideration.... what if we renamed `TestIndexResourcesManager` to just `TestResourcesManager`? For some reason the `Index` bit seems superfluous (like, it's covered under the word "resources," you know?) to me

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
aulorbe: awesomeness

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -12,11 +11,11 @@
 import java.util.List;
 
 import static io.pinecone.helpers.BuildUpsertRequest.buildRequiredUpsertRequestByDimension;
-import static io.pinecone.helpers.IndexManager.*;
+import static io.pinecone.helpers.TestUtilities.*;
```
aulorbe, line 14: 🤙 sweet

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -26,8 +25,8 @@ public class TestIndexResourcesManager {
             ? "us-east4-gcp"
             : System.getenv("PINECONE_ENVIRONMENT");
     private static final String metric = System.getenv("METRIC") == null
-            ? IndexMetric.COSINE.toString()
-            : System.getenv("METRIC");
+            ? IndexMetric.DOTPRODUCT.toString()
```
aulorbe, line 28: Remind me why we want this to be dotproduct again?

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -108,14 +109,15 @@ private String createOrGetPodIndex() throws InterruptedException, PineconeExcept
...
+        podIndexModel = pineconeClient.createPodsIndex(indexName, dimension, environment, "p1.x1", metric);
         waitUntilIndexIsReady(pineconeClient, indexName);
 
-        // Explicitly wait after ready to avoid the "no healthy upstream" issue
+        // Additional sleep after index marked as ready to avoid "no healthy upstream" error
         Thread.sleep(30000);
 
         // Seed data
-        seedIndex(podIndexVectorIds, indexName);
+        Index indexClient = pineconeClient.getIndexConnection(indexName);
```
aulorbe, line 119: W/[my latest PR](https://github.com/pinecone-io/pinecone-java-client/pull/115), I've introduced `getServerlessIndexConnection` and `getServerlessAsyncIndexConnection`. I think you should add a `getPodsIndexConnection` to make things more symmetrical? But not mandatory obvi.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -108,14 +109,15 @@ private String createOrGetPodIndex() throws InterruptedException, PineconeExcept
...
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(podIndexVectorIds, dimension), "");
```
aulorbe, line 120: The `""` w/my [latest PR](https://github.com/pinecone-io/pinecone-java-client/pull/115)will change to `defaultNamespace`

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -99,16 +94,11 @@ public void testIndexFromCollectionHappyPath() throws InterruptedException {
         Thread.sleep(30000);
 
         // If the index is ready, validate contents
-        if (indexDescription.getStatus().getReady()) {
+        if (indexDescription.getStatus().getState() == IndexModelStatus.StateEnum.READY) {
```
aulorbe, line 97: Nice

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -120,10 +110,17 @@ public void testIndexFromCollectionHappyPath() throws InterruptedException {
...
-        String metricForNewIndex = IndexMetric.DOTPRODUCT.toString();
+        // Use a different metric than the source index
+        String[] metrics = {IndexMetric.COSINE.toString(), IndexMetric.EUCLIDEAN.toString(), IndexMetric.DOTPRODUCT.toString()};
+        String targetMetric = IndexMetric.COSINE.toString();
+        for (String metric : metrics) {
+            if (!metric.equals(indexMetric)) {
+                targetMetric = metric;
+            }
+        }
+
```
aulorbe, line 121: I still don't get  this loop. We're saying if the `metric` does not equal the `indexMetric`, update `targetMetric` to be `metric`? For me, there are too many similarly-named variables here. Can we maybe switch `metric` to just `m`, and/or switch `indexMetric` to be something like `sourceIndexMetric`?

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
```diff
@@ -33,7 +33,7 @@ public void describeAndListIndex() {
         assertNotNull(indexModel);
         assertEquals(indexDimension, indexModel.getDimension());
         assertEquals(indexName, indexModel.getName());
-        assertEquals(IndexMetric.COSINE, indexModel.getMetric());
+        assertEquals(IndexMetric.DOTPRODUCT, indexModel.getMetric());
```
aulorbe, line 36: Same Q as in the other file -- why dotproduct here again?

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
aulorbe: (It'll need to be come `defaultNamespace`

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -91,10 +88,10 @@ public void updateAllParamsFetchAndQuerySyncTest() throws InterruptedException {
             for (String key : upsertIds) {
                 assert (fetchResponse.containsVectors(key));
             }
-        });
+        }, 3);
 
-        String idToUpdate = upsertIds.get(0);
-        List<Float> valuesToUpdate = Arrays.asList(201F, 202F, 203F);
+        String idToUpdate = sparseUpsertIds.get(0);
```
aulorbe, line 93: Something that @ssmith-pc [pointed out in my latest PR](https://github.com/pinecone-io/pinecone-java-client/pull/115#discussion_r1571581330) is that it's hard to understand quickly when looking at a test when values are like `...get(0)`. It'd be more obvious to readers if this value was just hardcoded. Is it possible to do that here?

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -121,49 +115,50 @@ public void updateAllParamsFetchAndQuerySyncTest() throws InterruptedException {
...
-            List<Float> expectedSparseValues = new ArrayList<>(scoredVectorV1.getSparseValuesWithUnsignedIndices().getValuesList());
-            List<Float> actualSparseValues = new ArrayList<>(sparseValuesList.get(0));
-            Collections.sort(actualSparseValues);
-            Collections.sort(expectedSparseValues);
-
+            List<Float> expectedSparseValues = sparseValuesList;
+            List<Float> actualSparseValues = scoredVectorV1.getSparseValuesWithUnsignedIndices().getValuesList();
             assertEquals(expectedSparseValues, actualSparseValues);
-        });
+        }, 3);
```
aulorbe, line 142: Is this asserting that both `expectedSparseValues` and `actualSparseValues` are equal to `3`?

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
```diff
@@ -248,79 +239,87 @@ public void updateAllParamsFetchAndQueryFutureTest() throws InterruptedException
...
             assertEquals(scoredVectorV1.getId(), idToUpdate);
 
             // Verify the updated values
             List<Float> valuesList = new ArrayList<>(scoredVectorV1.getValuesList());
-            Collections.sort(valuesList);
-            Collections.sort(valuesToUpdate);
             assertEquals(valuesToUpdate, valuesList);
 
-            // Verify the updated metadata
+            // Verify the metadata was updated
```
aulorbe, line 259: 👍

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
```diff
@@ -6,76 +6,74 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.fail;
 
 public class UpdateFetchAndQueryServerlessTest {
 
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
     private static Index index;
     private static AsyncIndex asyncIndex;
-    private static String namespace;
+    private static final String namespace = RandomStringBuilder.build("ns", 8);
```
aulorbe, line 37: This will change w/my PR

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
```diff
@@ -6,76 +6,74 @@
...
-        String apiKey = System.getenv("PINECONE_API_KEY");
-        String indexType = IndexModelSpec.SERIALIZED_NAME_SERVERLESS;
-        Pinecone pinecone = new Pinecone.Builder(apiKey).build();
+        Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
 
-        String indexName = findIndexWithDimensionAndType(pinecone, dimension, indexType);
-        if (indexName.isEmpty()) indexName = createNewIndex(pinecone, dimension, indexType, true);
-        index = pinecone.getIndexConnection(indexName);
-        asyncIndex = pinecone.getAsyncIndexConnection(indexName);
+        String indexName = indexManager.getServerlessIndexName();
```
aulorbe, line 47: Might want to note that this method `getServerlessIndexName()` also builds the index, if it's not built already (just b/c it's not obvious from the method name, as of right now)

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
```diff
@@ -6,76 +6,74 @@
...
+        dimension = indexManager.getDimension();
+        index = pineconeClient.getIndexConnection(indexName);
```
aulorbe, line 49: This will change w/my PR to `getServerlessIndexConnection`

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
```diff
@@ -6,76 +6,74 @@
...
+        asyncIndex = pineconeClient.getAsyncIndexConnection(indexName);
```
aulorbe, line 50: Same w/this, but to `Async`

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
aulorbe: Maybe we should just use `contains` so we're order-agnostic?

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryServerlessTest.java
```diff
@@ -23,25 +23,23 @@
...
-import static io.pinecone.helpers.IndexManager.*;
 import static org.junit.jupiter.api.Assertions.*;
 
 public class UpsertAndQueryServerlessTest {
-
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
     private static Index index;
     private static AsyncIndex asyncIndex;
+    private static int dimension;
+    private static final String namespace = RandomStringBuilder.build("ns", 8);
```
aulorbe, line 33: This will change w/my PR

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryServerlessTest.java
```diff
@@ -23,25 +23,23 @@
...
-
-        String indexName = findIndexWithDimensionAndType(pinecone, dimension, indexType);
-        if (indexName.isEmpty()) indexName = createNewIndex(pinecone, dimension, indexType, true);
-        index = pinecone.getIndexConnection(indexName);
-        asyncIndex = pinecone.getAsyncIndexConnection(indexName);
+        Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
+
+        String indexName = indexManager.getServerlessIndexName();
+        dimension = indexManager.getDimension();
+        index = pineconeClient.getIndexConnection(indexName);
```
aulorbe, line 41: This will change to `getServerlessIndexConnection` and same w/`Async`

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeletePodTest.java
```diff
@@ -6,124 +6,106 @@
...
 public class UpsertDescribeIndexStatsAndDeletePodTest {
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
+    private static Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
     private static Index index;
     private static AsyncIndex asyncIndex;
-    private static final int dimension = 3;
+    private static String indexName;
+    private static int dimension;
     private static final Struct nullFilterStruct = null;
+    private static final String namespace = RandomStringBuilder.build("ns", 8);
```
aulorbe, line 33: This will change to `defaultNamespace` w/my PR

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeleteServerlessTest.java
```diff
@@ -18,25 +19,24 @@
...
 
 public class UpsertDescribeIndexStatsAndDeleteServerlessTest {
-
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
+    private static Pinecone pineconeClient = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
     private static Index index;
     private static AsyncIndex asyncIndex;
+    private static String indexName;
+    private static int dimension;
+    private static final String namespace = RandomStringBuilder.build("ns", 8);
```
aulorbe, line 31: customNamespace

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeleteServerlessTest.java
```diff
@@ -18,25 +19,24 @@
...
-        int dimension = 3;
-        Pinecone pinecone = new Pinecone.Builder(apiKey).build();
-
-        String indexName = findIndexWithDimensionAndType(pinecone, dimension, indexType);
-        if (indexName.isEmpty()) indexName = createNewIndex(pinecone, dimension, indexType, true);
-        index = pinecone.getIndexConnection(indexName);
-        asyncIndex = pinecone.getAsyncIndexConnection(indexName);
+        indexName = indexManager.getServerlessIndexName();
+        dimension = indexManager.getDimension();
+        index = pineconeClient.getIndexConnection(indexName);
```
aulorbe, line 38: getServerlessIndexConnection, ""Async

aulorbe (approved):
No blockers, but there will be some changes to these tests once I merge my [latest PR](https://github.com/pinecone-io/pinecone-java-client/pull/115) in (I think I've noted all the instances that will have to be updated, but not 100% sure). 

LMK if you need help w/the merging. I'm happy to walk through essentially what my PR did. 

The TLDR is that we now have 2 diff namespaces in the shared index: `customNamespace` (`"example-namespace"`) and `defaultNamespace` (`""`); and we have 2 different sets of vector IDs that get upserted into each namespace (with 4 vector IDs in each set): `vectorIdsForCustomNamespace` and `vectorIdsForDefaultNamespace`.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
austin-denoble: Indexes must be dotproduct in order to allow querying sparse vectors: https://docs.pinecone.io/guides/data/understanding-hybrid-search#limitations

src/integration/java/io/pinecone/integration/controlPlane/pod/CreateDescribeListAndDeleteIndexTest.java
austin-denoble: Indexes must be dotproduct in order to allow querying sparse vectors: https://docs.pinecone.io/guides/data/understanding-hybrid-search#limitations

Most of our data plane tests use sparse vectors in some way so we need our indexes to be dotproduct.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
austin-denoble: The data plane tests all currently create their own namespaces which are used in the test file itself, so I don't think we necessarily need to change this yet.

If we reuse seeded data from `TestIndexResourcesManager` we'd have to make a lot of changes to how the data plane tests work as they currently seed their own data.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
austin-denoble: We could, but then it feels like you're using a hard-coded string that's being generated / defined in a different file (`TestResourcesManager`). Is that still not somewhat confusing as we then have to map the strings used in the tests back to their origin? I might be misunderstanding the feedback.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryPodTest.java
austin-denoble: The 3 that's added here is for setting the `backOff` to 3 for `assertWithRetry`. The `assertEquals` above is asserting what you're describing though.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
austin-denoble: Is that possible with two arrays, seeing if one contains the other regardless of ordering? I'm basically just checking each value using `contains` atm.

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
austin-denoble: Re-linking to my answer here: https://github.com/pinecone-io/pinecone-java-client/pull/99/files#r1575099673

The data plane test files all do a bit of self-isolation by creating their own namespaces, and seeding data. Feasibly we could change some of these to use the pre-seeded ids, but I think that may be good in a follow up. For now I just pulled the namespace generation up to the top level of the class so it's shared across all tests.

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndQueryServerlessTest.java
austin-denoble: Answer here: https://github.com/pinecone-io/pinecone-java-client/pull/99/files#r1576461309

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeletePodTest.java
austin-denoble: Answer here: https://github.com/pinecone-io/pinecone-java-client/pull/99/files#r1576461309

src/integration/java/io/pinecone/integration/dataPlane/UpsertDescribeIndexStatsAndDeleteServerlessTest.java
austin-denoble: https://github.com/pinecone-io/pinecone-java-client/pull/99/files#r1576461309

src/integration/java/io/pinecone/integration/dataPlane/UpdateFetchAndQueryServerlessTest.java
austin-denoble: This feedback made me reconsider a bunch of stuff in the `TestResourcesManager`. I removed some duplicative functions, and renamed things so that any call that either gets or creates an index or collection mentions that. A bit more verbose but I'm hoping it helps with clarity.

Also added docstrings to everything in there to try and make things a bit clearer. https://github.com/pinecone-io/pinecone-java-client/pull/99/commits/7d3aeb08ea705945c7e21714792fb27a516659ee

PR #106, opened 2024-04-10
Add new `queryByVectorId` and `queryByVector` functions to `IndexInterface`

austin-denoble:
## Problem
Because of the lack of optional arguments, we have a lot of overloaded functions in `IndexInterface` which are meant to make working with the data plane easier for users.

Currently, we're missing some variations of `queryByVectorId` which need to be added. We're also missing any functionality for `queryByVector` behavior which is a commonly used operations by users.

## Solution
- Add new function overloads to `IndexInterface` to support `queryByVectorId` minimal requests with namespace and/or `includeValues` and `includeMetadata`.
- Duplicate our `queryByVectorId` functions into new `queryByVector` functions.

All of these functions are going through the existing validation logic, so it should be safe to add these in calling the base `query` function. Unfortunately, there's not any unit tests for `Index` or `AsyncIndex` and those will need to be added in a follow up.

This is also a lot of overloads, I'm sure there is a better pattern for handling this from a UX perspective.

## Type of Change
- [X]  New feature (non-breaking change which adds functionality)

## Test Plan
CI tests, or running an example app locally.

### DIFF
--- a/src/main/java/io/pinecone/clients/AsyncIndex.java
+++ b/src/main/java/io/pinecone/clients/AsyncIndex.java
@@ -135,41 +135,50 @@ public class AsyncIndex implements IndexInterface<ListenableFuture<UpsertRespons
 
 
     @Override
-    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK, List<Float> vector,
-                                                                            String namespace, Struct filter,
+    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK,
+                                                                            List<Float> vector,
+                                                                            String namespace,
+                                                                            Struct filter,
                                                                             boolean includeValues,
                                                                             boolean includeMetadata) {
         return query(topK, vector, null, null, null, namespace, filter, includeValues, includeMetadata);
     }
 
     @Override
-    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK, List<Float> vector,
-                                                                            String namespace, Struct filter) {
+    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK,
+                                                                            List<Float> vector,
+                                                                            String namespace,
+                                                                            Struct filter) {
         return query(topK, vector, null, null, null, namespace, filter, false, false);
     }
 
     @Override
-    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK, List<Float> vector,
-                                                                            String namespace, boolean includeValues,
+    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK,
+                                                                            List<Float> vector,
+                                                                            String namespace,
+                                                                            boolean includeValues,
                                                                             boolean includeMetadata) {
         return query(topK, vector, null, null, null, namespace, null, includeValues, includeMetadata);
     }
 
     @Override
-    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK, List<Float> vector,
+    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK,
+                                                                            List<Float> vector,
                                                                             String namespace) {
         return query(topK, vector, null, null, null, namespace, null, false, false);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/io/pinecone/clients/Index.java
```diff
@@ -98,19 +98,63 @@ public QueryResponseWithUnsignedIndices queryByVectorId(int topK,
...
+    }
+
     @Override
     public QueryResponseWithUnsignedIndices queryByVectorId(int topK,
                                                             String id) {
         return query(topK, null, null, null, id, null, null, false, false);
     }
 
+    @Override
+    public QueryResponseWithUnsignedIndices queryByVector(int topK, List<Float> vector, String namespace, Struct filter, boolean includeValues, boolean includeMetadata) {
```
rohanshah18, line 129: nit: if we can have each parameter on new line just like other methods have it and the asyncIndex does for the same method, it'll be easy to read but totally optional

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -104,19 +104,75 @@ public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVectorId(int to
         return query(topK, null, null, null, id, namespace, filter, false, false);
     }
 
+    @Override
+    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVectorId(int topK,
+                                                                              String id,
```
aulorbe, line 109: Total nit, but maybe make `id` be `vectorId`, or make the method `querybyId`, just so the semantics of "id" carry through the method (or the semantics of "vector ID" if you go that route)

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -104,19 +104,75 @@ public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVectorId(int to
...
+                                                                            String namespace, Struct filter,
+                                                                            boolean includeValues,
+                                                                            boolean includeMetadata) {
+        return query(topK, vector, null, null, null, namespace, filter, includeValues, includeMetadata);
+    }
+
+    @Override
+    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVector(int topK, List<Float> vector,
+                                                                            String namespace, Struct filter) {
+        return query(topK, vector, null, null, null, namespace, filter, false, false);
```
aulorbe, line 148: Not saying this is wrong or right,  but OOC why do you allow users to have a `filter` when they query by vector, but not when they query by ID? I'd say they both probs would benefit from the inclusion of a `filter`, since users who use these types of queries (I'd imagine) are looking for similarity clusters, essentially. So getting all vectors similar to a given vector (or a given ID), filtered by `xyz` would be helpful

aulorbe (approved):
The code changes look good to me, but do we need any unit and integration tests for these? I think we do. I wouldn't merge w/o those.

src/main/java/io/pinecone/clients/AsyncIndex.java
austin-denoble: I started going down this path but we would need to update other functions and terminology and it felt like too much (all the validate stuff). For now just leaving it as `id` and the function name calls out what kind of id.

src/main/java/io/pinecone/clients/AsyncIndex.java
austin-denoble: I think we support what you're talking ab out on ln 100 in this file, am I misunderstanding?

```
    @Override
    public ListenableFuture<QueryResponseWithUnsignedIndices> queryByVectorId(int topK,
                                                                              String id,
                                                                              String namespace,
                                                                              Struct filter) {
        return query(topK, null, null, null, id, namespace, filter, false, false);
    }

```

austin-denoble:
> The code changes look good to me, but do we need any unit and integration tests for these? I think we do. I wouldn't merge w/o those.

We should have them, but we're a bit strapped for time at the moment. I've added Asana tickets to track adding unit tests for our `Index` and `AsyncIndex` classes along with `PineconeConnection`. For now, since these overloads are all calling the base function I think we should be ok.

PR #115, opened 2024-04-15
Add list endpoint

aulorbe:
## Problem

We need a `list` endpoint in the Java SDK to gain parity with our other SDKs. 

Asana ticket: https://app.asana.com/0/1203260648987893/1206928025656980/f

Documentation on this API endpoint: https://docs.pinecone.io/guides/data/manage-rag-documents#use-id-prefixes

## Solution

This PR enables this endpoint (generated code already provided syntax, it just was lacking implementation details in the SDK itself). 

New files: 
- `ListEndpointTest`: integration test for new endpoint
- `ListEndpointValidationTest`: unit tests for validation that occurs in new endpoint 

Updates to existing files:
- Updated `README`
- Updated `TestIndexResourcesManager`. Updates include:
    - Renamed `podIndexVectorIDs` to just `vectorIDs` so they can be used for serverless tests, too
    - Changed the content of this ^ ID list so that it contains IDs that start with different letters, in order to test the prefix filtering for the new endpoint
    - Added a `getServerlessIndexConnection` method b/c an `Index` obj is necessary to test the new endpoint (also made for `async`)
    - Made `seedIndex` accept a `namespace` parameter in order to test new endpoint + added it to the `getOrCreateServerlessIndex` method, so now that method seeds an index as well as builds one
-  Added `list` endpoint override and base methods in both `AsyncIndex.java` and`Index.java`
- Added abstract `list` definition in `IndexInterface.java` file

## Type of Change

## Test Plan

All tests locally and in GH are running successfully at the time of this writing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,24 @@ List<String> ids = Arrays.asList("v1", "v2", "v3");
 FetchResponse fetchResponse = index.fetch(ids, "example-namespace");
 ```
 
+## List vector IDs by Namespace
+The following example lists up to 100 vector IDs from a particular namespace.
+
+This method accepts optional parameters for `prefix` (list IDs filtered by a prefix), `limit` (retrieve `n` vector 
+IDs), and `paginationToken` (retrieve the next set of vector IDs).
+
+```java
+import io.pinecone.clients.Index;
+import io.pinecone.clients.Pinecone;
+import io.pinecone.proto.ListResponse;
+
+Pinecone pinecone = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
+String indexName = "example-index";
+Index index = pinecone.getIndexConnection(indexName);
+ListResponse listResponse = index.list("example-namespace");
+```
+
+
 ## Update vectors
 
 The following example updates vectors by ID.
--- a/README.md
+++ b/README.md
@@ -327,11 +327,13 @@ List<String> ids = Arrays.asList("v1", "v2", "v3");
 FetchResponse fetchResponse = index.fetch(ids, "example-namespace");
 ```
 
-## List vector IDs by Namespace
-The following example lists up to 100 vector IDs from a particular namespace.
+## List vector IDs
+The following example lists up to 100 vector IDs from a Pinecone index.
 
-This method accepts optional parameters for `prefix` (list IDs filtered by a prefix), `limit` (retrieve `n` vector 
-IDs), and `paginationToken` (retrieve the next set of vector IDs).
+This method accepts optional parameters for `namespace`, `prefix`, `limit`, and `paginationToken`. 
+
+The following 
+demonstrates how to use the `list` endpoint to get vector IDs from a specific `namespace`, filtered by a given `prefix`.
 
 ```java
 import io.pinecone.clients.Index;
@@ -341,7 +343,7 @@ import io.pinecone.proto.ListResponse;
 Pinecone pinecone = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
 String indexName = "example-index";
 Index index = pinecone.getIndexConnection(indexName);
-ListResponse listResponse = index.list("example-namespace");
+ListResponse listResponse = index.list("example-namespace", "prefix-");
 ```
 
 
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -81,8 +81,13 @@ public class TestIndexResourcesManager {
         return dimension;
     }
 
+
     public String getEnvironment() {
         return environment;
+    }
+
+    public String getNamespace() {
+        return this.namespace;
     }
 
     public List<String> getVectorIds() {
@@ -104,9 +109,6 @@ public class TestIndexResourcesManager {
         return collectionModel;
     }
 
-    public String getNamespace() {
-        return this.namespace;
-    }
 
     public void cleanupResources() {
         if (podIndexName != null) {
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -147,7 +147,7 @@ public class TestIndexResourcesManager {
         Thread.sleep(30000);
 
         // Seed default vector IDs into default namespace
-        seedDefaultNamespace(indexName);
+        seedIndex(vectorIdsForDefaultNamespace, indexName, defaultNamespace);
 
         this.podIndexName = indexName;
         return indexName;
@@ -169,8 +169,8 @@ public class TestIndexResourcesManager {
 
         // Seed default vector IDs into default namespace, seed custom vector IDs into custom namespace; all in
         // same index
-        seedDefaultNamespace(indexName);
-        seedCustomNamespace(indexName);
+        seedIndex(vectorIdsForDefaultNamespace, indexName, defaultNamespace);
+        seedIndex(vectorIdsForCustomNamespace, indexName, customNamespace);
 
         this.serverlessIndexName = indexName;
         return indexName;
@@ -207,37 +207,13 @@ public class TestIndexResourcesManager {
         return collectionName;
     }
 
-    private void seedDefaultNamespace(String indexName) throws InterruptedException {
+    private void seedIndex(List<String> vectorIds, String indexName, String namespace) throws InterruptedException {
         // Build upsert request
         Index indexClient = pineconeClient.getIndexConnection(indexName);
-
-        // Upsert vectors into both default namespace
-        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIdsForDefaultNamespace, dimension), defaultNamespace);
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
 
         // Wait for record freshness
         DescribeIndexStatsResponse indexStats = indexClient.describeIndexStats();
-
-        int totalTimeWaitedForVectors = 0;
-        while (indexStats.getTotalVectorCount() == 0 || totalTimeWaitedForVectors <= 60000) {
-            Thread.sleep(2000);
-            totalTimeWaitedForVectors += 2000;
-            indexStats = indexClient.describeIndexStats();
-        }
-        if (indexStats.getTotalVectorCount() == 0 && totalTimeWaitedForVectors >= 60000) {
-            throw new PineconeException("Failed to seed index " + indexName + "with vectors");
-        }
-    }
-
-    private void seedCustomNamespace(String indexName) throws InterruptedException {
-        // Build upsert request
-        Index indexClient = pineconeClient.getIndexConnection(indexName);
-
-        // Upsert vectors into both custom namespace
-        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIdsForCustomNamespace, dimension), customNamespace);
-
-        // Wait for record freshness
-        DescribeIndexStatsResponse indexStats = indexClient.describeIndexStats();
-
         int totalTimeWaitedForVectors = 0;
         while (indexStats.getTotalVectorCount() == 0 || totalTimeWaitedForVectors <= 60000) {
             Thread.sleep(2000);
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -43,7 +43,8 @@ public class TestIndexResourcesManager {
     private String collectionName;
     private CollectionModel collectionModel;
     private final List<String> vectorIds = Arrays.asList("id1", "id2", "prefix-id3");
-    private final String namespace = "example-namespace";
+    private final String customNamespace = "example-namespace";
+    private final String defaultNamespace = "";
 
 
     private TestIndexResourcesManager() {
@@ -86,8 +87,12 @@ public class TestIndexResourcesManager {
         return environment;
     }
 
-    public String getNamespace() {
-        return this.namespace;
+    public String getCustomNamespace() {
+        return this.customNamespace;
+    }
+
+    public String getDefaultNamespace() {
+        return this.defaultNamespace;
     }
 
     public List<String> getVectorIds() {
@@ -199,7 +204,12 @@ public class TestIndexResourcesManager {
     private void seedIndex(List<String> vectorIds, String indexName) throws InterruptedException {
         // Build upsert request
         Index indexClient = pineconeClient.getIndexConnection(indexName);
-        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
+
+        // Upsert vectors into both custom and default namespaces
+        List<String> namespaces = Arrays.asList(customNamespace, defaultNamespace);
+        for (String namespace : namespaces) {
+            indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
+        }
 
         // Wait for record freshness
         DescribeIndexStatsResponse indexStats = indexClient.describeIndexStats();
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -20,9 +20,8 @@ public class TestIndexResourcesManager {
     private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
     private static TestIndexResourcesManager instance;
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
-
     private final int dimension = System.getenv("DIMENSION") == null
-            ? 3
+            ? 4
             : Integer.parseInt(System.getenv("DIMENSION"));
     private final String environment = System.getenv("PINECONE_ENVIRONMENT") == null
             ? "us-east4-gcp"
@@ -40,16 +39,15 @@ public class TestIndexResourcesManager {
     private String podIndexName;
     private IndexModel podIndexModel;
     private IndexModel serverlessIndexModel;
-    private  String serverlessIndexName;
+    private String serverlessIndexName;
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> vectorIds = Arrays.asList("id1", "id2", "cidddd3");
+    private final List<String> vectorIds = Arrays.asList("id1", "id2", "prefix-id3");
     private final String namespace = "example-namespace";
 
 
     private TestIndexResourcesManager() {
         pineconeClient = new Pinecone.Builder(apiKey).build();
-
     }
 
     public static TestIndexResourcesManager getInstance() {
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -104,6 +104,10 @@ public class TestIndexResourcesManager {
         return collectionModel;
     }
 
+    public String getNamespace() {
+        return this.namespace;
+    }
+
     public void cleanupResources() {
         if (podIndexName != null) {
             pineconeClient.deleteIndex(podIndexName);
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -20,7 +20,7 @@ public class TestIndexResourcesManager {
     private static TestIndexResourcesManager instance;
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
     private final int dimension = System.getenv("DIMENSION") == null
-            ? 4
+            ? 3
             : Integer.parseInt(System.getenv("DIMENSION"));
     private final String environment = System.getenv("PINECONE_ENVIRONMENT") == null
             ? "us-east4-gcp"
@@ -37,13 +37,17 @@ public class TestIndexResourcesManager {
     private Pinecone pineconeClient;
     private String podIndexName;
     private IndexModel podIndexModel;
-    private String serverlessIndexName;
+    private IndexModel serverlessIndexModel;
+    private  String serverlessIndexName;
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> podIndexVectorIds = Arrays.asList("id1", "id2", "id3");
+    private final List<String> vectorIds = Arrays.asList("id1", "id2", "cidddd3");
+    private final String namespace = "example-namespace";
+
 
     private TestIndexResourcesManager() {
         pineconeClient = new Pinecone.Builder(apiKey).build();
+
     }
 
     public static TestIndexResourcesManager getInstance() {
@@ -51,6 +55,10 @@ public class TestIndexResourcesManager {
             instance = new TestIndexResourcesManager();
         }
         return instance;
+    }
+
+    public  Index getIndexConnection() {
+        return this.getInstance().pineconeClient.getIndexConnection(this.serverlessIndexName);
     }
 
     public String getPodIndexName() throws InterruptedException, PineconeException {
@@ -73,13 +81,18 @@ public class TestIndexResourcesManager {
         return environment;
     }
 
-    public List<String> getPodIndexVectorIds() {
-        return podIndexVectorIds;
+    public List<String> getVectorIds() {
+        return vectorIds;
     }
 
     public IndexModel getPodIndexModel() throws InterruptedException {
         podIndexModel = pineconeClient.describeIndex(createOrGetPodIndex());
         return podIndexModel;
+    }
+
+    public IndexModel getServerlessIndexModel() throws InterruptedException {
+        serverlessIndexModel = pineconeClient.describeIndex(createOrGetServerlessIndex());
+        return serverlessIndexModel;
     }
 
     public CollectionModel getCollectionModel() throws InterruptedException {
@@ -115,7 +128,7 @@ public class TestIndexResourcesManager {
         Thread.sleep(30000);
 
         // Seed data
-        seedIndex(podIndexVectorIds, indexName);
+        seedIndex(vectorIds, indexName);
 
         this.podIndexName = indexName;
         return indexName;
@@ -128,9 +141,17 @@ public class TestIndexResourcesManager {
 
         String indexName = RandomStringBuilder.build("serverless-index", 8);
 
-        pineconeClient.createServerlessIndex(indexName, metric, dimension, cloud, region);
+        serverlessIndexModel = pineconeClient.createServerlessIndex(indexName, metric, dimension, cloud,
+                region);
+        waitUntilIndexIsReady(pineconeClient, indexName);
 
-        serverlessIndexName = indexName;
+        // Explicitly wait after ready to avoid the "no healthy upstream" issue
+        Thread.sleep(30000);
+
+        // Seed data
+        seedIndex(vectorIds, indexName);
+
+        this.serverlessIndexName = indexName;
         return indexName;
     }
 
@@ -168,7 +189,7 @@ public class TestIndexResourcesManager {
     private void seedIndex(List<String> vectorIds, String indexName) throws InterruptedException {
         // Build upsert request
         Index indexClient = pineconeClient.getIndexConnection(indexName);
-        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), "");
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
 
         // Wait for record freshness
         DescribeIndexStatsResponse indexStats = indexClient.describeIndexStats();
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -1,5 +1,6 @@
 package io.pinecone.helpers;
 
+import io.pinecone.clients.AsyncIndex;
 import io.pinecone.clients.Index;
 import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeException;
@@ -19,6 +20,7 @@ public class TestIndexResourcesManager {
     private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
     private static TestIndexResourcesManager instance;
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
+
     private final int dimension = System.getenv("DIMENSION") == null
             ? 3
             : Integer.parseInt(System.getenv("DIMENSION"));
@@ -57,8 +59,12 @@ public class TestIndexResourcesManager {
         return instance;
     }
 
-    public  Index getIndexConnection() {
+    public  Index getServerlessIndexConnection() {
         return this.getInstance().pineconeClient.getIndexConnection(this.serverlessIndexName);
+    }
+
+    public AsyncIndex getServerlessAsyncIndexConnection() {
+        return this.getInstance().pineconeClient.getAsyncIndexConnection(this.serverlessIndexName);
     }
 
     public String getPodIndexName() throws InterruptedException, PineconeException {
--- a/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
+++ b/src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
@@ -42,7 +42,9 @@ public class TestIndexResourcesManager {
     private String serverlessIndexName;
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> vectorIds = Arrays.asList("id1", "id2", "prefix-id3");
+    private final List<String> vectorIdsForDefaultNamespace = Arrays.asList("def-id1", "def-id2", "def-prefix-id3", "def-prefix-id4");
+    private final List<String> vectorIdsForCustomNamespace = Arrays.asList("cus-id1", "cus-id2", "cus-prefix-id3", "cus" +
+            "-prefix-id4");
     private final String customNamespace = "example-namespace";
     private final String defaultNamespace = "";
 
@@ -59,11 +61,11 @@ public class TestIndexResourcesManager {
     }
 
     public  Index getServerlessIndexConnection() {
-        return this.getInstance().pineconeClient.getIndexConnection(this.serverlessIndexName);
+        return getInstance().pineconeClient.getIndexConnection(this.serverlessIndexName);
     }
 
     public AsyncIndex getServerlessAsyncIndexConnection() {
-        return this.getInstance().pineconeClient.getAsyncIndexConnection(this.serverlessIndexName);
+        return getInstance().pineconeClient.getAsyncIndexConnection(this.serverlessIndexName);
     }
 
     public String getPodIndexName() throws InterruptedException, PineconeException {
@@ -82,21 +84,24 @@ public class TestIndexResourcesManager {
         return dimension;
     }
 
-
     public String getEnvironment() {
         return environment;
     }
 
     public String getCustomNamespace() {
-        return this.customNamespace;
+        return customNamespace;
     }
 
     public String getDefaultNamespace() {
-        return this.defaultNamespace;
+        return defaultNamespace;
     }
 
-    public List<String> getVectorIds() {
-        return vectorIds;
+    public List<String> getVectorIdsFromDefaultNamespace() {
+        return vectorIdsForDefaultNamespace;
+    }
+
+    public List<String> getVectorIdsFromCustomNamespace() {
+        return vectorIdsForCustomNamespace;
     }
 
     public IndexModel getPodIndexModel() throws InterruptedException {
@@ -113,7 +118,6 @@ public class TestIndexResourcesManager {
         collectionModel = pineconeClient.describeCollection(createOrGetCollection());
         return collectionModel;
     }
-
 
     public void cleanupResources() {
         if (podIndexName != null) {
@@ -142,8 +146,8 @@ public class TestIndexResourcesManager {
         // Explicitly wait after ready to avoid the "no healthy upstream" issue
         Thread.sleep(30000);
 
-        // Seed data
-        seedIndex(vectorIds, indexName);
+        // Seed default vector IDs into default namespace
+        seedDefaultNamespace(indexName);
 
         this.podIndexName = indexName;
         return indexName;
@@ -163,8 +167,10 @@ public class TestIndexResourcesManager {
         // Explicitly wait after ready to avoid the "no healthy upstream" issue
         Thread.sleep(30000);
 
-        // Seed data
-        seedIndex(vectorIds, indexName);
+        // Seed default vector IDs into default namespace, seed custom vector IDs into custom namespace; all in
+        // same index
+        seedDefaultNamespace(indexName);
+        seedCustomNamespace(indexName);
 
         this.serverlessIndexName = indexName;
         return indexName;
@@ -201,15 +207,33 @@ public class TestIndexResourcesManager {
         return collectionName;
     }
 
-    private void seedIndex(List<String> vectorIds, String indexName) throws InterruptedException {
+    private void seedDefaultNamespace(String indexName) throws InterruptedException {
         // Build upsert request
         Index indexClient = pineconeClient.getIndexConnection(indexName);
 
-        // Upsert vectors into both custom and default namespaces
-        List<String> namespaces = Arrays.asList(customNamespace, defaultNamespace);
-        for (String namespace : namespaces) {
-            indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
+        // Upsert vectors into both default namespace
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIdsForDefaultNamespace, dimension), defaultNamespace);
+
+        // Wait for record freshness
+        DescribeIndexStatsResponse indexStats = indexClient.describeIndexStats();
+
+        int totalTimeWaitedForVectors = 0;
+        while (indexStats.getTotalVectorCount() == 0 || totalTimeWaitedForVectors <= 60000) {
+            Thread.sleep(2000);
+            totalTimeWaitedForVectors += 2000;
+            indexStats = indexClient.describeIndexStats();
         }
+        if (indexStats.getTotalVectorCount() == 0 && totalTimeWaitedForVectors >= 60000) {
+            throw new PineconeException("Failed to seed index " + indexName + "with vectors");
+        }
+    }
+
+    private void seedCustomNamespace(String indexName) throws InterruptedException {
+        // Build upsert request
+        Index indexClient = pineconeClient.getIndexConnection(indexName);
+
+        // Upsert vectors into both custom namespace
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIdsForCustomNamespace, dimension), customNamespace);
 
         // Wait for record freshness
         DescribeIndexStatsResponse indexStats = indexClient.describeIndexStats();
--- a/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
@@ -35,12 +35,14 @@ public class CollectionTest {
     private static CollectionModel collection;
     private static String indexName;
     private static String collectionName;
+    private static String namespace;
 
     @BeforeAll
     public static void setUp() throws InterruptedException {
         indexName = indexManager.getPodIndexName();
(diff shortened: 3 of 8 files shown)


src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -20,7 +20,7 @@ public class TestIndexResourcesManager {
     private static TestIndexResourcesManager instance;
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
     private final int dimension = System.getenv("DIMENSION") == null
-            ? 4
+            ? 3
```
aulorbe, line 23: Dimension has to == the # of vectors we upsert; so `vectorIds`, i.e. `3`

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -53,6 +57,10 @@ public static TestIndexResourcesManager getInstance() {
         return instance;
     }
 
+    public  Index getIndexConnection() {
```
aulorbe, line 60: Need `Index` obj to test new `list` endpoint

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -73,15 +81,20 @@ public String getEnvironment() {
         return environment;
     }
 
-    public List<String> getPodIndexVectorIds() {
-        return podIndexVectorIds;
+    public List<String> getVectorIds() {
```
aulorbe, line 84: Changed so can be index-type agnostic

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -73,15 +81,20 @@ public String getEnvironment() {
...
+        return vectorIds;
     }
 
     public IndexModel getPodIndexModel() throws InterruptedException {
         podIndexModel = pineconeClient.describeIndex(createOrGetPodIndex());
         return podIndexModel;
     }
 
+    public IndexModel getServerlessIndexModel() throws InterruptedException {
```
aulorbe, line 93: Added for symmetry

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -128,9 +141,17 @@ private String createOrGetServerlessIndex() throws InterruptedException, Pinecon
...
-        pineconeClient.createServerlessIndex(indexName, metric, dimension, cloud, region);
+        serverlessIndexModel = pineconeClient.createServerlessIndex(indexName, metric, dimension, cloud,
+                region);
+        waitUntilIndexIsReady(pineconeClient, indexName);
+
+        // Explicitly wait after ready to avoid the "no healthy upstream" issue
+        Thread.sleep(30000);
+
+        // Seed data
+        seedIndex(vectorIds, indexName);
```
aulorbe, line 152: Now this method seeds the index as well as creates it

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -168,7 +189,7 @@ private String createOrGetCollection() throws InterruptedException {
     private void seedIndex(List<String> vectorIds, String indexName) throws InterruptedException {
         // Build upsert request
         Index indexClient = pineconeClient.getIndexConnection(indexName);
-        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), "");
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
```
aulorbe, line 192: Now seeding can take a namespace arg

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        ListenableFuture<ListResponse> futureResponseWithPrefix = asyncIndexConnection.list("example-namespace", "i");
+        ListResponse asyncListResponseWithPrefix = Futures.getUnchecked(futureResponseWithPrefix);
+        assertTrue(asyncListResponseWithPrefix.toString().contains("id1"));
+        assertTrue(asyncListResponseWithPrefix.toString().contains("id2"));
+        assertFalse(asyncListResponseWithPrefix.toString().contains("cidddd3"));
+
+        ListenableFuture<ListResponse> futureResponseWithLimit = asyncIndexConnection.list("example-namespace", 1);
+        ListResponse asyncListResponseWithLimit = Futures.getUnchecked(futureResponseWithLimit);
+        assertTrue(asyncListResponseWithLimit.toString().contains("cidddd3"));
```
aulorbe, line 73: Not sure if this is the best way to test async stuff. ChatGPT also told me I could make a `Futures.addCallback()` that registers a callback to be executed upon completion of the `Future`? What's best practice here?

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -37,13 +39,17 @@ public class TestIndexResourcesManager {
     private Pinecone pineconeClient;
     private String podIndexName;
     private IndexModel podIndexModel;
-    private String serverlessIndexName;
+    private IndexModel serverlessIndexModel;
+    private  String serverlessIndexName;
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> podIndexVectorIds = Arrays.asList("id1", "id2", "id3");
+    private final List<String> vectorIds = Arrays.asList("id1", "id2", "cidddd3");
```
aulorbe, line 46: Changed so can be index-type agnostic

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -29,7 +29,7 @@ public class CollectionTest {
     private static final Logger logger = LoggerFactory.getLogger(CollectionTest.class);
     private static final ArrayList<String> indexesToCleanUp = new ArrayList<>();
     private static final String indexMetric = IndexMetric.COSINE.toString();
-    private static final List<String> upsertIds = indexManager.getPodIndexVectorIds();
+    private static final List<String> upsertIds = indexManager.getVectorIds();
```
aulorbe, line 32: Just name change; can ignore

aulorbe:
@austin-denoble This is the failed test. I believe it's failing due to flakiness, since we've seen this before, and it's related to pods, which I didn't update in this PR. Thoughts?

```
CollectionTest > testIndexFromCollectionHappyPath() FAILED
    java.lang.AssertionError: expected: <true> but was: <false>
        at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:35)
        at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:13)
        at io.pinecone.integration.controlPlane.pod.CollectionTest.testIndexFromCollectionHappyPath(CollectionTest.java:106)
```

austin-denoble:
> @austin-denoble This is the failed test. I believe it's failing due to flakiness, since we've seen this before, and it's related to pods, which I didn't update in this PR. Thoughts?
> 
> ```
> CollectionTest > testIndexFromCollectionHappyPath() FAILED
>     java.lang.AssertionError: expected: <true> but was: <false>
>         at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:35)
>         at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:13)
>         at io.pinecone.integration.controlPlane.pod.CollectionTest.testIndexFromCollectionHappyPath(CollectionTest.java:106)
> ```

@aulorbe yeah I just looked at this one, definitely flakiness. It looks like the index created from the collection had some delays in its vector data being fresh or current, from the assertion. Fine to ignore this one for now I think.

aulorbe:
> > @austin-denoble This is the failed test. I believe it's failing due to flakiness, since we've seen this before, and it's related to pods, which I didn't update in this PR. Thoughts?
> > ```
> > CollectionTest > testIndexFromCollectionHappyPath() FAILED
> >     java.lang.AssertionError: expected: <true> but was: <false>
> >         at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:35)
> >         at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:13)
> >         at io.pinecone.integration.controlPlane.pod.CollectionTest.testIndexFromCollectionHappyPath(CollectionTest.java:106)
> > ```
> 
> @aulorbe yeah I just looked at this one, definitely flakiness. It looks like the index created from the collection had some delays in its vector data being fresh or current, from the assertion. Fine to ignore this one for now I think.

Cool thanks!

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
austin-denoble: It shouldn't need to match the number of vectors we upsert, just the dimensionality of the vectors themselves, right?

This should be alright to change given the tests grab the dimension from here. If we use 3 though there's a possibility the indexes created by the `TestIndexResourcesManager` could be confused with indexes created by the current `IndexManager` since I think a lot of those indexes have a dimension of 3 and that's how the lookup is done. That'll be going away in my PR (https://github.com/pinecone-io/pinecone-java-client/pull/99), but we may want to leave this set to 4 for now if we can.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
austin-denoble: Just so it's really clear what the purpose of the prefix is, maybe just use `prefix-id3`?

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -37,13 +39,17 @@ public class TestIndexResourcesManager {
     private Pinecone pineconeClient;
     private String podIndexName;
     private IndexModel podIndexModel;
-    private String serverlessIndexName;
+    private IndexModel serverlessIndexModel;
+    private  String serverlessIndexName;
```
austin-denoble, line 43: nit: Looks like there's an extra space before `String` here.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -37,13 +39,17 @@ public class TestIndexResourcesManager {
...
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> podIndexVectorIds = Arrays.asList("id1", "id2", "id3");
+    private final List<String> vectorIds = Arrays.asList("id1", "id2", "cidddd3");
+    private final String namespace = "example-namespace";
+
 
     private TestIndexResourcesManager() {
         pineconeClient = new Pinecone.Builder(apiKey).build();
+
```
austin-denoble, line 52: nit: can probably remove this extra return

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -53,6 +59,14 @@ public static TestIndexResourcesManager getInstance() {
         return instance;
     }
 
+    public  Index getServerlessIndexConnection() {
```
austin-denoble, line 62: Cool, for a lot of my tests I was using the `getServerlessIndexName` / `getPodIndexName` calls to get the name and then call `.getIndexConnection()` or `.getAsyncIndexConnection` in the test file. This feels nicer though since it keeps the index'y pieces together. I can update my tests with this when it's merged. 👍

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -19,8 +20,9 @@ public class TestIndexResourcesManager {
     private static final Logger logger = LoggerFactory.getLogger(IndexManager.class);
     private static TestIndexResourcesManager instance;
     private static final String apiKey = System.getenv("PINECONE_API_KEY");
+
```
austin-denoble, line 23: nit: can probably remove this extra return

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -37,13 +39,17 @@ public class TestIndexResourcesManager {
...
     private String podIndexName;
     private IndexModel podIndexModel;
-    private String serverlessIndexName;
+    private IndexModel serverlessIndexModel;
+    private  String serverlessIndexName;
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> podIndexVectorIds = Arrays.asList("id1", "id2", "id3");
+    private final List<String> vectorIds = Arrays.asList("id1", "id2", "cidddd3");
+    private final String namespace = "example-namespace";
```
austin-denoble, line 47: For most of our data plane tests we create a namespace in the test file and use that for the specific test. I think it should be fine to define a namespace at this level.

We should create a getter for this though, so any consuming tests have access to the namespace that data was seeded under, so I'd look at adding that.

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private static final TestIndexResourcesManager indexManager = TestIndexResourcesManager.getInstance();
+    private static String indexName ;
+    private static Index indexConnection;
+    private static AsyncIndex asyncIndexConnection;
+
+    @BeforeAll
+    public static void setUp() throws InterruptedException {
+        indexName = indexManager.getServerlessIndexName();
+        indexConnection = indexManager.getServerlessIndexConnection();
+        asyncIndexConnection = indexManager.getServerlessAsyncIndexConnection();
```
austin-denoble, line 26: Clean! I like it.

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @AfterAll
+    public static void cleanUp() {
+        String apiKey = System.getenv("PINECONE_API_KEY");
+        Pinecone pinecone = new Pinecone.Builder(apiKey).build();
+        pinecone.deleteIndex(indexName);
```
austin-denoble, line 33: You do not need to explicitly delete the index here. It's meant to be shared across tests so having a single test file delete the index means the `TestIndexResourcesManager` will need to create it again for any additional tests, and wait to seed, which takes a lot of time.

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -0,0 +1,78 @@
...
+        indexConnection.close();
+        asyncIndexConnection.close();
+    }
+
+    @Test
+    public void testListEndpoint() throws InterruptedException {
+        Thread.sleep(10000); // wait for the index to be ready for operations
```
austin-denoble, line 40: We should be calling `waitUntilIndexReady()` with an additional `Thread.sleep()` in the `TestIndexResourcesManager.createOrGetServerlessIndex()` so we really shouldn't need to wait again here. I think you can probably pull this out unless you're seeing issues.

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -0,0 +1,78 @@
...
+        assertTrue(listResponseWithPrefix.contains("id1"));
+        assertTrue(listResponseWithPrefix.contains("id2"));
+        assertFalse(listResponseWithPrefix.contains("cidddd3")); // should not be in response
+
+        String listResponseWithLimit = indexConnection.list("example-namespace", 1).toString();
+        assertTrue(listResponseWithLimit.contains("cidddd3")); // (IDs returned in LIFO order)
+        assertFalse(listResponseWithLimit.contains("id2")); // should not be in response
+        assertFalse(listResponseWithLimit.contains("id1")); // should not be in response
+
+        //Async
```
austin-denoble, line 58: I would break this one large test function into different test functions that are specific to what is being tested. It just makes it a bit easier to read and understand, imo.

At least break the async vs. sync stuff into different functions, this is what Rohans been doing for a lot of our tests.

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
austin-denoble: I think for testing purposes we basically want to allow the future to execute rather than waiting on a callback and trying to assert that way. I needed to look up `Futures.getUnchecked()` but it seems like this is equivalent to what we're doing elsewhere where we're chaining `.get()` at the end.

Also looks like that's deprecated and what you've done is preferred, looks good to me! 
https://guava.dev/releases/19.0/api/docs/com/google/common/util/concurrent/Futures.html

Screenshot 2024-04-17 at 2 40 48 PM

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -846,6 +849,51 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
...
+    @Override
+    public ListenableFuture<ListResponse>  list(String namespace, String prefix, String paginationToken) {
+        validateListEndpointParameters(namespace, prefix, paginationToken, null, true, true, false);
+        ListRequest listRequest = ListRequest.newBuilder().setNamespace(namespace).setPrefix(prefix).
+                setPaginationToken(paginationToken).setLimit(100).build();
+        return asyncStub.list(listRequest);
+    }
+
+    @Override
+    public ListenableFuture<ListResponse>  list(String namespace, String prefix, String paginationToken, Integer limit) {
```
austin-denoble, line 890: super nit: I think there's extra space between `ListenableFuture<ListResponse>    list(...` in a lot of these. Is that just me?

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -846,6 +849,51 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
         return asyncStub.describeIndexStats(describeIndexStatsRequest);
     }
 
+    @Override
+    public ListenableFuture<ListResponse>  list(String namespace) {
```
austin-denoble, line 853: We also need docstrings in this file, you should be able to lift your docstrings from `Index` and put them here and update `Index` -> `AsyncIndex`.

Appreciate you adding the variety of overrides here. We discussed offline but we should really look at implementing some builder patterns for some of these more complicated interfaces at some point.

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -13,6 +13,8 @@
 
 import java.util.List;
 
+import static io.pinecone.clients.Index.validateListEndpointParameters;
```
austin-denoble, line 16: Since this is used in both places, it should be implemented as a concrete function in `IndexInterface`. You can look at the other validator functions on the interface for examples: https://github.com/pinecone-io/pinecone-java-client/blob/ee8c077d5707a27050705df2f5252d59537a13fb/src/main/java/io/pinecone/commons/IndexInterface.java#L42

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -779,6 +780,207 @@ public DescribeIndexStatsResponse describeIndexStats(Struct filter) {
...
+     *  <pre>{@code
+     *      import io.pinecone.clients.Index;
+     *      import io.pinecone.clients.Pinecone;
+     *      import io.pinecone.proto.ListResponse;
+     *
+     *      Pinecone pc = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
+     *      String indexName = "example-index";
+     *      Index index = pc.getIndexConnection(indexName);
+     *      ListResponse listResponse = index.list("example-namespace");
+     *      System.out.println(listResponse);
```
austin-denoble, line 795: nit: I would lose the `System.out.println()` statements in all of these. I don't think there's much value in it and we're not doing it elsewhere as far as I know.

austin-denoble, line 795: I'd look at some of the other code snippets in this file and `AsyncIndex`. @ssmith-pc had suggested moving to a format where we drop some of the cruft around importing and instantiating the client to focus on the specifics of the operation. So for this one it would look something like:
```suggestion
     *      import io.pinecone.proto.ListResponse;
     *
     *     ...
     *
     *      ListResponse listResponse = index.list("example-namespace");
```

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -779,6 +780,207 @@ public DescribeIndexStatsResponse describeIndexStats(Struct filter) {
         return blockingStub.describeIndexStats(describeIndexStatsRequest);
     }
 
+    /**
+     * {@inheritDoc}
+     * <p> Example</p>
```
austin-denoble, line 785: super nit: I'd lose the leading space in these `Example:` lines and add a colon, to match the other examples in other docstrings.

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -779,6 +780,207 @@ public DescribeIndexStatsResponse describeIndexStats(Struct filter) {
...
+            , boolean prefixRequired, boolean paginationTokenRequired, boolean limitRequired) {
+        if (namespace == null || namespace.isEmpty()) {
+            throw new PineconeValidationException("Namespace cannot be null or empty");
+        }
+        if (prefixRequired && (prefix == null || prefix.isEmpty())) {
+            throw new PineconeValidationException("Prefix cannot be null or empty");
+        }
+        if (paginationTokenRequired && (paginationToken == null || paginationToken.isEmpty())) {
+            throw new PineconeValidationException("Pagination token cannot be null or empty");
+        }
```
austin-denoble, line 977: Are there actually errors that are thrown if these values end up `null`? I know for the TypeScript client you could send requests with these fields empty and the server would handle things properly.

The `prefixRequired` is also a little difficult to reason about, we could maybe drop it if we don't actually need to validate that field. 🤔

src/test/java/io/pinecone/ListEndpointValidationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ListEndpointValidationTest {
+    @Test
+    public void testValidateListEndpointParameters() {
+
+        PineconeValidationException thrownNullNamespace =assertThrows(PineconeValidationException.class, () -> {
+            Index.validateListEndpointParameters(null, null, null, null, true, true, true);
+        });
+        assertEquals("Namespace cannot be null or empty", thrownNullNamespace.getMessage());
```
austin-denoble, line 17: nit: This would be tedious but breaking things into distinct tests with names that call out what is being tested might be nice for readability.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: Oh you're 100% right! I'll change it back to `4` :)

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: ty!

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: It hasn't seemed to _break_ anything, so I think it's okay! But yeah great idea re: the getter method. Adding that now.

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: yeah!

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
aulorbe: 🎉

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
aulorbe: Ahhh okay. But we don't explicitly delete it in the `TestIndexResourcesManager` singleton, do we?

aulorbe: Oh wait I think we can via the `cleanupResources` method... and I see now that it's used in the `CleanupAllTestResourcesListener` class. Nice!!

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
aulorbe: True, ty!

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
aulorbe: Okay cool, I'll leave it as-is then :)

src/main/java/io/pinecone/commons/IndexInterface.java
```diff
@@ -697,4 +697,63 @@ W update(String id, List<Float> values, Struct metadata, String namespace, List<
...
+    /**
+     * Retrieves up to 100 vector IDs from a given namespace.
+     *
+     * @param namespace The namespace that holds the vector IDs you want to retrieve.
+     * @return A generic type {@code Y} that contains vector IDs.
+     */
+    Z list(String namespace);
+
+    /**
+     * Retrieves up to `n` vector IDs from a given namespace, where `limit` == `n`.
```
austin-denoble, line 710: I don't think the backticks render in the docs output. You can double check by doing `gradle generateJavadoc` and then opening the `index.html` file in the docs folder yourself.

I've been using the `{@code n}` format myself, but it's up to you.

austin-denoble (approved):
Great work! Thanks for adding integration tests, unit tests, and updating all the relevant classes. I left a lot of feedback and a lot of nits - I'm assuming you can pick and choose how you go through those.

The things I'd really like to see before merging are `validateListEndpointParameters` moving to `IndexInterface`, and adding docstring examples to `AsyncIndex`.

src/main/java/io/pinecone/clients/AsyncIndex.java
aulorbe: Done!

src/main/java/io/pinecone/clients/AsyncIndex.java
aulorbe: Oops yes, thank you!

(And as discussed offline, I'll make a separate ticket to follow up this PR with a transform of these methods into a Builder pattern)

src/main/java/io/pinecone/clients/Index.java
aulorbe: Done!

src/main/java/io/pinecone/clients/Index.java
aulorbe: I thinkkkk we do need to validate that field b/c the logic is like:
- If you use the overloaded method whose signature is `namespace, prefix`, then you need a prefix. 
- So the test is like "if you're using a method where prefix is necessary, make sure it's not empty"

Does that make sense? Not sure ...

aulorbe: And by "if these values end up null?" do you mean like, if they result in a `null`, but aren't passed as `null` originally (I'm thinking like maybe division by 0 or something?)

austin-denoble: I meant if we remove the validations and allow passing null for all fields, what happens then?

src/test/java/io/pinecone/ListEndpointValidationTest.java
aulorbe: Done!

aulorbe: Also changed so now it tests the method in the interface class (had to mock implementing class `Index`)

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -779,6 +780,207 @@ public DescribeIndexStatsResponse describeIndexStats(Struct filter) {
...
+        validateListEndpointParameters(namespace, prefix, paginationToken, null, true, true, false);
+        ListRequest listRequest = ListRequest.newBuilder().setNamespace(namespace).setPrefix(prefix).
+                setPaginationToken(paginationToken).setLimit(100).build();
+        return blockingStub.list(listRequest);
+    }
+
+    /**
+     * Base method that retrieves a list of vector IDs from a specific namespace within an index.
+     *
+     * <p>The method internally constructs a {@link ListRequest} using the provided namespace
```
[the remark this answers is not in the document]
aulorbe: ahh you're right (looked it up!)! I'll change

src/main/java/io/pinecone/commons/IndexInterface.java
aulorbe: Ah ty!

src/main/java/io/pinecone/clients/Index.java
aulorbe: Just added a `list` method that accepts `0` args. My bad! Now we've got one super minimal (no args) and one super maximal (all args). It's not ideal to have so many variations, but at least we have a fast-follow planned to transform this whole endpoint into the Builder pattern. Things will be more clear then.

src/main/java/io/pinecone/commons/IndexInterface.java
```diff
@@ -25,7 +25,7 @@
  * @param <X> The return type for delete operations.
  * @param <Y> The return type for describing index stats.
  */
```
austin-denoble, line 27: Could you add the `Z` generic parameter to the docstring above?
```suggestion
 * @param <X> The return type for delete operations.
 * @param <Y> The return type for describing index stats.
 * @param <Z> The return type for the list operation.
 */
```

src/test/java/io/pinecone/ListEndpointValidationTest.java
```diff
@@ -0,0 +1,104 @@
...
+        // Mock sync Pinecone connection
+        connectionMock = mock(PineconeConnection.class);
+        stubMock = mock(VectorServiceGrpc.VectorServiceBlockingStub.class);
+        when(connectionMock.getBlockingStub()).thenReturn(stubMock);
+
+        index = new Index(connectionMock, indexName);
+    }
+
+    // TODO: write one for no args
+    // TODO: write ones for async
```
austin-denoble, line 39: Stray TODOs 👀

src/test/java/io/pinecone/ListEndpointValidationTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Test
+    public void testValidateSyncListNamespace() throws IOException {
```
austin-denoble, line 42: Thanks for splitting these up!

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -846,6 +848,178 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
         return asyncStub.describeIndexStats(describeIndexStatsRequest);
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Example
```
austin-denoble, line 854: "Example" text seem to be missing colons in this file.

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -110,7 +110,7 @@ public void testIndexFromCollectionHappyPath() throws InterruptedException {
                 assertEquals(describeResponse.getTotalVectorCount(), 3);
 
                 // Verify the vectors from the collection -> new index can be fetched
-                FetchResponse fetchedVectors = indexClient.fetch(upsertIds, "");
+                FetchResponse fetchedVectors = indexClient.fetch(upsertIds, "example-namespace");
```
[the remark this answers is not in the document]
austin-denoble: Sounds good, could we use a getter from `TestIndexResourcesManager` to get this rather than hard-coding?

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -846,6 +848,178 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
...
+    public ListenableFuture<ListResponse> list(String namespace, String prefix) {
+        validateListEndpointParameters(namespace, prefix, null, null, true, true, false, false);
+        ListRequest listRequest = ListRequest.newBuilder().setNamespace(namespace).setPrefix(prefix).build();
+        return asyncStub.list(listRequest);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * <p>Example>
```
austin-denoble, line 942: Leftover ">" here

austin-denoble (approved):
Overall this looks great, thanks for iterating!

src/main/java/io/pinecone/clients/AsyncIndex.java
aulorbe: ty!

src/main/java/io/pinecone/clients/AsyncIndex.java
aulorbe: yes!

src/test/java/io/pinecone/ListEndpointValidationTest.java
aulorbe: yes yes sry this commit was just to show you a preview of the no-args version, so it wasn't formal. Will clean up!

src/main/java/io/pinecone/commons/IndexInterface.java
aulorbe: done!

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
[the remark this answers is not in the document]
aulorbe: Def!

aulorbe:
> ```
> UpsertAndQueryServerlessTest > upsertOptionalVectorsAndQueryIndexSyncTest() FAILED
>     java.lang.AssertionError: expected: not <null>
>         at io.pinecone.helpers.AssertRetry.assertWithRetry(AssertRetry.java:35)
>         at io.pinecone.integration.dataPlane.UpsertAndQueryServerlessTest.upsertOptionalVectorsAndQueryIndexSyncTest(UpsertAndQueryServerlessTest.java:82)
> ```
> 
> Is passing for me locally, and I debugged each line to ensure everything is checking out as it should. I'll increase the `backoff` from `4` to `5` if it fails again (writing this as it's still running)

Oo it passed this time, yay! Must've been flakiness. I've got a message out to @rohanshah18 about adding in exponential backoffs and/or jitter to our retries, so we'll see if we can avoid this in the future!

README.md
```diff
@@ -327,6 +327,24 @@ List<String> ids = Arrays.asList("v1", "v2", "v3");
...
+
+```java
+import io.pinecone.clients.Index;
+import io.pinecone.clients.Pinecone;
+import io.pinecone.proto.ListResponse;
+
+Pinecone pinecone = new Pinecone.Builder(System.getenv("PINECONE_API_KEY")).build();
+String indexName = "example-index";
+Index index = pinecone.getIndexConnection(indexName);
+ListResponse listResponse = index.list("example-namespace");
```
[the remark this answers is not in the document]
aulorbe: Sure!

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -53,6 +58,14 @@ public static TestIndexResourcesManager getInstance() {
         return instance;
     }
 
+    public  Index getServerlessIndexConnection() {
+        return this.getInstance().pineconeClient.getIndexConnection(this.serverlessIndexName);
+    }
+
+    public AsyncIndex getServerlessAsyncIndexConnection() {
+        return this.getInstance().pineconeClient.getAsyncIndexConnection(this.serverlessIndexName);
+    }
```
ssmith-pc, line 67: nit, you don't need to specify `this.` in most of these methods. The only time you need to specify `this.` is when there is a conflicting parameter, like there would typically be in a setter method.

```java
public void setName(String name) {
    this.name = name;
}
```

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -37,10 +38,14 @@ public class TestIndexResourcesManager {
     private Pinecone pineconeClient;
     private String podIndexName;
     private IndexModel podIndexModel;
+    private IndexModel serverlessIndexModel;
     private String serverlessIndexName;
     private String collectionName;
     private CollectionModel collectionModel;
-    private final List<String> podIndexVectorIds = Arrays.asList("id1", "id2", "id3");
+    private final List<String> vectorIds = Arrays.asList("id1", "id2", "prefix-id3");
```
ssmith-pc, line 45: Since we're adding support for getting a list of things, we should probably test for retrieving multiple things that start with `prefix-`?

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -168,7 +204,12 @@ private String createOrGetCollection() throws InterruptedException {
...
         Index indexClient = pineconeClient.getIndexConnection(indexName);
-        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), "");
+
+        // Upsert vectors into both custom and default namespaces
+        List<String> namespaces = Arrays.asList(customNamespace, defaultNamespace);
+        for (String namespace : namespaces) {
+            indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIds, dimension), namespace);
+        }
 
         // Wait for record freshness
```
ssmith-pc, line 214: Now that we're upserting multiple, do we want to check for the total number of vectors we expect before continuing?

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
ssmith-pc: I'm also thinking it would be helpful--both in terms of readability as well as testing functionality--if we used separate IDs for the set of IDs that are stored in the default namespace vs. those stored in the named namespace. 

That way it'll make test assertions more easy to follow as well as ensure that we don't accidentally verify that the wrong behavior is correct, just b/c the ideas we were checking happened to be the same.

I'm thinking something simple like:

```java
private final List<String> vectorIdsDefaultNs = Arrays.asList("def-id1", "def-id2", "def-prefix-id3", "def-prefix-id4");
private final List<String> vectorIdsCustomNs = Arrays.asList("ns-id1", "ns-id2", "ns-prefix-id3", "ns-prefix-id4");
```

And then activated/used in the appropriate tests

src/main/java/io/pinecone/commons/IndexInterface.java
```diff
@@ -697,4 +698,120 @@ W update(String id, List<Float> values, Struct metadata, String namespace, List<
...
+     * @param namespace The namespace parameter which is validated based on the {@code namespaceRequired} flag.
+     * @param prefix The prefix parameter which is validated based on the {@code prefixRequired} flag.
+     * @param paginationToken The pagination token parameter which is validated based on the {@code paginationTokenRequired} flag.
+     * @param limit The limit for the number of items, validated to be a positive integer if {@code limitRequired} is true.
+     * @param namespaceRequired Specifies if the namespace parameter is required and should be validated.
+     * @param prefixRequired Specifies if the prefix parameter is required and should be validated.
+     * @param paginationTokenRequired Specifies if the pagination token parameter is required and should be validated.
+     * @param limitRequired Specifies if the limit parameter is required and should be a positive integer.
+     * @throws PineconeValidationException if any parameter fails its validation check based on its requirements.
+     */
```
ssmith-pc, line 733: Props for including jdoc in the PR

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -846,6 +848,178 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
...
+     *     ListResponse asyncListResponse = Futures.getUnchecked(futureResponse);
+     *  }</pre>
+     */
+    @Override
+    public ListenableFuture<ListResponse> list(String namespace, String prefix, Integer limit) {
+        validateListEndpointParameters(namespace, prefix, null, limit, true, true, false, true);
+        ListRequest listRequest = ListRequest.newBuilder().setNamespace(namespace).setPrefix(prefix).
+                setLimit(limit).build();
+        return asyncStub.list(listRequest);
+    }
```
ssmith-pc, line 960: I would consider changing the type of `limit` to be an `int`. We can keep it as an `Integer` in the `validateListEndpointParameters` method to allow method overloads to set it to null when it's not relevant.

src/main/java/io/pinecone/commons/IndexInterface.java
```diff
@@ -697,4 +698,120 @@ W update(String id, List<Float> values, Struct metadata, String namespace, List<
...
+    default void validateListEndpointParameters(String namespace, String prefix, String paginationToken, Integer limit
+            , boolean namespaceRequired, boolean prefixRequired, boolean paginationTokenRequired,
+                                                boolean limitRequired) {
```
ssmith-pc, line 736: The format of these params is non-standard

ssmith-pc (approved):
Overall looks good! Love that you included jdocs and tests all in the same PR. 

Left a few minor comments

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: Ahh great idea!

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: ah ty!

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
aulorbe: I'm actually about to change this ,but I'll move this logic to `createOrGetServerlessIndex` instead

src/main/java/io/pinecone/clients/AsyncIndex.java
aulorbe: Ah ty! I still am not used to `int` vs `Integer`, thanks!

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public void testAsyncListEndpoint() throws InterruptedException {
+        // Confirm default vector IDs are returned when no namespace is specified
+        ListenableFuture<ListResponse> futureResponseNoArgs = asyncIndexConnection.list();
+        ListResponse asyncListResponseNoArgs = Futures.getUnchecked(futureResponseNoArgs);
+
+        assertEquals(asyncListResponseNoArgs.getVectorsList().size(), defaultVectorIDs.size());
+
+        assertTrue(asyncListResponseNoArgs.getVectorsList().toString().contains(defaultVectorIDs.get(0)));
+        assertTrue(asyncListResponseNoArgs.getVectorsList().toString().contains(defaultVectorIDs.get(1)));
+        assertTrue(asyncListResponseNoArgs.getVectorsList().toString().contains(defaultVectorIDs.get(2)));
```
ssmith-pc, line 95: I get why you're using `defaultVectorIDs.get(0)`, etc. everywhere, but I think it makes the test harder to read since it obscures the expectation

src/integration/java/io/pinecone/helpers/TestIndexResourcesManager.java
```diff
@@ -165,10 +207,33 @@ private String createOrGetCollection() throws InterruptedException {
...
+            throw new PineconeException("Failed to seed index " + indexName + "with vectors");
+        }
+    }
+
+    private void seedCustomNamespace(String indexName) throws InterruptedException {
+        // Build upsert request
+        Index indexClient = pineconeClient.getIndexConnection(indexName);
+
+        // Upsert vectors into both custom namespace
+        indexClient.upsert(buildRequiredUpsertRequestByDimension(vectorIdsForCustomNamespace, dimension), customNamespace);
```
ssmith-pc, line 237: Since the seed methods are largely the same, it seems like they could just be the same method that takes a namespace and list of vectors as input.

aulorbe: Ahhh true !

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
aulorbe: ah okay! I can hardcode :)

PR #116, opened 2024-04-25
Add new `build-and-publish-docs.yml` GitHub Workflow

austin-denoble:
## Problem
The `pinecone-java-client` repo doesn't currently support building and pushing client reference documentation to our docs repo like the TypeScript and Python clients.

## Solution
Add new `build-and-publish-docs.yml` file to create a new workflow that allows triggering a docs build and deployment.

For now I've left this as a workflow which can be manually dispatched, or called from another workflow. We're reusing the `gradle generateJavadoc` task which was added to the `build.gradle` file previously.

## Type of Change

## Test Plan
We'll need to trigger the workflow once it's in `main`.

### DIFF
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -8,7 +8,7 @@ inputs:
   gradle-version:
     description: 'The Gradle version to use'
     required: false
-    default: '7.3'
+    default: '6.8'
 runs:
   using: 'composite'
   steps:
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -1,0 +1,27 @@
+name: 'Build client documentation'
+description: 'Generates client documentation using Javadoc'
+inputs:
+  java-version:
+    description: 'The Java version to use'
+    required: false
+    default: '8'
+  gradle-version:
+    description: 'The Gradle version to use'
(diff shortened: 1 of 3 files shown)


.github/actions/build-docs/action.yml
```diff
@@ -0,0 +1,27 @@
...
+description: 'Generates client documentation using Javadoc'
+inputs:
+  java-version:
+    description: 'The Java version to use'
+    required: false
+    default: '8'
+  gradle-version:
+    description: 'The Gradle version to use'
+    required: false
+    default: '7.3'
```
rohanshah18, line 11: If the Java version is 8, we should set gradle's default version to 6.8
https://github.com/pinecone-io/pinecone-java-client/blob/main/.github/workflows/pr.yml#L17

rohanshah18 (approved):
Looks great, left one small comment. LGTM!

haruska (approved):
LGTM

.github/actions/build-docs/action.yml
austin-denoble: What's the reason for that, is it a compatibility thing? It doesn't seem to align with the compatibility matrix here so I'm mainly curious why we need to: https://docs.gradle.org/current/userguide/compatibility.html

PR #118, opened 2024-04-30
Simplify `list` endpoint methods via Builder pattern

aulorbe:
## Problem

There are currently a lot of overloaded methods for `list`, which makes efficiently reading the code difficult. This PR refactors the code so that users can hit the `list` endpoint through a [builder](https://www.geeksforgeeks.org/builder-pattern-in-java/), which simplifies the construction of complex objects.

Asana ticket: https://app.asana.com/0/1203260648987893/1207081353236748/f 

## Type of Change

- [x] This change requires a documentation update
- [x] Non-code change (docs, etc)

## Test Plan

Unit tests, integration tests, Javadocs, and updates to the README included in PR.


src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -848,176 +848,124 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
...
+            return this;
+        }
+
+        /**
+         * Builds the {@link ListRequest} object to send to the list endpoint in order to retrieve vector IDs back from an index.
+         * @return The {@link ListenableFuture} object.
+         */
+        public ListenableFuture<ListResponse> build() {
+            // Logic to validate parameters
+            validateListEndpointParameters(namespace, prefix, paginationToken, limit,
```
austin-denoble, line 957: To be honest I'm still really confused about all the booleans in this validation function (`namespaceRequired`, `prefixRequired`, `paginationTokenRequest`, `limitRequired`). I'm just really not clear why we need to do validations this way. 

For example, I don't think we need to validate the namespace at all right? They can pass either "" for default if they want or nothing. Am I misunderstanding an issue here?

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -848,176 +848,124 @@ public ListenableFuture<DescribeIndexStatsResponse> describeIndexStats(Struct fi
...
+                    namespace != null, prefix != null, paginationToken != null, limit != null);
+
+            // Build and execute the list request
+            ListRequest.Builder builder = ListRequest.newBuilder();
+            if (namespace != null) builder.setNamespace(namespace);
+            if (prefix != null) builder.setPrefix(prefix);
+            if (paginationToken != null) builder.setPaginationToken(paginationToken);
+            if (limit != null) builder.setLimit(limit);
+            ListRequest listRequest = builder.build();
```
austin-denoble, line 966: Given that we already supposedly validated these fields, I'm not sure we actually need to be checking nulls here?

We should be able to chain things like so without caring what was passed after running the validation:
```suggestion
            ListRequest.Builder builder = ListRequest.newBuilder()
                .setNamespace(namespace)
                .setPrefix(prefix)
                .setPaginationToken(paginationToken)
                .setLimit(limit)
                .build()
```

austin-denoble (commented):
After discussing offline it seems like we probably don't want to go with a builder here. I've left some comments around the `validateListEndpointParameters` function in `IndexInterface` which I feel could maybe use some tweaking if you want to re-purpose this PR.

aulorbe:
As per IRL convo, we're going to close this PR and instead open a new PR that reduces the # of overloaded methods for the `list` endpoint.

There will now only be 3 options for users: 
- All params
- Namespace, prefix
- No params


This new PR will also refactor the `validateListEndpointParameters` method to be more straightforward

PR #127, opened 2024-05-10
Added ApiException as a cause to mapHttpStatusError

kkashkovskii:
## Problem

When the request fails before sending a request / receiving an HTTP response, the exception cause is ignored, and the client returns an empty error message:
```
Caused by: io.pinecone.exceptions.PineconeUnmappedHttpException: null
	at io.pinecone.exceptions.HttpErrorMapper.mapHttpStatusError(HttpErrorMapper.java:21)
	at io.pinecone.clients.Pinecone.handleApiException(Pinecone.java:794)
	at io.pinecone.clients.Pinecone.describeIndex(Pinecone.java:441) 
``` 

## Solution

Added the apiException as the cause HttpErrorMapper.mapHttpStatusError to make debugging easier.

## Type of Change

## Test Plan

not required


rohanshah18 (approved):
Thanks for the PR. I have run the integration tests with your changes and they ran successfully.

PR #130, opened 2024-05-13
Add additional list functions to reach parity with RESTful list requests

rasharab:
## Problem

https://github.com/pinecone-io/pinecone-java-client/issues/126

## Solution

With REST, we are currently able to list vectors without a prefix but with a pagination token.
This is not currently available with v1 of the pinecone-java-client.

With this PR, we have added a function list with a pagination token and limit but no prefix.

Issue #126, opened 2024-05-10
Additional functions for Index.list and Pinecone.Builder


rasharab:
### Is this your first time submitting a feature request?

- [X] I have searched the existing issues, and I could not find an existing issue for this feature
- [X] I am requesting a straightforward extension of existing functionality

### Describe the feature

Hi guys,

For Pinecone.Builder, I want to specify my own custom PineconeConfig with a custom managed channel:
i.e.
      val builder = Pinecone.Builder(pineconeConfig.apiKey)
      builder.setPineconeConfig(pineconeConfig)

The reason is we use customManagedChannel for the PineconeConfig with some additional configuration.

----------

For Index.list api, we want to call list **with** a paginationToken, but **not** a prefix.
That's currently not possible from what I'm seeing in your Index.java.

```
    public ListResponse list(String namespace) {
        this.validateListEndpointParameters(namespace, (String)null, (String)null, (Integer)null, true, false, false, false);
        ListRequest listRequest = ListRequest.newBuilder().setNamespace(namespace).build();
        return this.blockingStub.list(listRequest);
    }

    public ListResponse list(String namespace, int limit) {
        this.validateListEndpointParameters(namespace, (String)null, (String)null, limit, true, false, false, true);
        ListRequest listRequest = ListRequest.newBuilder().setNamespace(namespace).setLimit(limit).build();
        return this.blockingStub.list(listRequest);
    }
```
However, it is possible if I just use your RESTful apis.

### Describe alternatives you've considered

<not applicable>

### Who will this benefit?

Those who want more control.

### Are you interested in contributing this feature?

Sure

### Anything else?

<not applicable>

rasharab:
I can submit these changes if you'd like?

rasharab:
PR to address my first point;
https://github.com/pinecone-io/pinecone-java-client/pull/129

rohanshah18:
Sorry I was out for a few days and I misled you before but for custom Managed gRPC channel, we instantiate the `PineconeConfig` class followed by `PineconeConnection` class and finally the `Index` or `AsyncIndex` class depending on the stub you'll like to use. So the code will look something like this:
```java
import io.grpc.ManagedChannel;
import io.pinecone.clients.Index;
import io.pinecone.configs.PineconeConfig;
import io.pinecone.configs.PineconeConnection;
import io.grpc.netty.GrpcSslContexts;
import io.grpc.netty.NegotiationType;
[...]

PR #133, opened 2024-05-24
Add additional list functions to reach parity with RESTful list requests #130

rasharab:
## Problem

https://github.com/pinecone-io/pinecone-java-client/issues/126

## Solution

With REST, we are currently able to list vectors without a prefix but with a pagination token.
This is not currently available with v1 of the pinecone-java-client.

With this PR, we have added a function list with a pagination token and limit but no prefix.


rohanshah18:
Code review for this PR can be found here: https://github.com/pinecone-io/pinecone-java-client/pull/130

rohanshah18 (approved):
Thank you for adding the tests, I have verified the changes by running the integration tests locally. LGTM!

PR #136, opened 2024-06-10
Add proxy configuration for OkHTTPClient and NettyChannelBuilder 

rohanshah18:
## Problem

In order to configure a proxy: 
1. For data plane operations, users have to first instantiate the PineconeConfig class followed by defining the custom NettyChannelBuilder and building the managedChannel to configure the proxy. And finally, after setting the customManagedChannel in PineconeConfig, PineConnection and Index/AsyncIndex classes can be instantiated.
2. For control plane operations, users have to define the OkHttpClient, configure the proxy, and set it to the Pinecone builder object.

## Solution

Provide ability to configure proxies using `proxyHost` and `proxyPort` for both control plane (OkHttpClient) and data plane (gRPC calls via NettyChannelBuilder) operations without having the need to instantiate all of the classes as shown in the example below.

1. Data Plane operations via NettyChannelBuilder: 

Before:
```java
import io.grpc.HttpConnectProxiedSocketAddress;
import io.grpc.ManagedChannel;
import io.grpc.ProxiedSocketAddress;
import io.grpc.ProxyDetector;
import io.pinecone.clients.Index;
import io.pinecone.configs.PineconeConfig;
import io.pinecone.configs.PineconeConnection;
import io.grpc.netty.GrpcSslContexts;
import io.grpc.netty.NegotiationType;
import io.grpc.netty.NettyChannelBuilder;

import javax.net.ssl.SSLException;
import java.net.InetSocketAddress;
import java.net.SocketAddress;
import java.util.concurrent.TimeUnit;

import java.util.Arrays;

...
        String apiKey = System.getenv("PINECONE_API_KEY");
        String proxyHost = System.getenv("PROXY_HOST");
        int proxyPort = Integer.parseInt(System.getenv("PROXY_PORT"));

        PineconeConfig config = new PineconeConfig(apiKey);
        String endpoint = System.getenv("PINECONE_HOST");
        NettyChannelBuilder builder = NettyChannelBuilder.forTarget(endpoint);

        ProxyDetector proxyDetector = new ProxyDetector() {
            @Override
            public ProxiedSocketAddress proxyFor(SocketAddress targetServerAddress) {
                SocketAddress proxyAddress = new InetSocketAddress(proxyHost, proxyPort);
                
                return HttpConnectProxiedSocketAddress.newBuilder()
                        .setTargetAddress((InetSocketAddress) targetServerAddress)
                        .setProxyAddress(proxyAddress)
                        .build();
            }
        };

        // Create custom channel
        try {
            builder = builder.overrideAuthority(endpoint)
                    .negotiationType(NegotiationType.TLS)
                    .keepAliveTimeout(5, TimeUnit.SECONDS)
                    .sslContext(GrpcSslContexts.forClient().build())
                    .proxyDetector(proxyDetector);
        } catch (SSLException e) {
            throw new PineconeException("SSL error opening gRPC channel", e);
        }

        // Build the managed channel with the configured options
        ManagedChannel channel = builder.build();
        config.setCustomManagedChannel(channel);
        PineconeConnection connection = new PineconeConnection(config);
        Index index = new Index(connection, "PINECONE_INDEX_NAME");
        // Data plane operations
        // 1. Upsert data
        System.out.println(index.upsert("v1", Arrays.asList(1F, 2F, 3F, 4F)));
        // 2. Describe index stats
        System.out.println(index.describeIndexStats());
```
After:
```java
import io.pinecone.clients.Index;
import io.pinecone.clients.Pinecone;

...
        String apiKey = System.getenv("PINECONE_API_KEY");
        String proxyHost = System.getenv("PROXY_HOST");
        int proxyPort = Integer.parseInt(System.getenv("PROXY_PORT"));

        Pinecone pinecone = new Pinecone.Builder(apiKey)
                .withProxy(proxyHost, proxyPort)
                .build();

        Index index = pinecone.getIndexConnection("PINECONE_INDEX_NAME");
        // Data plane operation routed through the proxy server
        // 1. Upsert data
        System.out.println(index.upsert("v1", Arrays.asList(1F, 2F, 3F, 4F)));
        // 2. Describe index stats
        System.out.println(index.describeIndexStats());
```

2. Control Plane operations via OkHttpClient:

Before:
```java
import io.pinecone.clients.Pinecone;
import okhttp3.OkHttpClient;
import java.net.InetSocketAddress;
import java.net.Proxy;

...
        String apiKey = System.getenv("PINECONE_API_KEY");
        String proxyHost = System.getenv("PROXY_HOST");
        int proxyPort = Integer.parseInt(System.getenv("PROXY_PORT"));

        // Instantiate OkHttpClient instance and configure the proxy
        OkHttpClient client = new OkHttpClient.Builder()
                .proxy(new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort)))
                .build();

        // Instantiate Pinecone class with the custom OkHttpClient object
        Pinecone pinecone = new Pinecone.Builder(apiKey)
                .withOkHttpClient(client)
                .build();

        // Control plane operation routed through the proxy server
        System.out.println(pinecone.describeIndex("PINECONE_INDEX"));
```

After:
```java
import io.pinecone.clients.Pinecone;

...
        String apiKey = System.getenv("PINECONE_API_KEY");
        String proxyHost = System.getenv("PROXY_HOST");
        int proxyPort = Integer.parseInt(System.getenv("PROXY_PORT"));

        Pinecone pinecone = new Pinecone.Builder(apiKey)
                .withProxy(proxyHost, proxyPort)
                .build();

        // Control plane operation routed through the proxy server
        System.out.println(pinecone.describeIndex("PINECONE_INDEX"));
```

Note:
Users need to set up certificate authorities (CAs) to establish secure connections. Certificates verify server identities and encrypt data exchanged between the SDK and servers. By focusing on proxy host and port details, the SDK simplifies network setup while ensuring security.

## Type of Change

## Test Plan

Added unit tests. Given that we had issues with adding integration tests with mitm proxy in python SDK, I'm going to skip adding mitm proxy to github CI and have instead tested locally by spinning mitm proxy and successfully ran both control and data plane operations.

### DIFF
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -810,8 +810,6 @@ public class Pinecone {
         private String sourceTag;
         private ProxyConfig proxyConfig;
         private OkHttpClient customOkHttpClient;
-        private boolean isProxyConfigured = false;
-        private boolean isCustomOkHttpClient = false;
 
         /**
          * Constructs a new {@link Builder} with the mandatory API key.
@@ -874,7 +872,6 @@ public class Pinecone {
          */
         public Builder withOkHttpClient(OkHttpClient okHttpClient) {
             this.customOkHttpClient = okHttpClient;
-            isCustomOkHttpClient = true;
             return this;
         }
 
@@ -910,7 +907,6 @@ public class Pinecone {
          */
         public Builder withProxy(String proxyHost, int proxyPort) {
             this.proxyConfig = new ProxyConfig(proxyHost, proxyPort);
-            isProxyConfigured = true;
             return this;
         }
 
@@ -927,18 +923,11 @@ public class Pinecone {
             PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig);
             config.validate();
 
-            if(isProxyConfigured && isCustomOkHttpClient) {
+            if (proxyConfig != null && customOkHttpClient != null) {
                 throw new PineconeConfigurationException("Invalid configuration: Both Custom OkHttpClient and Proxy are set. Please configure only one of these options.");
             }
 
-            OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder();
-
-            if(isProxyConfigured) {
-                Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyConfig.getHost(), proxyConfig.getPort()));
-                okHttpClientBuilder.proxy(proxy);
-            }
-
-            ApiClient apiClient = (isCustomOkHttpClient) ? new ApiClient(customOkHttpClient) : new ApiClient(okHttpClientBuilder.build());
+            ApiClient apiClient = (customOkHttpClient != null) ? new ApiClient(customOkHttpClient) : new ApiClient(buildOkHttpClient());
             apiClient.setApiKey(config.getApiKey());
             apiClient.setUserAgent(config.getUserAgent());
 
@@ -950,6 +939,15 @@ public class Pinecone {
             manageIndexesApi.setApiClient(apiClient);
 
             return new Pinecone(config, manageIndexesApi);
+        }
+
+        private OkHttpClient buildOkHttpClient() {
+            OkHttpClient.Builder builder = new OkHttpClient.Builder();
+            if(proxyConfig != null) {
+                Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyConfig.getHost(), proxyConfig.getPort()));
+                builder.proxy(proxy);
+            }
+            return builder.build();
         }
     }
 }
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -49,6 +49,10 @@ public class Pinecone {
         this.manageIndexesApi = manageIndexesApi;
     }
 
+    public PineconeConfig getConfig() {
+        return config;
+    }
+
     /**
      * Creates a new serverless index with the specified parameters.
      * <p>
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -807,8 +807,7 @@ public class Pinecone {
 
         // Optional fields
         private String sourceTag;
-        private ProxyConfig controlPlaneProxyConfig;
-        private ProxyConfig dataPlaneProxyConfig;
+        private ProxyConfig proxyConfig;
         private final OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder();
         private OkHttpClient customOkHttpClient = new OkHttpClient();
         private boolean isCustomOkHttpClient = false;
@@ -909,8 +908,7 @@ public class Pinecone {
          * @return This {@link Builder} instance for chaining method calls.
          */
         public Builder withProxy(String proxyHost, int proxyPort) {
-            this.controlPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
-            this.dataPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
+            this.proxyConfig = new ProxyConfig(proxyHost, proxyPort);
             Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort));
             okHttpClientBuilder.proxy(proxy);
             return this;
@@ -926,7 +924,7 @@ public class Pinecone {
          * @return A new {@link Pinecone} instance configured based on the builder parameters.
          */
         public Pinecone build() {
-            PineconeConfig config = new PineconeConfig(apiKey, sourceTag, controlPlaneProxyConfig, dataPlaneProxyConfig);
+            PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig);
             config.validate();
 
             ApiClient apiClient = (isCustomOkHttpClient) ? new ApiClient(customOkHttpClient) : new ApiClient(okHttpClientBuilder.build());
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -3,10 +3,7 @@ package io.pinecone.clients;
 import io.pinecone.configs.PineconeConfig;
 import io.pinecone.configs.PineconeConnection;
 import io.pinecone.configs.ProxyConfig;
-import io.pinecone.exceptions.FailedRequestInfo;
-import io.pinecone.exceptions.HttpErrorMapper;
-import io.pinecone.exceptions.PineconeException;
-import io.pinecone.exceptions.PineconeValidationException;
+import io.pinecone.exceptions.*;
 import okhttp3.OkHttpClient;
 import org.openapitools.client.ApiClient;
 import org.openapitools.client.ApiException;
@@ -49,7 +46,7 @@ public class Pinecone {
         this.manageIndexesApi = manageIndexesApi;
     }
 
-    public PineconeConfig getConfig() {
+    PineconeConfig getConfig() {
         return config;
     }
 
@@ -812,8 +809,9 @@ public class Pinecone {
         // Optional fields
         private String sourceTag;
         private ProxyConfig proxyConfig;
-        private final OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder();
-        private OkHttpClient customOkHttpClient = new OkHttpClient();
+        private OkHttpClient.Builder okHttpClientBuilder;
+        private OkHttpClient customOkHttpClient;
+        private boolean isProxyConfigured = false;
         private boolean isCustomOkHttpClient = false;
 
         /**
@@ -914,7 +912,8 @@ public class Pinecone {
         public Builder withProxy(String proxyHost, int proxyPort) {
             this.proxyConfig = new ProxyConfig(proxyHost, proxyPort);
             Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort));
-            okHttpClientBuilder.proxy(proxy);
+            okHttpClientBuilder = new OkHttpClient.Builder().proxy(proxy);
+            isProxyConfigured = true;
             return this;
         }
 
@@ -929,6 +928,9 @@ public class Pinecone {
          */
         public Pinecone build() {
             PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig);
+
+            if(isProxyConfigured && isCustomOkHttpClient)
+                throw new PineconeConfigurationException("Invalid configuration: Both Custom OkHttpClient and Proxy are set. Please configure only one of these options.");
             config.validate();
 
             ApiClient apiClient = (isCustomOkHttpClient) ? new ApiClient(customOkHttpClient) : new ApiClient(okHttpClientBuilder.build());
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -879,10 +879,10 @@ public class Pinecone {
         }
 
         /**
-         * Sets a proxy for the Pinecone client to use for control plane requests.
+         * Sets a proxy for the Pinecone client to use for control and data plane requests.
          * <p>
-         * When a proxy is configured using this method, all control plane requests made by the Pinecone client will be routed
-         * through the specified proxy server.
+         * When a proxy is configured using this method, all control and data plane requests made by the Pinecone client
+         * will be routed through the specified proxy server.
          * <p>
          * It's important to note that both proxyHost and proxyPort parameters should be provided to establish
          * the connection to the proxy server.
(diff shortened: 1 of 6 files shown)


rohanshah18:
Some considerations that went into the design:
1. If the user wants to use separate proxies for control and data plane, then they should be able to and that's why I added withDataPlaneProxy() and withControlPlaneProxy(). 
2. ProxyConfig is a separate class so all properties related to proxy configurations are in one place and is not cluttering the PineconeConfig class.

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -145,14 +149,27 @@ private void onConnectivityStateChanged() {
                 channel.getState(false), channel);
     }
 
-    public static ManagedChannel buildChannel(String host) {
+    private ManagedChannel buildChannel(String host) {
```
rohanshah18, line 152: buildChannel is used within this class only so made it private. The method need not be static and the reason I had to remove the static keyword is because I wanted to use the PineconeConfig config which is private final. Also the reason why PineconeConfig can't be static is because not all connections will share the same config object.

src/test/java/io/pinecone/PineconeBuilderTest.java
```diff
@@ -106,6 +106,6 @@ public void PineconeWithSourceTag()  throws IOException {
 
         assertEquals(expectedIndex, index);
         verify(mockClient, times(1)).newCall(requestCaptor.capture());
-        assertEquals("lang=java; pineconeClientVersion=" + pineconeClientVersion + "; source_tag=testsourcetag", requestCaptor.getValue().header("User-Agent"));
+        assertEquals("lang=java; pineconeClientVersion=" + pineconeClientVersion + "; source_tag=testSourceTag", requestCaptor.getValue().header("User-Agent"));
```
rohanshah18, line 109: Not related to proxy but for some reason, the `testsourcetag` value was accidentally set to lowercase instead of camelcase.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -867,7 +873,73 @@ public Builder withSourceTag(String sourceTag) {
...
+         * @param proxyHost The hostname or IP address of the proxy server. Must not be null.
+         * @param proxyPort The port number of the proxy server. Must not be null.
+         * @return This {@link Builder} instance for chaining method calls.
+         */
+        public Builder withControlPlaneProxy(String proxyHost, int proxyPort) {
+            this.controlPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
+            Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort));
+            okHttpClientBuilder.proxy(proxy);
+            return this;
+        }
```
ssmith-pc, line 912: I'm wondering how this approach of adding named proxy-overrides (e.g. "control plane proxy", "data plane proxy", etc.) scales as we continue to add more API surface e.g. Inference API is coming soon, and probably Knowledge Models too when it makes sense. 

Instead, I'm wondering if it makes sense to keep it simple:

```java
        Pinecone pinecone = new Pinecone.Builder(apiKey)
                .withProxy(proxyHost, proxyPort)
                .build();

        System.out.println(pinecone.describeIndex("PINECONE_INDEX"));
```

and if some really needs to set a different proxy for control plane vs data plane then they would do it by instantiating a different client instance:

```java
        Pinecone controlPinecone = new Pinecone.Builder(apiKey)
                .withProxy(controlProxyHost, controlProxyPort)
                .build();

        Pinecone dataPinecone = new Pinecone.Builder(apiKey)
                .withProxy(dataProxyHost, dataProxyPort)
                .build();
```

ssmith-pc (commented):
I haven't fully reviewed the changes but have a high level UX question first

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: That's a great point! I started off with .withProxy(String host, String port) as well but given the control and data plane rely on OkHttpClient and NettyChannelBuilder respectively, we must pass in two extra flags to identify which operation needs the proxy to be configured with. We could have used enum but I was under the impression that enum can only allow us to choose one value at any time (which is still correct) but I just learned about varargs so my proposed design is now:


```java
public Builder withProxy(String proxyHost, int proxyPort, ProxyType... proxyTypes) {
    for (ProxyType proxyType : proxyTypes) {
        switch (proxyType) {
            case CONTROL_PLANE:
                this.controlPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort));
                okHttpClientBuilder.proxy(proxy);
                break;
            case DATA_PLANE:
                this.dataPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                break;
            case INFERENCE_API:
                this.inferenceApiProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                // Additional configuration for inference API if necessary
                break;
            case KNOWLEDGE_MODEL:
                this.knowledgeModelProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                // Additional configuration for knowledge model if necessary
                break;
        }
    }
    return this;
}
```


So the usage will look like this:

```java
// Configure control plane only
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort, ProxyType.CONTROL_PLANE)
        .build();

// Configure data plane only
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort, ProxyType.DATA_PLANE)
        .build();

// Configure both control and data plane
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort, ProxyType.CONTROL_PLANE, ProxyType.DATA_PLANE)
        .build();

// Configure neither control or data plane
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort)
        .build();
```

Although I'm not yet convinced with the enum name ProxyType, given it can mean actual proxy types. Thoughts?

rohanshah18: That's a great point! I started off with .withProxy(String host, String port) as well but given the control and data plane rely on OkHttpClient and NettyChannelBuilder respectively, we must pass in two extra flags to identify which operation needs the proxy to be configured with. We could have used enum but I was under the impression that enum can only allow us to choose one value at any time (which is still correct) but I just learned about varargs so my proposed design is now (added placeholder for Inference api and knowledge model in the sample below but won't be in the codebase):


```java
public Builder withProxy(String proxyHost, int proxyPort, OperationType... operationTypes) {
    for (OperationType operationType : operationTypes) {
        switch (operationType) {
            case CONTROL_PLANE:
                this.controlPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyHost, proxyPort));
                okHttpClientBuilder.proxy(proxy);
                break;
            case DATA_PLANE:
                this.dataPlaneProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                break;
            case INFERENCE_API:
                this.inferenceApiProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                break;
            case KNOWLEDGE_MODEL:
                this.knowledgeModelProxyConfig = new ProxyConfig(proxyHost, proxyPort);
                break;
        }
    }
    return this;
}
```


So the usage will look like this:

```java
// Configure control plane only
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort, OperationType.CONTROL_PLANE)
        .build();

// Configure data plane only
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort, OperationType.DATA_PLANE)
        .build();

// Configure both control and data plane
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort, OperationType.CONTROL_PLANE, OperationType.DATA_PLANE)
        .build();

// Configure neither control or data plane
Pinecone pinecone = new Pinecone.Builder(apiKey)
        .withProxy(proxyHost, proxyPort)
        .build();
```

```java
public enum OperationType {
    CONTROL_PLANE,
    DATA_PLANE
    // KNOWLEDGE_MODEL
    // INFERENCE_API
}
```

ssmith-pc: Could we just optimistically configure both Ok client and Netty Bldr with the proxy config? 

I still feel like the different (operation types) X (proxy configs) could be hard to reason about especially as the API surface areas grow.

rohanshah18: After discussing internally, we have decided to go with @ssmith-pc's approach of configuring both OkHttpClient and NettyChannelBuilder if .withProxy() is called instead of providing .withControlPlaneProxy and .withDataPlaneProxy since end user can instantiate Pinecone client twice with .withProxy(), one for control plane proxy url and the other for data plane proxy url if it comes to the point where the user needs to have two separate proxies for control and data plane operations.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -46,6 +49,10 @@ public class Pinecone {
         this.manageIndexesApi = manageIndexesApi;
     }
 
+    public PineconeConfig getConfig() {
```
rohanshah18, line 52: Added getConfig for testing purpose

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -881,13 +928,12 @@ public Builder withOkHttpClient(OkHttpClient okHttpClient) {
          * @return A new {@link Pinecone} instance configured based on the builder parameters.
          */
         public Pinecone build() {
-            PineconeConfig clientConfig = new PineconeConfig(apiKey);
-            clientConfig.setSourceTag(sourceTag);
-            clientConfig.validate();
+            PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig);
```
rohanshah18, line 931: We dont have to explicitly call .setSourceTag() if the sourceTag is part of the constructor

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -881,13 +928,12 @@ public Builder withOkHttpClient(OkHttpClient okHttpClient) {
...
+            config.validate();
 
-            ApiClient apiClient = new ApiClient(okHttpClient);
-            apiClient.setApiKey(clientConfig.getApiKey());
-            apiClient.setUserAgent(clientConfig.getUserAgent());
+            ApiClient apiClient = (isCustomOkHttpClient) ? new ApiClient(customOkHttpClient) : new ApiClient(okHttpClientBuilder.build());
```
rohanshah18, line 934: renamed clientConfig to config since there's no concept of clientConfig in v1

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -102,21 +106,6 @@ private VectorServiceGrpc.VectorServiceFutureStub generateAsyncStub(Metadata met
                 .withMaxOutboundMessageSize(DEFAULT_MAX_MESSAGE_SIZE);
     }
 
-    /**
```
rohanshah18, line 105: Moved this function to the end of the class for readability

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: If just for testing suggest making it package-private

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -805,7 +811,10 @@ public static class Builder {
 
         // Optional fields
         private String sourceTag;
-        private OkHttpClient okHttpClient = new OkHttpClient();
+        private ProxyConfig proxyConfig;
+        private final OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder();
+        private OkHttpClient customOkHttpClient = new OkHttpClient();
```
ssmith-pc, line 816: Why do we need to differentiate building an Ok client internally vs a custom one? 

At the very least it feels a little awkward to me to pre-instantiate both options.

ssmith-pc: If somebody did this they'd get back unexpected behavior compared to the alternate order. Seems like it should be an error -- you shouldn't be able to set an Ok client twice.

```java
Pinecone pc = new Pinecone.Builder()
    .withProxy(...)
    .withOkHttpClient(...)
    .build();
```

rohanshah18: 1. We should not instantiate the Builder and the customOkHttpClient.
2. You are right, if the order of operations is .withProxy(...).withOkHttpClient(...), then the proxy config will be ignored and the custom OkHttpClient will be set. My intention was to prevent user from setting both the custom client and configuring proxy so what I can do is throw an error if both are called. Thoughts?

ssmith-pc: yeah, that would be preferable I think vs. having them get unexpected behavior at runtime

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: Do users ever construct PineconeConfig directly? I see there are some constructor overloads

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -805,7 +808,10 @@ public static class Builder {
 
         // Optional fields
         private String sourceTag;
-        private OkHttpClient okHttpClient = new OkHttpClient();
+        private ProxyConfig proxyConfig;
+        private OkHttpClient customOkHttpClient;
+        private boolean isProxyConfigured = false;
+        private boolean isCustomOkHttpClient = false;
```
ssmith-pc, line 814: Why do we need extra flags vs. just checking if the members are null/non-null?

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -881,13 +924,23 @@ public Builder withOkHttpClient(OkHttpClient okHttpClient) {
...
+
+            if(isProxyConfigured) {
+                Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyConfig.getHost(), proxyConfig.getPort()));
+                okHttpClientBuilder.proxy(proxy);
+            }
 
-            ApiClient apiClient = new ApiClient(okHttpClient);
-            apiClient.setApiKey(clientConfig.getApiKey());
-            apiClient.setUserAgent(clientConfig.getUserAgent());
+            ApiClient apiClient = (isCustomOkHttpClient) ? new ApiClient(customOkHttpClient) : new ApiClient(okHttpClientBuilder.build());
```
ssmith-pc, line 941: This feels a little awkward to me, how it still creates the OkClient even if there is a custom client. Is this any better?

```java
        private OkHttpClient buildOkHttpClient() {
            OkHttpClient.Builder builder = new OkHttpClient.Builder();
            if(isProxyConfigured) {
                Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyConfig.getHost(), proxyConfig.getPort()));
                builder.proxy(proxy);
            }
            return builder.build();
        }

        public Pinecone build() {
            PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig);
            config.validate();

            if(isProxyConfigured && isCustomOkHttpClient) {
                throw new PineconeConfigurationException("Invalid configuration: Both Custom OkHttpClient and Proxy are set. Please configure only one of these options.");
            }

            ApiClient apiClient = (isCustomOkHttpClient) ? new ApiClient(customOkHttpClient) : new ApiClient(buildOkHttpClient());
     
...........
```

rohanshah18: Yeah makes sense, so let me know what you think now. The only change is the name of the method and removing the check on if proxy is configured or not, since the proxy has to be configured if the method is called.
```java
        private OkHttpClient buildProxyConfiguredOkHttpClient() {
            OkHttpClient.Builder builder = new OkHttpClient.Builder();
            Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(proxyConfig.getHost(), proxyConfig.getPort()));
            builder.proxy(proxy);
            return builder.build();
        }
```

rohanshah18: yeah makes sense

ssmith-pc (approved):

PR #138, opened 2024-06-26
Fix useragent for grpc

rohanshah18:
## Problem

The user agent string was defaulting to netty-java-grpc/1.60.2 even though the user agent + source tag was passed in.

## Solution

Instead of passing user agent as extra headers to the stubs, set the user agent string using .userAgent() on NettyChannelBuilder object.

## Type of Change

## Test Plan

Added a screenshot of the user agent being correctly populated in datadog on the internal ticket.

### DIFF
--- a/src/main/java/io/pinecone/configs/PineconeConfig.java
+++ b/src/main/java/io/pinecone/configs/PineconeConfig.java
@@ -202,20 +202,11 @@ public class PineconeConfig {
      * @return The user agent string.
      */
     public String getUserAgent() {
-        return buildUserAgent("pineconeClientVersion");
+        return buildUserAgent();
     }
 
-    /**
-     * Builds the user agent string for the Pinecone client's gRPC requests.
-     *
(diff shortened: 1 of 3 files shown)


src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -202,20 +202,11 @@ public void validate() {
...
-        return buildUserAgent("pineconeClientVersion");
+        return buildUserAgent();
     }
 
-    /**
-     * Builds the user agent string for the Pinecone client's gRPC requests.
-     *
-     * @return The user agent string for gRPC requests.
-     */
-    public String getUserAgentGrpc() {
```
rohanshah18, line 213: removed this method since netty-java-grpc/ is already appended to the user-agent string

ssmith-pc (approved):
Love to see all that deleted code 👍

PR #139, opened 2024-06-27
Update pinecone client version to 1.2.2 and remove redundant client version declaration

rohanshah18:
## Problem

Prepare for releasing v1.2.2 and remove some redundancies when it comes to declaring pinecone client version in the config and unit tests.

## Solution

Update pineconeClientVersion to v1.2.2 and remove redundancies for updating pineconeClientVersion.

## Type of Change

- [X] None of the above: Refactoring codebase and updating client version.

## Test Plan

Ran unit tests and tested with a locally published jar.

### DIFF
--- a/src/main/java/io/pinecone/commons/Constants.java
+++ b/src/main/java/io/pinecone/commons/Constants.java
@@ -1,0 +1,5 @@
+package io.pinecone.commons;
+
+public class Constants {
+    public static final String pineconeClientVersion = "v1.2.2";
+}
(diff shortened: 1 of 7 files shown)


src/main/java/io/pinecone/commons/Constants.java
```diff
@@ -0,0 +1,5 @@
+package io.pinecone.commons;
```
austin-denoble, line 1: Nice, I like putting this in a specific place for now. 👍

austin-denoble (approved):
🚢

PR #140, opened 2024-07-02
Add codegen script and update generated code for 2024-07

jhamon:
## Problem

Need to update generated code for the upcoming 2024-07 API release.

## Solution

This PR closely follows the approach used in the python and typescript repositories.

- I create a new `codegen` directory to act as the home for everything codegen related. Inside this directory there are several things:
  - Git submodule of our apis repo
  - A `build-oas.sh` script which does everything to generate the code off the spec

The script does the following:
- Pulls in the latest from the apis repo
- Builds the api repo to produce versioned spec files
- Generates java code off the spec file.
- Copies files into their respective places in `src/`

Run the script like this, passing api version number as a positional argument:

```sh
```

## Non-generated changes

The updated spec produced slightly different generated code than in the past. In order for the project to compile, I had to make some adjustments:

- Options for metric type are now part of the `IndexModel` class instead of a standalone metric object, requiring a small refactor along the lines of `IndexMetric.COSINE.toString()` to `IndexModel.MetricEnum.COSINE.toString()` and similar for other metric types.
- The name and organization of create index spec objects has changed. Now you use:
  - `IndexSpec` instead of `CreateInexRequestSpec`.
  - `PodSpec` instead of `CreateIndexRequestSpecPod`
  - `PodSpecMetadataConfig` instead of `CreateIndexRequestSpecPodMetadataConfig`

## Type of Change

- [x] None of the above: Update generated core, adjust usage for new generated classes

## Test Plan

Code builds and tests pass


codegen/build-oas.sh
```diff
@@ -0,0 +1,86 @@
...
+	oas_file="codegen/apis/_build/${version}/${module_name}_${version}.oas.yaml"
+	
+	verify_file_exists $oas_file
+
+	# Cleanup previous build files
+	echo "Cleaning up previous build files"
+	rm -rf "${build_dir}"
+
+	# Generate client module
+	docker run --rm -v $(pwd):/workspace openapitools/openapi-generator-cli:v7.0.1 generate \
```
rohanshah18, line 59: Is there a reason for using different openapi-generator version, compared to python's client generation script?
https://github.com/pinecone-io/pinecone-python-client/pull/361/files#diff-cae10a8f6beee215d2e44a3d55d2797a4e0844ea00bcb2b40bd53a76fb345da5R71

codegen/build-oas.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
+set -eux -o pipefail
+
+version=$1 # e.g. 2024-07
+modules=("control")
```
rohanshah18, line 6: Unlike python and ts, since Java SDK is gRPC only for data plane, is that the reason we have control only mentioned in the modules?

rohanshah18:
When I ran ./codegen/build-oas.sh 2024-07 locally in your branch, I'm getting the following:
`error: No justfile found`

Is there any steps I'm missing wrt copying the justfile from apis repo?

rohanshah18 (approved):
Thanks for adding the codegen script. 
I have a few questions:
1. When I ran ./codegen/build-oas.sh 2024-07 locally by checking out your branch, I'm getting the following:
error: No justfile found
Is there any steps I'm missing wrt copying the justfile from the apis repo into pinecone-java-client repo?

2. Re: The name and organization of create index spec objects has changed. Now you use:
`IndexSpec` instead of `CreateInexRequestSpec`.
`PodSpec` instead of `CreateIndexRequestSpecPod`
`PodSpecMetadataConfig` instead of `CreateIndexRequestSpecPodMetadataConfig`
Is this expected to stay the same for future versions of code generation? This is a breaking change for customers and it is okay since we are going from v1.x.x to v2.0.0 but from v2 to v3 and onwards, it'll be nice to have a consistency.

codegen/build-oas.sh
jhamon: The generator cli docker image bakes in language-specific code generation logic so it's not surprising that we might need different versions for different projects. The python sdk is stuck on an older version (v5) because in v6 of the generator cli a lot of breaking changes were made in the python generator that we aren't ready to consume in the sdk because it would create breaking changes for our users. For Java, we don't have that same constraint and since we did previous rounds of generation with a v7 generator CLI I put the same in this script to keep the code diff in these generated files to a minimum.

codegen/build-oas.sh
jhamon: Yes, we don't need the rest actions for java. But I still left this basic structure in place because I think in the future we could easily end up with multiple things here. For example, if we eventually expose management plane operations.

jhamon:
@rohanshah18 You may need to run `git submodule init && git submodule update` before running the script. This script depends on finding the contents of the apis repo (including the justfile) in a submodule located at `codegen/apis` but the error message suggests the submodule hasn't been initialized yet. You can check this by seeing what files are inside `ls codegen/apis`. If it's empty, you have some submodule issue. If you confirm the commands above work, we should probably capture them in a readme or something. Otherwise, more troubleshooting will be needed.

The names of generated objects are a direct consequence of the way things are named and structured in the spec files. A lot of small details shifted in the migration over to the apis repo, but I hope they will be more stable over time. At least the new names are better / shorter in my opinion.

rohanshah18:
After @jhamon's latest commit, the script runs successfully with the following command
`./codegen/build-oas.sh 2024-07`

PR #141, opened 2024-07-16
Add support for deletion protection

rohanshah18:
## Problem

Add support for deletion protection.

## Solution

Generated code using the following commands: 
```
```
Next, as a part of this PR, I have: 
1. Added deletionProtection enum as an argument to createPodsIndex() and createServerlessIndex()
2. Renamed configureIndex() to configurePodsIndex() and added deletionProtection enum as an argument.
3. Added configureServerlessIndex() that accepts deletionProtection enum only.

## Type of Change

## Test Plan

Added integration tests


src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -466,18 +505,15 @@ public IndexModel describeIndex(String indexName) throws PineconeException {
      * @param indexName The name of the index to configure.
      * @param podType The new podType for the index. Can be null if not changing the pod type.
      * @param replicas The desired number of replicas for the index, lowest value is 0. Can be null if not changing the number of replicas.
+     * @param deletionProtection Enable or disable deletion protection for the index.
      * @return {@link IndexModel} representing the configured index.
      * @throws PineconeException if an error occurs during the operation, the index does not exist, or if any of the arguments are invalid.
      */
-    public IndexModel configureIndex(String indexName, String podType, Integer replicas) throws PineconeException {
+    public IndexModel configurePodsIndex(String indexName, String podType, Integer replicas, DeletionProtection deletionProtection) throws PineconeException {
```
jhamon, line 512: The deletion protection feature works on both serverless and pod indexes. So it seems like an additional configure method is needed for serverless.

rohanshah18: Added here: https://github.com/pinecone-io/pinecone-java-client/pull/141/files#diff-266ed57e8c66ca7cb9b5bb7c97c25d293555998fbc6c8d50f6f70c80da53c4e3R623

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -531,16 +568,73 @@ public IndexModel configureIndex(String indexName, Integer replicas) throws Pine
...
      *
      * @param indexName The name of the index.
      * @param podType The new podType for the index.
      * @return {@link IndexModel} of the configured index.
      * @throws PineconeException if an error occurs during the operation, the index does not exist, or if the podType is invalid.
      */
-    public IndexModel configureIndex(String indexName, String podType) throws PineconeException {
-        return configureIndex(indexName, podType, null);
+    public IndexModel configurePodsIndex(String indexName, String podType) throws PineconeException {
+        return configurePodsIndex(indexName, podType, null, DeletionProtection.DISABLED);
```
jhamon, line 580: I don't think this is good. If you have an index with deletion protection enabled, we shouldn't switch it off unless you explicitly request to change it or else people will be surprised later if their data is deleted when they thought they were being protected. This implementation would result in calling `configurePodsIndex("my-index", "p2.x1")` switching deletion protection to disabled even though the user didn't request that. 

The request that is made under the hood is a PATCH request and it's not required to pass this field at all unless you want to change it. Not sure how the generated java classes here are setup, but if they are requiring a value then I see two options:

1. Eliminate this overload convenience. Require the user to specify deletion protection value whenever they are calling `configurePodsIndex`. 
2. Call `describeIndex` to get the current value and use that here instead of defaulting to disabled. This is what I'm doing for python because of some issues with the generated code basically making it impossible to not pass an option even though the underlying API does not require it. I think adding another round trip call to the server is okay from a performance standpoint because these are not high volume operations.

jhamon (approved):
Please look into my feedback on the configure action, I think this is pretty important to avoid surprising people by accidentally turning it off.

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Agreed, this is a great catch and I'm planning to go with option 2 of calling describeIndex to get the current value and set it for the configure index call.

austin-denoble: Props to Jen for flagging this to me as well. In TypeScript it wasn't a problem and we're able to send the request without attaching `deletionProtection` if it wasn't provided.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -64,11 +64,16 @@ PineconeConfig getConfig() {
...
      * @throws PineconeException if the API encounters an error during index creation or if any of the arguments are invalid.
      */
-    public IndexModel createServerlessIndex(String indexName, String metric, int dimension, String cloud,
-                                            String region) throws PineconeException {
+    public IndexModel createServerlessIndex(String indexName,
+                                            String metric,
+                                            int dimension,
+                                            String cloud,
+                                            String region,
+                                            DeletionProtection deletionProtection) throws PineconeException {
```
ssmith-pc, line 76: Is this the only `createServerlessIndex` method? Do we want to add an overload so it doesn't break any clients?

austin-denoble (approved):
LGTM other than Jen's feedback. Nice job getting this out the door.

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes, it is the only `createServerlessIndex()`. Sorry I'm not sure what you meant by `it doesn't break any clients`? Like what is clients?

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: I like how the TS client configureIndex takes a dictionary of options so we don't have to keep adding overloads each time there's a new config. 

Would be nice to do something similar here in Java land

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: customers who have coded to the current `createServerlessIndex` signature

rohanshah18: I see. I was thinking that since this is a major release, a breaking change is fine.

PR #144, opened 2024-07-18
Update generated code for fixing Collection Model's dimension error

rohanshah18:
## Problem

Vector dimension should be 20,000 but CollectionModel's docstring has 2,000.

## Solution

Generate code using the following commands with the updated openapi spec 2024-07.
```

```
## Type of Change

## Test Plan

NA

### DIFF
--- a/src/main/java/org/openapitools/control/client/model/CollectionModel.java
+++ b/src/main/java/org/openapitools/control/client/model/CollectionModel.java
@@ -49,7 +49,7 @@ import org.openapitools.control.client.JSON;
 /**
  * The CollectionModel describes the configuration and status of a Pinecone collection.
  */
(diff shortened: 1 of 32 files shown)


src/main/java/org/openapitools/control/client/model/CollectionModel.java
```diff
@@ -199,7 +199,7 @@ public CollectionModel dimension(Integer dimension) {
    /**
    * The dimension of the vectors stored in each record held in the collection.
    * minimum: 1
-   * maximum: 2000
+   * maximum: 20000
```
rohanshah18, line 202: 2,000 is now updated to 20,000

austin-denoble (approved):
🚢

PR #150, opened 2024-09-04
Add support for disabling TLS

rohanshah18:
## Problem

Add support for enabling/disabling TLS.

## Solution

Added enableTls boolean flag which is set to true by default but when set to false, it will be disabled.

## Type of Change

## Test Plan

Since the flag is set to true by default, the existing integration tests should work as it is.

### DIFF
--- a/src/main/java/io/pinecone/configs/PineconeConfig.java
+++ b/src/main/java/io/pinecone/configs/PineconeConfig.java
@@ -52,6 +52,7 @@ public class PineconeConfig {
     private String sourceTag;
     private ProxyConfig proxyConfig;
     private ManagedChannel customManagedChannel;
+    private boolean usePineconeLocal = false;
 
     /**
      * Constructs a {@link PineconeConfig} instance with the specified API key.
@@ -205,6 +206,14 @@ public class PineconeConfig {
      */
     public String getUserAgent() {
         return buildUserAgent();
+    }
+
+    public boolean getUsePineconeLocal() {
+        return usePineconeLocal;
+    }
+
+    public void setPineconeLocal(boolean pineconeLocal) {
+        this.usePineconeLocal = pineconeLocal;
     }
 
     private String buildUserAgent() {
--- a/src/main/java/io/pinecone/configs/PineconeConfig.java
+++ b/src/main/java/io/pinecone/configs/PineconeConfig.java
@@ -213,7 +213,7 @@ public class PineconeConfig {
      *
      * @return enableTls
      */
-    public boolean getEnableTLS() {
+    public boolean isTLSEnabled() {
         return enableTLS;
     }
 
--- a/src/main/java/io/pinecone/configs/PineconeConfig.java
+++ b/src/main/java/io/pinecone/configs/PineconeConfig.java
@@ -222,7 +222,7 @@ public class PineconeConfig {
      *
      * @param enableTLS true to enable TLS, false to disable it.
      */
-    public void setEnableTLS(boolean enableTLS) {
+    public void setTLSEnabled(boolean enableTLS) {
         this.enableTLS = enableTLS;
     }
 
--- a/src/main/java/io/pinecone/configs/PineconeConfig.java
+++ b/src/main/java/io/pinecone/configs/PineconeConfig.java
@@ -52,7 +52,7 @@ public class PineconeConfig {
     private String sourceTag;
     private ProxyConfig proxyConfig;
     private ManagedChannel customManagedChannel;
-    private boolean usePineconeLocal = false;
+    private boolean enableTLS = true;
 
     /**
      * Constructs a {@link PineconeConfig} instance with the specified API key.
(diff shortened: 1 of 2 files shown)


src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -207,6 +208,24 @@ public String getUserAgent() {
         return buildUserAgent();
     }
 
+    /**
+     * Returns true if TLS is enabled or false otherwise.
+     *
+     * @return enableTls
+     */
+    public boolean getEnableTLS() {
```
haruska, line 216: just a note on the naming. Should probably be something like "tlsEnabled"

haruska (commented):
Couple considerations:

1. Is this TLS flag needed? In the example code:

https://github.com/pinecone-io/pinecone-java-client/blob/da7ebc26ff4bf34898c35d9b74e16678635700d8/src/main/java/io/pinecone/configs/PineconeConfig.java#L30-L38

It looks like you can set `negotiationType(NegotiationType.PLAINTEXT)` to turn off TLS?

2. Is the flag really global to PineconeClient? Or is it per-connection to an index?
3. Another example had the host with a `http(s)://` prefix and `:PORT` suffix. Would the http vs https prefix on the host be enough without setting a specific flag? Or is this considered bad practice?

rohanshah18:
> Couple considerations:
> 
> 1. Is this TLS flag needed? In the example code:
> 
> https://github.com/pinecone-io/pinecone-java-client/blob/da7ebc26ff4bf34898c35d9b74e16678635700d8/src/main/java/io/pinecone/configs/PineconeConfig.java#L30-L38
> 
> It looks like you can set `negotiationType(NegotiationType.PLAINTEXT)` to turn off TLS?
> 
> 2. Is the flag really global to PineconeClient? Or is it per-connection to an index?
> 3. Another example had the host with a `http(s)://` prefix and `:PORT` suffix. Would the http vs https prefix on the host be enough without setting a specific flag? Or is this considered bad practice?

1. Yes, we set `negotiationType(NegotiationType.PLAINTEXT)` to turn off TLS. This is where the flag is used: https://github.com/pinecone-io/pinecone-java-client/pull/150/files#diff-b976ef187206eaaba2d35fc19f368cbc2674b74e86ac31b75e0e969378d4608dR142
If I were to not use the flag, how can I know whether we want to turn on or off the TLS? 
2. We have a config object which is instantiated when a pinecone class is instantiated. And the same config object is further used for Index class which is responsible for gRPC index connections.
3. I'm not sure if I understood pt#3 but I'll go ahead and explain how we handle the host. The flag is enabled by default so tls is enabled and hence when trying to establish gRPC connection, we actually format the host by ignoring http(s) here:
https://github.com/pinecone-io/pinecone-java-client/blob/da7ebc26ff4bf34898c35d9b74e16678635700d8/src/main/java/io/pinecone/configs/PineconeConnection.java#L183

src/main/java/io/pinecone/configs/PineconeConfig.java
rohanshah18: Yes, tlsEnabled makes more sense but since the rest of the properties in this class have getters and setters with getProp1(), getProp2(), etc., I went with getEnableTLS(). But if you still feel that we should change the name to tlsEnable(), I can make that change.

ssmith-pc: Most java naming guidelines would actually use `is` for the getter i.e. `isTLSEnabled`, and the `set` companion would be `setTLSEnabled`. I don't think it has to match with the other getters -- it's more important that it semantically makes sense.

If it was returning a nullable `Boolean` then getTLSEnabled would be appropriate

src/main/java/io/pinecone/configs/PineconeConnection.java
```diff
@@ -139,10 +139,17 @@ private ManagedChannel buildChannel() {
...
-            builder = builder.overrideAuthority(endpoint)
-                    .negotiationType(NegotiationType.TLS)
-                    .sslContext(GrpcSslContexts.forClient().build())
-                    .userAgent(config.getUserAgent());
+            if(config.isTLSEnabled()) {
+                builder = builder
+                        .overrideAuthority(endpoint)
+                        .negotiationType(NegotiationType.TLS)
+                        .sslContext(GrpcSslContexts.forClient().build())
+                        .userAgent(config.getUserAgent());
```
ssmith-pc, line 147: Seems like we would want to set the user agent in both cases

ssmith-pc (approved):
Looks straightforward, few minor comments

src/main/java/io/pinecone/configs/PineconeConnection.java
rohanshah18: when tls is disabled, how do we plan on receiving the user-agent in the logs? Or are you thinking of intercepting the call via proxy for debugging purpose? Like what are the other use cases we are looking at with TLS disabled and having the user-agent?

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -207,6 +208,24 @@ public String getUserAgent() {
...
+    }
+
+    /**
+     * Sets whether TLS is enabled.
+     *
+     * @param enableTLS true to enable TLS, false to disable it.
+     */
+    public void setEnableTLS(boolean enableTLS) {
+        this.enableTLS = enableTLS;
+    }
```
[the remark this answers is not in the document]
rohanshah18: Done!

PR #151, opened 2024-09-10
Draft: add embed endpoint

rohanshah18:
## Problem

Describe the purpose of this change. What problem is being solved and why?

## Solution

Describe the approach you took. Link to any relevant bugs, issues, docs, or other resources.

## Type of Change

## Test Plan

### DIFF
--- a/src/main/java/io/pinecone/clients/Inference.java
+++ b/src/main/java/io/pinecone/clients/Inference.java
@@ -7,8 +7,9 @@ import org.openapitools.control.client.model.EmbedRequestInputsInner;
 import org.openapitools.control.client.model.EmbedRequestParameters;
 import org.openapitools.control.client.model.EmbeddingsList;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 public class Inference {
     InferenceApi inferenceApi;
@@ -17,10 +18,12 @@ public class Inference {
         inferenceApi = new InferenceApi();
     }
 
-    public EmbeddingsList embed(String model, String truncate, String inputType, List<String> inputs) throws ApiException {
-        EmbedRequestParameters embedRequestParameters = new EmbedRequestParameters().inputType(inputType);
-        if(truncate != null && !truncate.isEmpty())
-            embedRequestParameters.truncate(truncate);
+    // ToDo: Add serialization unit test for checking if truncate and input_type are correctly serialized when passed via map
+    // Cond1: when input_type and truncate are set -> both should have the set values especially truncate's default value END should be overwritten
+    // Cond2: when only input_type is set, truncate should be by default set to END
+    public EmbeddingsList embed(String model, Map<String, Object> parameters, List<String> inputs) throws ApiException {
+        EmbedRequestParameters embedRequestParameters = new EmbedRequestParameters();
+        parameters.forEach(embedRequestParameters::putAdditionalProperty);
         List<EmbedRequestInputsInner> EmbedRequestInputsInnerList = convertInputStringToEmbedRequestInputsInner(inputs);
         EmbedRequest embedRequest = new EmbedRequest()
                 .model(model)
@@ -31,10 +34,8 @@ public class Inference {
     }
 
     private List<EmbedRequestInputsInner> convertInputStringToEmbedRequestInputsInner(List<String> inputs) {
-        List<EmbedRequestInputsInner> embedRequestInputsInnerList = new ArrayList<EmbedRequestInputsInner>();
-        for(String input:inputs) {
-            embedRequestInputsInnerList.add(new EmbedRequestInputsInner().text(input));
-        }
-        return embedRequestInputsInnerList;
+        return inputs.stream()
+                .map(input -> new EmbedRequestInputsInner().text(input))
+                .collect(Collectors.toList());
     }
 }
--- a/src/main/java/io/pinecone/clients/Inference.java
+++ b/src/main/java/io/pinecone/clients/Inference.java
@@ -1,0 +1,40 @@
+package io.pinecone.clients;
+
+import org.openapitools.control.client.ApiException;
+import org.openapitools.control.client.api.InferenceApi;
+import org.openapitools.control.client.model.EmbedRequest;
+import org.openapitools.control.client.model.EmbedRequestInputsInner;
+import org.openapitools.control.client.model.EmbedRequestParameters;
+import org.openapitools.control.client.model.EmbeddingsList;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Inference {
+    InferenceApi inferenceApi;
+
+    public Inference() {
+        inferenceApi = new InferenceApi();
+    }
+
+    public EmbeddingsList embed(String model, String truncate, String inputType, List<String> inputs) throws ApiException {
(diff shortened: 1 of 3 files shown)


src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public EmbeddingsList embed(String model, String truncate, String inputType, List<String> inputs) throws ApiException {
+        EmbedRequestParameters embedRequestParameters = new EmbedRequestParameters().inputType(inputType);
+        if(truncate != null && !truncate.isEmpty())
+            embedRequestParameters.truncate(truncate);
+        List<EmbedRequestInputsInner> EmbedRequestInputsInnerList = convertInputStringToEmbedRequestInputsInner(inputs);
+        EmbedRequest embedRequest = new EmbedRequest()
+                .model(model)
+                .parameters(embedRequestParameters)
+                .inputs(EmbedRequestInputsInnerList);
```
ssmith-pc, line 28: Other models might support different so would be better as a map.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -871,6 +871,10 @@ public AsyncIndex getAsyncIndexConnection(String indexName) throws PineconeValid
         return new AsyncIndex(connection, indexName);
     }
 
+    public Inference getInference() {
+        return new Inference();
+    }
```
ssmith-pc, line 876: How heavy is it to create this object, or should we cache Inference in Pinecone? 

e.g. if someone effectively did this every time:

```java
pc.getInference().embed(...);
pc.getInference().embed(...);
pc.getInference().embed(...);
pc.getInference().embed(...);
```

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -871,6 +871,10 @@ public AsyncIndex getAsyncIndexConnection(String indexName) throws PineconeValid
         return new AsyncIndex(connection, indexName);
     }
 
+    public Inference getInference() {
```
ssmith-pc, line 874: This feels a little odd to me as a getter. Wondering if just `pc.inference().embed(...)` is better?

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,40 @@
...
+        return inferenceApi.embed(embedRequest);
+    }
+
+    private List<EmbedRequestInputsInner> convertInputStringToEmbedRequestInputsInner(List<String> inputs) {
+        List<EmbedRequestInputsInner> embedRequestInputsInnerList = new ArrayList<EmbedRequestInputsInner>();
+        for(String input:inputs) {
+            embedRequestInputsInnerList.add(new EmbedRequestInputsInner().text(input));
+        }
+        return embedRequestInputsInnerList;
+    }
```
ssmith-pc, line 39: This would be better using stream API: 

```java
inputs.stream()
                 .map(input -> new EmbedRequestInputsInner().text(input))
                 .collect(Collectors.toList());
```

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Can you please elaborate on this more? Are you suggesting `List<String> inputs` should be a `map` instead? If so, how does the key value pair be assigned to the `inputs` since `inputs` is of type `List<EmbedRequestInputsInner>` where `EmbedRequestInputsInner` has only one member which is `text` of type String.

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes, I like pc.inference().embed(...) more

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Why do you think using stream is better here? Wouldn't it create a new object for intermediate step?

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Would you please elaborate on this, I didn't understand your question here. Are you concerned that if the user, instead of using the same object, instantiates a new object for Inference class and will result in multiple objects in heap memory instead of referencing to one single object in the heap memory?

rohanshah18: Yeah I see your point! Do you feel like java users are traditionally used to getting the Inference object and reusing it rather than instantiating every single time to call the embed()?

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: I think [pc.inference().embed(...) definitely feels more like python usage and it aligns with rest of our SDKs but given its java, do you feel like users will directly call pc.inference().embed()? And pc.inference() will still return the Inference object regardless, so the naming does make sense if not compared with other SDKs.

rohanshah18: Also for dataplane opeartions, we have getIndexConnection() and getAsyncIndexConnection().

src/main/java/io/pinecone/clients/Inference.java
ssmith-pc: I mean for the params like truncate and input_type -- those apply to the multilingual-e5-large model, but they might not apply to the next embedding model we may add. 

You can use the Python impl as a reference, where it takes params as basically an optional map (I know, Java optionals are not a thing)

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: I don't think there's a java convention for this, so just thinking about the ergonomics of how we deliver the API to them, and if there's something here that's heavy to construct/connect/etc. then we should try to lazy construct that and reuse across requests

src/main/java/io/pinecone/clients/Inference.java
ssmith-pc: I'd have to look more closely at the underlying impl, but I think using the stream impl gives java flexibility to optimize how/when the underlying list actually gets created, and it may never have to allocate a new list at all depending on the usage.

src/main/java/io/pinecone/clients/Pinecone.java
ssmith-pc: Maybe calling it `getInferenceClient()` would resolve the ambiguity and encourage use like:

```java
client = pc.getInferenceClient();
client.embed(...);
client.embed(...);
```

ssmith-pc: I'm not worried about the extra objects if they're lightweight to construct. So I'm asking  how expensive are they to construct? 

> How heavy is it to create this object, or should we cache Inference in Pinecone?

src/main/java/io/pinecone/clients/Inference.java
ssmith-pc: This is your call, and don't think it'll make much difference in how fast things run on such small datasets. I just find the stream api more pleasant to use, and gives java an opportunity to optimize what's happening under the hood more so than if you explicitly instantiate a list. But this is not a showstopper by any means

PR #152, opened 2024-09-18
Prep for release v2.1.0 and add proxy config & disabling TLS examples to README

rohanshah18:
## Problem

- Prepare for releasing v2.1.0.
- Added support proxy configuration and for disabling TLS, so add examples to the README.

## Solution

- Update pinecone client version to `v2.1.0`.
- Add proxy config and disabling TLS examples to README

## Type of Change

## Test Plan

NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,10 @@ public class InitializeClientExample {
 }
 ```
 
-#### Passing OkHttpClient
+#### Passing OkHttpClient for control plane operations
+If you need to provide a custom `OkHttpClient`, you can do so by using the `withOkHttpClient()` method of the 
+`Pinecone.Builder` class to pass in your `OkHttpClient` object.
+
 ```java
 import io.pinecone.clients.Pinecone;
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Maven:
 
 Gradle:
 ```
-implementation "io.pinecone:pinecone-client:1.2.2"
+implementation "io.pinecone:pinecone-client:2.0.0"
```
austin-denoble, line 26: Should this be `2.1.0`?

README.md
```diff
@@ -152,7 +220,6 @@ IndexModel indexModel = pinecone.createPodsIndex(indexName, dimension, environme
 
 
 ## List indexes
-
 The following example returns all indexes (and their corresponding metadata) in your project.
```
austin-denoble, line 223: **nit:** I think it's generally considered best practice leave line breaks on either side of headings in markdown. My IDE has some formatting that automatically applies these when I'm working in these files: https://www.markdownguide.org/basic-syntax/#heading-best-practices

austin-denoble (approved):
LGTM, seems like there's a version in an example in the README which may need to be tweaked.

README.md
rohanshah18: Ahh interesting, let me add new lines

README.md
rohanshah18: So this was a typo from the last time and going forward, I'm planning to follow python SDKs format where we only update the versions in the README for major releases. Also I'll be adding a CI job to automate publishing.

PR #153, opened 2024-09-20
Add embed endpoint

rohanshah18:
## Problem

Add embed endpoint for Java SDK to allow users to create embeddings for text data such as passage or query using a specified model. More details on inference api can be found [here](https://docs.pinecone.io/guides/inference/understanding-inference).

## Solution

Users can now call embed endpoint with the following parameters:
1. `String model`: Accepts a string from the specified [models](https://docs.pinecone.io/models/overview) .
2. `Map<String, Object> parameters`: Accepts `input_type` and `truncate` as keys with their corresponding values in a Map. The values are expected to be scalar. Please note that the default value of `truncate` is set to `END` if not specified.
3. `List<String> inputs`: The list must be of size atleast 1.

As a part of this change, I have added a `getInferenceClient()` in `Pinecone` class to follow a similar pattern with other SDKs. The underlying client uses `OkHTTPClient` for REST calls. The method `getInferenceClient()` returns an instance of `Inference` class which I added as a wrapper to the Inference API. So far, this wrapper contains `embed()` endpoint only, with the plan of adding `rerank()` soon. Lastly, I have added docstrings and updated README with an example of the embed endpoint.

## Type of Change

## Test Plan

Added integration tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,44 +550,39 @@ pinecone.deleteCollection("example-collection");
 The Pinecone SDK now supports creating embeddings via the [Inference API](https://docs.pinecone.io/guides/inference/understanding-inference).
 
 ```java
-import io.pinecone.clients.Inference;
 import io.pinecone.clients.Pinecone;
 import org.openapitools.control.client.ApiException;
 import org.openapitools.control.client.model.Embedding;
 import org.openapitools.control.client.model.EmbeddingsList;
-import org.openapitools.control.client.model.EmbeddingsListUsage;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 ...
-        
+
 Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
 Inference inference = pinecone.getInferenceClient();
 
-List<String> inputs = new ArrayList<>(1);
+// Prepare input sentences to be embedded
+List<String> inputs = new ArrayList<>();
 inputs.add("The quick brown fox jumps over the lazy dog.");
 inputs.add("Lorem ipsum");
 
+// Specify the embedding model and parameters
 String embeddingModel = "multilingual-e5-large";
 
 Map<String, Object> parameters = new HashMap<>();
 parameters.put("input_type", "query");
 parameters.put("truncate", "END");
 
+// Generate embeddings for the input data
 EmbeddingsList embeddings = inference.embed(embeddingModel, parameters, inputs);
-// Get model
-String embeddingsModel = embeddings.getModel();
 
 // Get embedded data
 List<Embedding> embeddedData = embeddings.getData();
-
-// Get total usage tokens
-int tokens = embeddings.getUsage().getTotalTokens();
 ```
 
 ## Examples
 
 - The data and control plane operation examples can be found in `io/pinecone/integration` folder.
-- A full end-to-end Semantic Search example can be found in the [Java Examples](https://github.com/pinecone-io/java-examples/tree/main) repo on Github.
--- a/README.md
+++ b/README.md
@@ -47,8 +47,11 @@ The `Pinecone` class is your main entry point into the Pinecone Java SDK. You ca
 your `apiKey`, either by passing it as an argument in your code or by setting it as an environment variable called 
 `PINECONE_API_KEY`.
 
-Note: for pod-based indexes, you will also need an `environment` variable. You can set pass this as an argument in 
-your code or set it as an environment variable called `PINECONE_ENVIRONMENT`.
+This internally instantiates a single shared OkHttpClient instance, which is used for both control plane and inference
+operations. Note that the OkHttpClient performs best when you create a single `OkHttpClient` instance and reuse it 
+for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing 
+connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes 
+resources on idle pools. More details on the OkHttpClient can be found [here](https://github.com/square/okhttp/blob/f2771425cb714a5b0b27238bd081b2516b4d640f/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt#L54).
 
 ```java
 import io.pinecone.clients.Pinecone;
@@ -550,7 +553,9 @@ The Pinecone SDK now supports creating embeddings via the [Inference API](https:
 import io.pinecone.clients.Inference;
 import io.pinecone.clients.Pinecone;
 import org.openapitools.control.client.ApiException;
+import org.openapitools.control.client.model.Embedding;
 import org.openapitools.control.client.model.EmbeddingsList;
+import org.openapitools.control.client.model.EmbeddingsListUsage;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -572,6 +577,14 @@ parameters.put("input_type", "query");
 parameters.put("truncate", "END");
 
 EmbeddingsList embeddings = inference.embed(embeddingModel, parameters, inputs);
+// Get model
+String embeddingsModel = embeddings.getModel();
+
+// Get embedded data
+List<Embedding> embeddedData = embeddings.getData();
+
+// Get total usage tokens
+int tokens = embeddings.getUsage().getTotalTokens();
 ```
 
 ## Examples
--- a/README.md
+++ b/README.md
@@ -542,6 +542,38 @@ Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
 pinecone.deleteCollection("example-collection");
 ```
 
+## Inference
+
+The Pinecone SDK now supports creating embeddings via the [Inference API](https://docs.pinecone.io/guides/inference/understanding-inference).
+
+```java
+import io.pinecone.clients.Inference;
+import io.pinecone.clients.Pinecone;
+import org.openapitools.control.client.ApiException;
+import org.openapitools.control.client.model.EmbeddingsList;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+...
+        
+Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
+Inference inference = pinecone.getInferenceClient();
+
+List<String> inputs = new ArrayList<>(1);
+inputs.add("The quick brown fox jumps over the lazy dog.");
+inputs.add("Lorem ipsum");
+
+String embeddingModel = "multilingual-e5-large";
+
+Map<String, Object> parameters = new HashMap<>();
+parameters.put("input_type", "query");
+parameters.put("truncate", "END");
+
+EmbeddingsList embeddings = inference.embed(embeddingModel, parameters, inputs);
+```
+
 ## Examples
 
 - The data and control plane operation examples can be found in `io/pinecone/integration` folder.
--- a/src/main/java/io/pinecone/clients/Inference.java
+++ b/src/main/java/io/pinecone/clients/Inference.java
@@ -12,14 +12,61 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+/**
+ * The Inference class provides methods to interact with Pinecone's embedding API through the Java SDK. It allows users
+ * to send input data to generate embeddings using a specified model.
+ * <p>
+ * This class utilizes the {@link InferenceApi} to make API calls to the Pinecone embedding service.
+ *
+ * <pre>{@code
+ *      import io.pinecone.clients.Inference;
+ *      import io.pinecone.clients.Pinecone;
+ *      import org.openapitools.control.client.ApiException;
+ *      import org.openapitools.control.client.model.EmbeddingsList;
+ *
+ *      import java.util.ArrayList;
+ *      import java.util.HashMap;
+ *      import java.util.List;
+ *      import java.util.Map;
+ *
+ *     // Prepare input sentences to be embedded
+ *     List<String> inputs = new ArrayList<>(1);
+ *     inputs.add("The quick brown fox jumps over the lazy dog.");
+ *     inputs.add("Lorem ipsum");
+ *
+ *     // Specify the embedding model and parameters
+ *     String embeddingModel = "multilingual-e5-large";
+ *
+ *     Map<String, Object> parameters = new HashMap<>();
+ *     parameters.put("input_type", "query");
+ *     parameters.put("truncate", "END");
+ *
+ *     // Generate embeddings for the input data
+ *     EmbeddingsList embeddings = inference.embed(embeddingModel, parameters, inputs);
+ * }</pre>
+ */
 public class Inference {
 
-    InferenceApi inferenceApi;
+    private final InferenceApi inferenceApi;
 
+    /**
+     * Constructs an Inference instance.
+     *
+     * @param apiClient The ApiClient object used to configure the API connection.
+     */
     public Inference(ApiClient apiClient) {
         inferenceApi = new InferenceApi(apiClient);
     }
 
+    /**
+     * Sends input data and parameters to the embedding model and returns a list of embeddings.
+     *
+     * @param model The embedding model to use.
+     * @param parameters A map containing model-specific parameters.
+     * @param inputs A list of input strings to generate embeddings for.
+     * @return EmbeddingsList containing the embeddings for the provided inputs.
+     * @throws ApiException If the API call fails, an ApiException is thrown.
+     */
     public EmbeddingsList embed(String model, Map<String, Object> parameters, List<String> inputs) throws ApiException {
         EmbedRequestParameters embedRequestParameters = new EmbedRequestParameters();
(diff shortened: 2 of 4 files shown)


src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public EmbeddingsList embed(String model, Map<String, Object> parameters, List<String> inputs) throws ApiException {
+        EmbedRequestParameters embedRequestParameters = new EmbedRequestParameters();
+        parameters.forEach(embedRequestParameters::putAdditionalProperty);
+
+        List<EmbedRequestInputsInner> EmbedRequestInputsInnerList = convertInputStringToEmbedRequestInputsInner(inputs);
+        EmbedRequest embedRequest = new EmbedRequest()
+                .model(model)
+                .parameters(embedRequestParameters)
+                .inputs(EmbedRequestInputsInnerList);
```
ssmith-pc, line 78: The casing on this variable `EmbedRequestInputsInnerList` is non-standard (begins with a capital).

I would actually just get rid of the variable altogether, and pass `convertInputsStringToEmbedRequestInputsInner(inputs)` directly to `EmbedRequest().inputs(...)`

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+        return inferenceApi.embed(embedRequest);
+    }
+
+    /**
+     * Converts a list of input strings to EmbedRequestInputsInner objects.
+     *
+     * @param inputs A list of input strings.
+     * @return A list of EmbedRequestInputsInner objects containing the input data.
+     */
+    private List<EmbedRequestInputsInner> convertInputStringToEmbedRequestInputsInner(List<String> inputs) {
```
ssmith-pc, line 89: This is a really long method name, would suggest shortening it, maybe `convertToEmbedInputs`. The parameter type and return type already make the other details clear

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -871,6 +871,18 @@ public AsyncIndex getAsyncIndexConnection(String indexName) throws PineconeValid
...
     }
 
+    /**
+     * A method to create and return a new instance of the {@link Inference} client.
+     * <p>
+     * This method initializes the Inference client using the current ApiClient
+     * from the {@link ManageIndexesApi}. The Inference client can then be used
+     * to interact with Pinecone's embedding API.
+     * @return A new {@link Inference} client instance.
+     */
```
ssmith-pc, line 881: We should include guidance about whether this instance should be cached and re-used, and whether it's thread-safe

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Sends input data and parameters to the embedding model and returns a list of embeddings.
+     *
+     * @param model The embedding model to use.
+     * @param parameters A map containing model-specific parameters.
+     * @param inputs A list of input strings to generate embeddings for.
+     * @return EmbeddingsList containing the embeddings for the provided inputs.
+     * @throws ApiException If the API call fails, an ApiException is thrown.
```
ssmith-pc, line 68: I would put the code sample on this block instead of at the class level

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+ *     List<String> inputs = new ArrayList<>(1);
+ *     inputs.add("The quick brown fox jumps over the lazy dog.");
+ *     inputs.add("Lorem ipsum");
+ *
+ *     // Specify the embedding model and parameters
+ *     String embeddingModel = "multilingual-e5-large";
+ *
+ *     Map<String, Object> parameters = new HashMap<>();
+ *     parameters.put("input_type", "query");
+ *     parameters.put("truncate", "END");
```
ssmith-pc, line 42: This syntax is a bit more concise for code samples:

```java
List<String> inputs = List.of(
    "The quick brown fox jumps over the lazy dog.",
    "Lorem ipsum"
);

Map<String, Object> parameters = Map.of(
        "input_type", "query",
        "truncate", "END"
);
```

README.md
```diff
@@ -542,6 +542,38 @@ Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
...
+inputs.add("The quick brown fox jumps over the lazy dog.");
+inputs.add("Lorem ipsum");
+
+String embeddingModel = "multilingual-e5-large";
+
+Map<String, Object> parameters = new HashMap<>();
+parameters.put("input_type", "query");
+parameters.put("truncate", "END");
+
+EmbeddingsList embeddings = inference.embed(embeddingModel, parameters, inputs);
```
ssmith-pc, line 574: How would someone work with the `EmbeddingsList` response object? Should include that in the example

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Sends input data and parameters to the embedding model and returns a list of embeddings.
+     *
+     * @param model The embedding model to use.
+     * @param parameters A map containing model-specific parameters.
+     * @param inputs A list of input strings to generate embeddings for.
+     * @return EmbeddingsList containing the embeddings for the provided inputs.
+     * @throws ApiException If the API call fails, an ApiException is thrown.
+     */
+    public EmbeddingsList embed(String model, Map<String, Object> parameters, List<String> inputs) throws ApiException {
```
ssmith-pc, line 70: Is it normal for the java client to return the openapi types (EmbeddingsList, ApiException) directly instead of some kind of wrapped object?

rohanshah18: Yes, that's the pattern we followed for the rest of the endpoints.

README.md
aulorbe: +1

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+import org.openapitools.control.client.model.EmbedRequestInputsInner;
+import org.openapitools.control.client.model.EmbedRequestParameters;
+import org.openapitools.control.client.model.EmbeddingsList;
+
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * The Inference class provides methods to interact with Pinecone's embedding API through the Java SDK. It allows users
```
aulorbe, line 16: I'd change this wording: "The Inference class provides methods to interact with Pinecone's embedding API" since won't be entirely correct once you add things like `rerank` -- the Inference class actually exposes the _Inference_ API, not the embedding API (it exposes, the embedding _endpoint_ of the Inference API). 

I'd say "The Inference class provides methods to interact with Pinecone's Inference API"

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+ * to send input data to generate embeddings using a specified model.
+ * <p>
+ * This class utilizes the {@link InferenceApi} to make API calls to the Pinecone embedding service.
```
aulorbe, line 19: I wouldn't call it an "embedding service", I'd just say "to the Inference API's `/embed` endpoint"

src/main/java/io/pinecone/clients/Inference.java
aulorbe, line 16: I'd also consider linking out to documentation in this docstring, esp for the parameters, since it's unclear if those are just example params or mandatory params etc

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+ *     // Generate embeddings for the input data
+ *     EmbeddingsList embeddings = inference.embed(embeddingModel, parameters, inputs);
+ * }</pre>
+ */
+public class Inference {
+
+    private final InferenceApi inferenceApi;
+
+    /**
+     * Constructs an Inference instance.
```
aulorbe, line 53: Nit: I'd say "Constructs an instance of the Inference class" or "...Inference API"

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+    public Inference(ApiClient apiClient) {
+        inferenceApi = new InferenceApi(apiClient);
+    }
+
+    /**
+     * Sends input data and parameters to the embedding model and returns a list of embeddings.
+     *
+     * @param model The embedding model to use.
+     * @param parameters A map containing model-specific parameters.
+     * @param inputs A list of input strings to generate embeddings for.
```
aulorbe, line 66: Nit: I'd remove "strings" here, since we may in the future embed more than just strings (e.g. images).

src/main/java/io/pinecone/clients/Inference.java
aulorbe, line 68: This might be irrelevant, but I don't entirely remember -- does the call fail if the API endpoint is down or there's something wrong with the request? Or does the call succeed, but the content of the `data` field (or whatever) is an error message?

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+        EmbedRequest embedRequest = new EmbedRequest()
+                .model(model)
+                .parameters(embedRequestParameters)
+                .inputs(EmbedRequestInputsInnerList);
+
+        return inferenceApi.embed(embedRequest);
+    }
+
+    /**
+     * Converts a list of input strings to EmbedRequestInputsInner objects.
```
aulorbe, line 84: I'd remove "strings" here, just say something like "input data" or "input objects" or "input"

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param inputs A list of input strings.
```
aulorbe, line 86: Again (strings)

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -871,6 +871,18 @@ public AsyncIndex getAsyncIndexConnection(String indexName) throws PineconeValid
         return new AsyncIndex(connection, indexName);
     }
 
+    /**
+     * A method to create and return a new instance of the {@link Inference} client.
```
aulorbe, line 875: It's still the Pinecone client, right?

For the other clients, the gateway to the Inference API is a method on the Pinecone client, so something like `pinecone.inference.embed`.

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -871,6 +871,18 @@ public AsyncIndex getAsyncIndexConnection(String indexName) throws PineconeValid
...
+     * <p>
+     * This method initializes the Inference client using the current ApiClient
+     * from the {@link ManageIndexesApi}. The Inference client can then be used
+     * to interact with Pinecone's embedding API.
```
aulorbe, line 879: Nit: it's the Inference API, not the embedding API.

aulorbe (approved):
Just some language nits :)

src/main/java/io/pinecone/clients/Inference.java
austin-denoble: We currently return and use a lot of the generated openapi types such as `IndexModel`, `IndexList`, `CollectionList`, etc: https://github.com/pinecone-io/pinecone-java-client/blob/27ea4884b27aeb81c85a8cecc0a4c2e8f288680c/src/main/java/io/pinecone/clients/Pinecone.java#L717.

I think the only custom models we have at the moment are the things defined under `/unsigned_indices_model`: https://github.com/pinecone-io/pinecone-java-client/tree/main/src/main/java/io/pinecone/unsigned_indices_model

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -0,0 +1,94 @@
...
+public class Inference {
+
+    private final InferenceApi inferenceApi;
+
+    /**
+     * Constructs an Inference instance.
+     *
+     * @param apiClient The ApiClient object used to configure the API connection.
+     */
+    public Inference(ApiClient apiClient) {
```
austin-denoble, line 57: Do we need to be checking for a possible `null` here?

austin-denoble (commented):
Overally looks good to me, I agree with the feedback @ssmith-pc left as well.

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Fixed!

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Added the instructions in the README and in the Pinecone class.

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Fixed!

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Fixed!

README.md
rohanshah18: Added the getters method!

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Yes, fixed!

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Fixed!

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: This is supposed to take in List<String>.

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes, it is pinecone client which has getInferenceClient().

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: The call will fail if its 5xx

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: No, since the ApiClient is always instantiated as a part of instantiating Pinecone class

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes, needs to be updated!

austin-denoble (approved):
LGTM, thanks for iterating.

README.md
ssmith-pc: Can we go a step further and show how it could be used with the upsert or query APIs?

ssmith-pc: Also, I think only the `getData()` method is worth showing in the example -- the getModel and getUsage are not as important to showcase and are easier for folks to figure out

src/main/java/io/pinecone/clients/Inference.java
ssmith-pc: ok cool

README.md
```diff
@@ -542,6 +545,48 @@ Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
...
+List<String> inputs = new ArrayList<>(1);
+inputs.add("The quick brown fox jumps over the lazy dog.");
+inputs.add("Lorem ipsum");
+
+String embeddingModel = "multilingual-e5-large";
+
+Map<String, Object> parameters = new HashMap<>();
+parameters.put("input_type", "query");
+parameters.put("truncate", "END");
+
```
ssmith-pc, line 578: If we keep code samples in both places, I would update them so they match.

ssmith-pc (approved):
I left a couple more comments related to code samples, but overall looks good to me. Thanks for addressing all the feedback.

src/main/java/io/pinecone/clients/Inference.java
rohanshah18: Wait I just realized Java 8 doesn't support this, its from Java 9 and later. Will revert this change.

README.md
rohanshah18: > Can we go a step further and show how it could be used with the upsert or query APIs?
Yes, I'll add a SampleApp file in the /examples folder which will have an end to end example i.e. from generating embeddings to upserting and querying.

src/main/java/io/pinecone/clients/Inference.java
ssmith-pc: It's just sample code though?

rohanshah18: Yes but since the SDK supports Java 8+, I would prefer all functionalities that are supported in Java 8 as well. Also its just the syntax right? Like the there will be one object instantiated regardless of the approach?

PR #158, opened 2024-10-18
Generate code based on 2024-10 api spec

rohanshah18:
## Problem

Generate code based on 2024-10 api spec.

## Solution
OpenAPI based code was regenerated using the 2024-10 api sec which resulted in the following changes:
1. Updated `build-oas.sh` to include all control plane, data plane, and inference modules.
2. Inference is now broken out of control plane so the underlying `inference` code generated by openAPI spec is now updated. This also resulted in instantiating a separate `ApiClient` which is a part of inference module now instead of the shared `ApiClient` class of control plane. Also as a part of this refactoring, the `customOkHttpClient` is now a part of the `PineconeConfig` class, so once the `customOkHttpClient` is set by the user at the time of instantiating `Pinecone` using its builder class, it'll be stored in the `PineconeConfig` instance which can further be leveraged by the `Inference` api calls as well. 
3. Control and data plane modules are now prepended with `db_` so this resulted in updating import statements for a lot of existing classes.

## Type of Change

## Test Plan

Since no new functionality is added, the existing integration tests should run fine.


src/integration/java/io/pinecone/helpers/TestResourcesManager.java
```diff
@@ -5,7 +5,7 @@
 import io.pinecone.clients.Pinecone;
 import io.pinecone.exceptions.PineconeException;
 import io.pinecone.proto.DescribeIndexStatsResponse;
-import org.openapitools.control.client.model.*;
+import org.openapitools.db_control.client.model.*;
```
rohanshah18, line 8: example of updating import statement since `db_` is now appended to control and data modules.

src/main/java/io/pinecone/clients/Inference.java
```diff
@@ -1,20 +1,21 @@
...
-import org.openapitools.control.client.ApiClient;
-import org.openapitools.control.client.ApiException;
-import org.openapitools.control.client.api.InferenceApi;
-import org.openapitools.control.client.model.EmbedRequest;
-import org.openapitools.control.client.model.EmbedRequestInputsInner;
-import org.openapitools.control.client.model.EmbedRequestParameters;
-import org.openapitools.control.client.model.EmbeddingsList;
+import io.pinecone.configs.PineconeConfig;
+import okhttp3.OkHttpClient;
+import org.openapitools.inference.client.ApiClient;
```
rohanshah18, line 5: example of updating import statement since inference is broken out of control module

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -1033,14 +1033,14 @@ public Builder withProxy(String proxyHost, int proxyPort) {
          * @return A new {@link Pinecone} instance configured based on the builder parameters.
          */
         public Pinecone build() {
-            PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig);
+            PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig, customOkHttpClient);
```
rohanshah18, line 1036: `customOkHttpClient` being passed to the `PineconeConfig` constructor at the time of initializing `PineconeConfig` in `Pinecone` class

src/main/java/io/pinecone/configs/PineconeConfig.java
```diff
@@ -70,21 +72,23 @@ public PineconeConfig(String apiKey) {
...
      */
     public PineconeConfig(String apiKey, String sourceTag) {
-        this(apiKey, sourceTag, null);
+        this(apiKey, sourceTag, null, null);
     }
 
     /**
-     * Constructs a {@link PineconeConfig} instance with the specified API key, source tag, control plane proxy
-     * configuration, and data plane proxy configuration.
+     * Constructs a {@link PineconeConfig} instance with the specified API key, source tag, a HTTP proxy configuration,
```
rohanshah18, line 79: `customOkHttpClient` is now a member of `PineconeConfig`. It makes a lot of sense since gRPC custom channel is already a part of this config class.

ssmith-pc (approved):
LGTM! Thanks for separating the API update from the rerank feature impl

PR #159, opened 2024-10-18
Add rerank

rohanshah18:
## Problem

Add rerank.

## Solution

1. Added the following two methods for rerank:
a. Method that accepts required parameters only and the optional parameters are set to default. More details on the api and its default parameters can be found [here](https://docs.pinecone.io/reference/api/2024-10/inference/rerank).
```java
      rerank(String model,
                       String query,
                       List<Map<String, String>> documents)
```
b. Method that accepts all parameters:
```java
      rerank(String model,
                       String query,
                       List<Map<String, String>> documents,
                       List<String> rankFields,
                       int topN,
                       boolean returnDocuments,
                       Map<String, Object> parameters)
```
2. Added integration tests for both rerank methods.
3. Added an example in the README file.

## Type of Change

## Test Plan

Added integration test.


ssmith-pc (approved):
LGTM!

PR #161, opened 2024-10-22
Add imports

rohanshah18:
## Problem

Add four endpoints of the `BulkOperationsApi`.

## Solution

Added the following four endpoints of the `BulkOperationsApi`:
1. `startImport(String uri, String integrationId, ImportErrorMode.OnErrorEnum errorMode)`
2. `describeImport(Integer limit, String paginationToken)`
3. `listImport(String id)`
4. `cancelImport(String id)`

## Type of Change

## Test Plan

Added unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,7 +704,7 @@ ImportModel importDetails = asyncIndex.describeImport("1");
 
 ## Cancel an import
 
-import org.openapitools.db_data.client.ApiException;
+The following example attempts to cancel an ongoing import operation using its unique identifier.
 
 ```java
 import org.openapitools.db_data.client.ApiException;
--- a/README.md
+++ b/README.md
@@ -643,6 +643,82 @@ RerankResult result = inference.rerank(model, query, documents, rankFields, topN
 System.out.println(result.getData());
 ```
 
+# Imports
+## Start an import
+
+The following example initiates an asynchronous import of vectors from object storage into the index.
+
+```java
+import org.openapitools.db_data.client.ApiException;
+import org.openapitools.db_data.client.model.ImportErrorMode;
+import org.openapitools.db_data.client.model.StartImportResponse;
+...
+
+// Initialize pinecone object
+Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
+// Get async imports connection object
+AsyncIndex asyncIndex = pinecone.getAsyncIndexConnection("PINECONE_INDEX_NAME");
+
+// s3 uri
+String uri = "s3://path/to/file.parquet";
+
+// Start an import
+StartImportResponse response = asyncIndex.startImport(uri, "123-456-789", ImportErrorMode.OnErrorEnum.CONTINUE);
+```
+
+## List imports
+
+The following example lists all recent and ongoing import operations for the specified index.
+
+```java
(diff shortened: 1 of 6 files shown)


src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -1039,6 +1068,124 @@ public ListenableFuture<ListResponse> list(String namespace, String prefix, Stri
...
+     *     ListImportsResponse response = asyncIndex.listImport(limit, paginationToken);
+     *  }</pre>
+     *
+     * @param limit The maximum number of operations to return per page. Default is 100.
+     * @param paginationToken The token to continue a previous listing operation. Can be null or empty.
+     * @return {@link ListImportsResponse} containing the list of recent and ongoing import operations.
+     * @throws ApiException if there are issues processing the request or communicating with the server.
+     *         This includes network issues, server errors, or serialization issues with the request or response.
+     */
+    public ListImportsResponse listImport(Integer limit, String paginationToken) throws ApiException {
```
jhamon, line 1136: Should this be plural?

```suggestion
    public ListImportsResponse listImports(Integer limit, String paginationToken) throws ApiException {
```

Also, an overload of this method with no arguments could be useful since the limit is supposed to have default 100 and pagination token is only known for pages 2+ of the results.

README.md
```diff
@@ -643,6 +643,82 @@ RerankResult result = inference.rerank(model, query, documents, rankFields, topN
...
+import org.openapitools.db_data.client.model.ListImportsResponse;
+...
+
+// Initialize pinecone object
+Pinecone pinecone = new Pinecone.Builder("PINECONE_API_KEY").build();
+// Get async imports connection object
+AsyncIndex asyncIndex = pinecone.getAsyncIndexConnection("PINECONE_INDEX_NAME");
+
+// List imports
+ListImportsResponse response = asyncIndex.listImport(100, "some-pagination-token");
```
jhamon, line 684: We need a way to call this without pagination token, since it won't be known ahead of the first call. They could pass `null` I guess, but that's kinda ugly.

src/main/java/io/pinecone/clients/AsyncIndex.java
```diff
@@ -1039,6 +1068,124 @@ public ListenableFuture<ListResponse> list(String namespace, String prefix, Stri
         return asyncStub.list(listRequest);
     }
 
+    /**
```
rohanshah18, line 1071: The reason why I added imports in AsyncIndex is because the operations are async and for data plane based.

jhamon (approved):

README.md
ssmith-pc: Not suggesting we do it in this PR, but seems like we could improve the user experience around pagination by seamlessly/transparently handling it within the SDK impl, and exposing via a standard iterator for example.

jhamon: > exposing via a standard iterator 

Yes, we do this in the python implementation. Seems like a good follow-up item.

rohanshah18: Created an internal ticket for this.

src/main/java/io/pinecone/clients/AsyncIndex.java
rohanshah18: Fixed!

PR #168, opened 2024-11-25
Add support for passing base path for control plane operations

rohanshah18:
## Problem

Currently users don't have an option to pass in the base path for control plane operations.  

## Solution

Add `host` and `enableTls` fields to Pinecone's inner class called `Builder` and further set the same values for `PineconeConfig`. This allows the same configuration to be used for data plane operations, eliminating the need for users to manually instantiate `PineconeConfig` and pass it to the `Pinecone` class, as was required in the previous approach..

Existing approach for data plane operations:
```java
// Instantiate ManageIndexesApi class
ManageIndexesApi manageIndexesApi = new ManageIndexesApi();

// Instantiate PineconeConfig class
PineconeConfig config = new PineconeConfig("apiKey");
// Disable tls for gRPC calls
config.setTLSEnabled(false);
// Set host for gRPC calls
config.setHost(host);

// Instantiate Pinecone class using PineconeConfig and ManageIndexesApi
Pinecone pinecone = new Pinecone(config, manageIndexesApi);

// Get index connection object for the index
Index index = pinecone.getIndexConnection("index");

// Upsert records into the index
index.upsert("v1", Arrays.asList(1f, 2f, 3f));
```

New approach for control and data plane operations:
```java
// Sets the host for both control and data plane operations
Pinecone pinecone = new Pinecone.Builder("apiKey")
       .withHost(host)
       .withTlsEnabled(false)
       .build();

// Create index (control plane operation)
pinecone.createServerlessIndex("index", "cosine", 3, "aws","us-east-1", DeletionProtection.DISABLED);

// Get index connection object for the index
Index index = pinecone.getIndexConnection("index");

// Upsert records into the index (data plane operation)
index.upsert("v1", Arrays.asList(1f, 2f, 3f));
```
## Type of Change

## Test Plan

Existing integration tests should work as it and tested locally using the emulator.

### DIFF
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -1025,11 +1025,52 @@ public class Pinecone {
             return this;
         }
 
+        /**
+         * Sets a custom host URL for the control and data plane operations.
+         * <p>
+         * This method allows you to specify a custom base URL for Pinecone control and data plane requests.
+         * <p>
+         * Example usage:
+         * <pre>{@code
+         * Pinecone client = new Pinecone.Builder("PINECONE_API_KEY")
+         *     .withHost("http://localhost:5080")
+         *     .build();
+         *
+         * // Requests will now be sent to the specified host.
+         * client.listIndexes();
+         * }</pre>
+         *
+         * @param host The custom host URL for the Pinecone service. Must be a valid URL.
+         * @return This {@link Builder} instance for chaining method calls.
+         */
         public Builder withHost(String host) {
             this.host = host;
             return this;
         }
 
+        /**
+         * Configures whether TLS (Transport Layer Security) should be enabled for data plane operations.
+         * <p>
+         * By default, TLS is enabled for data plane requests to ensure secure communication for data plane operations.
+         * This method can be used to disable TLS if needed (e.g., for testing or when communicating with non-secure
+         * endpoints). Disabling TLS in a production environment is not recommended due to potential security risks.
+         * <p>
+         * Example usage:
+         * <pre>{@code
+         * Pinecone client = new Pinecone.Builder("PINECONE_API_KEY")
+         *     .withTlsEnabled(false)
+         *     .build();
+         *
+         * // Get index for data plane operations
+         * Index index = pinecone.getIndexConnection("PINECONE_INDEX_NAME");
+         *
+         * // Requests will now be made without TLS encryption (not recommended for production use).
+         * index.upsert("v1", Arrays.asList(1f, 2f, 3f));
+         * }</pre>
+         *
+         * @param enableTls {@code true} to enable TLS (default), {@code false} to disable it.
+         * @return This {@link Builder} instance for chaining method calls.
+         */
         public Builder withTlsEnabled(boolean enableTls) {
             this.enableTls = enableTls;
             return this;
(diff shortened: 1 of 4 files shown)


src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -1023,6 +1025,16 @@ public Builder withProxy(String proxyHost, int proxyPort) {
...
+        public Builder withHost(String host) {
+            this.host = host;
+            return this;
+        }
+
+        public Builder withTlsEnabled(boolean enableTls) {
+            this.enableTls = enableTls;
+            return this;
+        }
+
```
austin-denoble, line 1037: Should these have some sort of documentation similar to `withProxy` above?

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -1034,13 +1046,23 @@ public Builder withProxy(String proxyHost, int proxyPort) {
...
 
-            ApiClient apiClient = (customOkHttpClient != null) ? new ApiClient(customOkHttpClient) : new ApiClient(buildOkHttpClient(proxyConfig));
+            ApiClient apiClient;
+            if (customOkHttpClient != null) {
+                apiClient = new ApiClient(customOkHttpClient);
+            } else {
+                apiClient = new ApiClient(buildOkHttpClient(proxyConfig));
+                if(host!=null && !host.isEmpty()) {
+                    config.setHost(host);
+                    apiClient.setBasePath(host);
```
austin-denoble, line 1063: If the host is applied to the `apiClient` does it also need to be passed to `config` here? From what it looks like, this value is overwritten when you target an index: https://github.com/pinecone-io/pinecone-java-client/blob/main/src/main/java/io/pinecone/clients/Pinecone.java#L844-L874

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -1034,13 +1046,23 @@ public Builder withProxy(String proxyHost, int proxyPort) {
          */
         public Pinecone build() {
             PineconeConfig config = new PineconeConfig(apiKey, sourceTag, proxyConfig, customOkHttpClient);
+            config.setTLSEnabled(enableTls);
```
austin-denoble, line 1049: So this is only needed for `PineconeConfig` and we don't need it for the `ApiClient` itself?

austin-denoble (approved):
This looks ok to me, but I had a couple questions since it's been a while since I've looked at this codebase.

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes I need to the docstrings

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes, this enableTls flag is only needed for gRPC connection so the pineconeConfig object is set and further fetched at the time of establishing gRPC connection.

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Currently we are making an extra describeIndex call using the same Pinecone object so the overwritten value is still pointing to the same host but this extra call can be removed. The only approach I came up with for now, in order to solve this, is instantiating the PineconeConfig object outside of Builder class which I don't like it. So I have a ticket to remove this extra call in future and come up with a better solution. Also the ticket will take care of asyncIndexConnection as well.

PR #169, opened 2024-12-19
Add index tags 

rohanshah18:
## Problem

Add support for Index tags.

## Solution

Index tags can be passed as a Map<String, String> to the following operations:
1. create serverless index
2. create pod Index
3. configure serverless index
4. configure pod index

When configuring tags, following are the special cases:
1. If an empty map is passed as an input, the tags will remain unchanged
2. If an empty value is passed as an input, the tag will be removed

## Type of Change

## Test Plan

Updated integration tests for all 4 operations.

### DIFF
--- a/src/integration/java/io/pinecone/integration/controlPlane/serverless/DeletionProtectionTest.java
+++ b/src/integration/java/io/pinecone/integration/controlPlane/serverless/DeletionProtectionTest.java
@@ -8,6 +8,7 @@ import org.openapitools.db_control.client.model.DeletionProtection;
 import org.openapitools.db_control.client.model.IndexModel;
 
 import java.util.HashMap;
+import java.util.Map;
 
 public class DeletionProtectionTest {
     private static final Pinecone controlPlaneClient = new Pinecone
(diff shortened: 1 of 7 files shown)


src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
```diff
@@ -70,13 +69,6 @@ public void testSyncListEndpoint() throws InterruptedException {
                 listResponseWithPaginationNoPrefix1.getPagination().getNext()
         );
         assertEquals(listResponseWithPaginationNoPrefix2.getVectorsList().size(), 2);
-        ListResponse listResponseWithPaginationNoPrefix3 = indexConnection.list(
```
rohanshah18, line 73: Remove extra checks that are failing consistently in integration tests

src/integration/java/io/pinecone/integration/controlPlane/serverless/DeletionProtectionTest.java
```diff
@@ -16,8 +18,10 @@ public class DeletionProtectionTest {
     @Test
     public void createIndexWithDeletionProtectionEnabled() {
         String indexName = RandomStringBuilder.build("create-serv", 8);
+        HashMap<String, String> tags = new HashMap<>();
+        tags.put("test", "deletion-protection-enabled");
         // Create serverless index with deletion protection enabled
-        controlPlaneClient.createServerlessIndex(indexName, "cosine", 3, "aws", "us-west-2", DeletionProtection.ENABLED);
+        controlPlaneClient.createServerlessIndex(indexName, "cosine", 3, "aws", "us-west-2", DeletionProtection.ENABLED, tags);
```
austin-denoble, line 24: You may want to add an assertion for the tags that are applied to the index since we're doing that for `DeletionProtection`.

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
austin-denoble: Why are they failing?

austin-denoble (approved):
LGTM, just left a few comments. Nice work!

src/integration/java/io/pinecone/integration/dataPlane/ListEndpointTest.java
rohanshah18: The total number of vectors in the custom namespace ("example-namespace") is 4. When `listResponseWithPaginationNoPrefix1` returns 2 vectors and includes a pagination token pointing to the remaining vector IDs, these are then returned as part of `listResponseWithPaginationNoPrefix2`, which has a limit of 2. Since there are no more vector IDs left, no pagination token will be returned in the response of `listResponseWithPaginationNoPrefix2`. As a result, the assertion on line #79 will fail with the error message: "Pagination token cannot be null or empty." Similarly, this issue will also occur in the asynchronous test.

src/integration/java/io/pinecone/integration/controlPlane/serverless/DeletionProtectionTest.java
rohanshah18: Good point, added the additional assertions

PR #174, opened 2025-01-24
Replace protoc with buf and add build script for generating code from the protos

rohanshah18:
## Problem

The code generation using proto file was done using protoc and the process was manual.

## Solution

Added a script to generate java classes using [buf](https://buf.build/docs/generate/overview/). 

## Type of Change

## Test Plan

Integration tests should run successfully.


build.gradle
```diff
@@ -52,7 +52,7 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.9.1'
     implementation 'io.gsonfire:gson-fire:1.8.5'
     implementation 'org.openapitools:jackson-databind-nullable:0.2.6'
-    implementation 'com.google.protobuf:protobuf-java:3.25.2'
+    implementation 'com.google.protobuf:protobuf-java:4.29.3'
```
rohanshah18, line 55: Had to bump protobuf-java dependency version so its compatible with Buf

src/integration/java/io/pinecone/helpers/TestUtilities.java
```diff
@@ -1,15 +1,10 @@
 package io.pinecone.helpers;
 
-import io.pinecone.clients.Index;
```
rohanshah18, line 3: removed extra imports as a part of safe delete operation of intellij when removing older proto generated classes

src/integration/java/io/pinecone/integration/controlPlane/pod/CollectionTest.java
```diff
@@ -5,9 +5,9 @@
 import io.pinecone.exceptions.PineconeException;
 import io.pinecone.exceptions.PineconeValidationException;
 import io.pinecone.helpers.RandomStringBuilder;
-import io.pinecone.proto.FetchResponse;
 import io.pinecone.helpers.TestResourcesManager;
 
+import io.pinecone.proto.FetchResponse;
```
rohanshah18, line 10: Sorry this should not be a change but removed the import as a part of safe delete operation of intellij when removing older proto generated classes and later I manually added it but added it on line#10 so it shows as a change 😢.

jhamon (approved):

PR #181, opened 2025-03-21
Add upsert and search records

rohanshah18:
## Problem

Describe the purpose of this change. What problem is being solved and why?

## Solution

Describe the approach you took. Link to any relevant bugs, issues, docs, or other resources.

## Type of Change

## Test Plan

### DIFF
--- a/src/main/java/io/pinecone/clients/Pinecone.java
+++ b/src/main/java/io/pinecone/clients/Pinecone.java
@@ -982,6 +982,12 @@ public class Pinecone {
         return new AsyncIndex(config, connection, indexName);
     }
 
+    public RestIndex getRestIndexConnection(String indexName) throws PineconeValidationException {
+        PineconeConfig perConnectionConfig = new PineconeConfig(config.getApiKey(), config.getSourceTag());
+        perConnectionConfig.setHost(getIndexHost(indexName));
+        return new RestIndex(perConnectionConfig);
+    }
+
     /**
      * A method to create and return a new instance of the {@link Inference} client.
      * <p>
--- a/src/main/java/io/pinecone/clients/RestIndex.java
+++ b/src/main/java/io/pinecone/clients/RestIndex.java
@@ -1,0 +1,44 @@
+package io.pinecone.clients;
+
+import io.pinecone.configs.PineconeConfig;
+import okhttp3.OkHttpClient;
+import org.openapitools.db_data.client.ApiClient;
+import org.openapitools.db_data.client.ApiException;
+import org.openapitools.db_data.client.Configuration;
+import org.openapitools.db_data.client.api.VectorOperationsApi;
+import org.openapitools.db_data.client.model.*;
+
+import java.util.List;
+
+import static io.pinecone.clients.Pinecone.buildOkHttpClient;
+
+public class RestIndex {
+    private final VectorOperationsApi vectorOperations;
+
+    public RestIndex(PineconeConfig config) {
+        OkHttpClient customOkHttpClient = config.getCustomOkHttpClient();
+        ApiClient apiClient = (customOkHttpClient != null) ? new ApiClient(customOkHttpClient) : new ApiClient(buildOkHttpClient(config.getProxyConfig()));
+        apiClient.setApiKey(config.getApiKey());
+        apiClient.setUserAgent(config.getUserAgent());
+        apiClient.addDefaultHeader("X-Pinecone-Api-Version", Configuration.VERSION);
+
+        this.vectorOperations = new VectorOperationsApi(apiClient);
+        String protocol = config.isTLSEnabled() ? "https://" : "http://";
+        vectorOperations.setCustomBaseUrl(protocol + config.getHost());
+    }
+
+    public void upsertRecords(String namespace, List<UpsertRecord> upsertRecord) throws ApiException {
+        vectorOperations.upsertRecordsNamespace(namespace, upsertRecord);
+    }
+
+    public SearchRecordsResponse searchRecords(String id, String namespace, Object filter, int topK, SearchRecordsVector vector) throws ApiException {
+        SearchRecordsRequestQuery searchRecordsRequestquery = new SearchRecordsRequestQuery()
+                .id(id)
+                .filter(filter)
+                .topK(topK)
+                .vector(vector);
+        SearchRecordsRequest request = new SearchRecordsRequest().query(searchRecordsRequestquery);
+
+        return vectorOperations.searchRecordsNamespace(namespace, request);
+    }
+}
--- a/src/main/java/org/openapitools/db_data/client/ApiClient.java
+++ b/src/main/java/org/openapitools/db_data/client/ApiClient.java
@@ -931,6 +931,18 @@ public class ApiClient {
             return RequestBody.create((File) obj, MediaType.parse(contentType));
         } else if ("text/plain".equals(contentType) && obj instanceof String) {
             return RequestBody.create((String) obj, MediaType.parse(contentType));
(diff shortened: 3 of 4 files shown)


src/main/java/io/pinecone/clients/RestIndex.java
```diff
@@ -0,0 +1,44 @@
...
+        apiClient.setApiKey(config.getApiKey());
+        apiClient.setUserAgent(config.getUserAgent());
+        apiClient.addDefaultHeader("X-Pinecone-Api-Version", Configuration.VERSION);
+
+        this.vectorOperations = new VectorOperationsApi(apiClient);
+        String protocol = config.isTLSEnabled() ? "https://" : "http://";
+        vectorOperations.setCustomBaseUrl(protocol + config.getHost());
+    }
+
+    public void upsertRecords(String namespace, List<UpsertRecord> upsertRecord) throws ApiException {
```
ssmith-pc, line 30: Can we add an overload that accepts `List<Map<String, Object>>`?

ssmith-pc: This feels kind of cumbersome:

```java
        UpsertRecord record1 = new UpsertRecord();
        record1.id("rec1");
        record1.putAdditionalProperty("category", "digestive system");
        record1.putAdditionalProperty("chunk_text", "Apples are a great source of dietary fiber, which supports digestion and helps maintain a healthy gut.");
        ArrayList<UpsertRecord> records = new ArrayList<>();

        UpsertRecord record2 = new UpsertRecord();
        record2.id("rec2");
        record2.putAdditionalProperty("category", "cultivation");
        record2.putAdditionalProperty("chunk_text", "Apples originated in Central Asia and have been cultivated for thousands of years, with over 7,500 varieties available today.");
```

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -944,6 +978,269 @@ public ListResponse list(String namespace, String prefix, String paginationToken
...
+     *                     stored.
+     * @throws ApiException If there is an issue with the upsert operation. This could include network errors,
+     *                      invalid input data, or issues communicating with the Pinecone service.
+     */
+    public void upsertRecords(String namespace, List<Map<String, String>> upsertRecords) throws ApiException {
+        List<UpsertRecord> records = new ArrayList<>();
+        for(Map<String, String> record: upsertRecords) {
+            UpsertRecord upsertRecord = new UpsertRecord();
+            for (Map.Entry<String, String> entry : record.entrySet()) {
+                if(entry.getKey().equals("_id")) {
```
austin-denoble, line 1015: Is it possible for an existing `id` to be overwritten by this logic? The server will respond with a 400 if you try and send both `id` and `_id`:

```
{"error":{"code":"INVALID_ARGUMENT","message":"Error parsing request: Invalid input: Document cannot have both _id and id fields (_id='rec1', id='rec1')"},"status":400}
```

Maybe instead of mapping one to the other in the client we just pass through what the user has provided, and if they need to clean up ids they can? I just checked for the presence of one or the other on the record in typescript: https://github.com/pinecone-io/pinecone-ts-client/blob/e5b39d7a6c77836b5b0ae1f696ce7e6f6020f82e/src/data/vectors/upsertRecords.ts#L29-L37

src/main/java/io/pinecone/clients/Index.java
```diff
@@ -944,6 +978,269 @@ public ListResponse list(String namespace, String prefix, String paginationToken
...
+     * @param topK The maximum number of results to be returned by the search.
+     * @param filter (Optional) A filter to apply to the search query. It can be used to narrow down the search
+     *               based on specific criteria.
+     * @param rerank (Optional) A reranking operation that can be applied to refine or reorder the search results.
+     *               Pass null if no reranking is required.
+     * @return A {@link SearchRecordsResponse} object containing the search results, including the top matching records.
+     * @throws ApiException If there is an issue with the search operation. This could include network errors,
+     *                      invalid input data, or issues communicating with the Pinecone service.
+     */
+    public SearchRecordsResponse searchRecordsByText(String text,
```
austin-denoble, line 1223: Nice job handling the lack of optionals in all of these methods. 👍

src/main/java/io/pinecone/clients/Pinecone.java
```diff
@@ -726,10 +779,16 @@ public IndexModel configurePodsIndex(String indexName, DeletionProtection deleti
...
      * @param deletionProtection Enable or disable deletion protection for the index.
      * @param tags A map of tags to associate with the Index.
+     * @param embed Convert an existing index to an integrated index by specifying the embedding model and field_map.
+     *              The index vector type and dimension must match the model vector type and dimension, and the index
+     *              similarity metric must be supported by the model
      * @return {@link IndexModel} representing the configured index.
      * @throws PineconeException if an error occurs during the operation, the index does not exist, or if any of the arguments are invalid.
      */
-    public IndexModel configureServerlessIndex(String indexName, DeletionProtection deletionProtection, Map<String, String> tags) throws PineconeException {
+    public IndexModel configureServerlessIndex(String indexName,
```
austin-denoble, line 788: This would be a breaking change, right? In terms of it altering the method signature for `configureServerlessIndex`.

If we're intending for these changes to release as a major version bump that's fine, but otherwise we may want to look at overloading this method too to keep things backwards compatible.

austin-denoble (commented):
Overall this looks good, nice work! I've left a few comments around handling of ids in records, and a possibly breaking change to `configureServerlessIndex`.

src/main/java/io/pinecone/clients/Index.java
rohanshah18: Since the input type is a Map<String, String>, what should be the key?

src/main/java/io/pinecone/clients/Pinecone.java
rohanshah18: Yes that was the intention since I'm trying to limit the number of overloads for control plane

src/main/java/io/pinecone/clients/RestIndex.java
rohanshah18: After chatting offline, we decided to replace UpsertRecord class with Map<String, String>

src/integration/java/io/pinecone/integration/dataPlane/UpsertAndSearchRecordsTest.java
```diff
@@ -0,0 +1,83 @@
...
+                .inputs(inputsMap);
+
+        List<String> fields = new ArrayList<>();
+        fields.add("category");
+        fields.add("chunk_text");
+
+        // Wait for vectors to be upserted
+        Thread.sleep(5000);
+
+        SearchRecordsResponse recordsResponse = index.searchRecords(namespace, query, fields, null);
```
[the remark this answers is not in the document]
rohanshah18: Added search by: 
- text
- vector
- id

src/main/java/org/openapitools/db_data/client/ApiClient.java
```diff
@@ -931,6 +931,18 @@ public RequestBody serialize(Object obj, String contentType) throws ApiException
             return RequestBody.create((File) obj, MediaType.parse(contentType));
         } else if ("text/plain".equals(contentType) && obj instanceof String) {
             return RequestBody.create((String) obj, MediaType.parse(contentType));
+        } else if ("application/x-ndjson".equals(contentType)) {
```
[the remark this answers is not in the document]
rohanshah18: Good point, will merge this change as a part of next PR

austin-denoble (approved):
LGTM, thanks for iterating

PR #197, opened 2025-09-25
Update gradle version, fix flaky integration tests, and add cleanup job

rohanshah18:
## Problem

1. Integration tests have been failing 
2. [submit-gradle](https://github.com/pinecone-io/pinecone-java-client/actions/runs/18015820954/job/51260319239?pr=197#logs) job was failing because it used java 21 with incompatible gradle version.

## Solution

1. Fix flaky integration tests
2. Add cleanup job that deletes all indexes after integration tests are completed (regardless of success/failure of the job)
3. Update gradle version to 8.5 for all workflows

## Type of Change

## Test Plan

Integration tests should run successfully


src/main/java/io/pinecone/helpers/IndexCleanupUtility.java
```diff
@@ -0,0 +1,36 @@
...
+            for(IndexModel model : pinecone.listIndexes().getIndexes()) {
+                String indexName = model.getName();
+                if(model.getDeletionProtection().equals(DeletionProtection.ENABLED)) {
+                    if(model.getSpec().getPod() != null) {
+                        pinecone.configurePodsIndex(indexName, DeletionProtection.DISABLED);
+                    }
+                    pinecone.configureServerlessIndex(indexName, DeletionProtection.DISABLED, null, null);
+                }
+                Thread.sleep(5000);
+                pinecone.deleteIndex(indexName);
```
austin-denoble, line 26: There's a small chance the index may not be deletable if it's stuck in an upgrading or stuck or something like that. It might be nice to add a try catch here and explicitly handle possible errors on delete to make sure the cleanup process continues to other indexes.

